PR #6, opened 2026-02-12
SLCORE-2044 setup pipelines

sophio-japharidze-sonarsource:


.actions/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "commonjs"
```
damien-urruty-sonarsource: What is this for? It's the first time I see such a file there

.github/workflows/build.yml
```diff
@@ -21,15 +21,230 @@ permissions:
   contents: read
 
 jobs:
+  prepare:
+    runs-on: ubuntu-latest
```
damien-urruty-sonarsource: I think we cannot use that. Since it's a private repo we need to use private runners. See https://xtranet-sonarsource.atlassian.net/wiki/spaces/Platform/pages/3694231566/GitHub+Actions+Runners+-+GitHub#Private-Self-Hosted-Runners-(sonar-*)

Same below in other jobs

.github/workflows/build.yml
```diff
@@ -21,15 +21,230 @@ permissions:
...
+    name: Prepare
+    outputs:
+      PACKAGE_VERSION: ${{ steps.project_version.outputs.PACKAGE_VERSION }}
+      PROJECT_VERSION: ${{ steps.project_version.outputs.PROJECT_VERSION }}
+      BUILD_NUMBER: ${{ steps.get-build-number.outputs.BUILD_NUMBER }}
+    steps:
+      - uses: actions/checkout@v4
```
damien-urruty-sonarsource: As it was discussed with Andrii the other day when talking about Renovate, he strongly suggested to use commit hashes instead of tags, to make sure the builds are reproducible. And btw we could configure Renovate too on the repo

.github/workflows/build.yml
```diff
@@ -21,15 +21,230 @@ permissions:
   contents: read
 
 jobs:
+  prepare:
+    runs-on: ubuntu-latest
+    name: Prepare
```
damien-urruty-sonarsource: What is the purpose of this step? Just to populate some variables for the following steps? Shouldn't we build first and then propagate the versions?

.github/workflows/build.yml
```diff
@@ -21,15 +21,230 @@ permissions:
...
+        env:
+          ARTIFACTORY_PRIVATE_READER_USERNAME: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_USER }}
+          ARTIFACTORY_PRIVATE_READER_PASSWORD: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+          NPM_CONFIG_registry: https://repox.jfrog.io/artifactory/api/npm/npm
+          NPM_CONFIG_//repox.jfrog.io/artifactory/api/npm/:_authToken: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+        run: |
+          npm ci
+
+      - name: Build
+        run: npm run build
```
damien-urruty-sonarsource: Is there a reason not to use the common `build-npm` action? Or maybe at least `config-npm`? I would recommend to use the common actions if we can

.github/workflows/build.yml
```diff
@@ -21,15 +21,230 @@ permissions:
...
+          NPM_CONFIG_registry: https://repox.jfrog.io/artifactory/api/npm/npm
+          NPM_CONFIG_//repox.jfrog.io/artifactory/api/npm/:_authToken: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+        run: npm ci
+
+      - name: Run all tests
+        run: npm run test:all
+
+  build-binaries:
+    name: Build Binary - ${{ matrix.platform }}
+    runs-on: ${{ matrix.os }}
```
damien-urruty-sonarsource: I think Bun supports cross-compilation, so you don't need to run from different OS-specific runners. What I would suggest is to keep a single job to build all 3 artifacts. In fact this could be the first job to be run, usually we try to build artifacts in the first job, then promote them in later jobs

.actions/package.json
sophio-japharidze-sonarsource: the base `package.json` of this project specifies `type: module`, and it was conflicting with the commonjs script from shared github actions.

vnaskos-sonar:
You can also copy-paste the [notify-failure workflow](https://github.com/SonarSource/sonarlint-vscode/blob/master/.github/workflows/notify-failure.yml)

mise.toml
```diff
@@ -0,0 +1,7 @@
+[tools]
+node = '20'
+bun = "1.3.9"
+
+[tasks.install-system-deps]
+# Install system dependences and disable man-db auto-update to prevent it from hanging during installation
+run = "echo 'man-db man-db/auto-update boolean false' | sudo debconf-set-selections && sudo dpkg-reconfigure man-db && sudo apt-get update && sudo apt-get install -y xvfb x11-utils ffmpeg libxrandr2 libglu1-mesa"
```
vnaskos-sonar: This can be removed as well, I noticed that we don't use it in the build.yml (no `mise run mise run install-system-deps`). Since this is a cli and no UI is required, it shouldn't be needed.

.github/workflows/build.yml
sophio-japharidze-sonarsource: Yes, for sure, but there are problems with network & certificates when bun tries to download something. We will neex help from EngExp to make it work.

kirill-knize-sonarsource (approved):
LGTM!

PR #9, opened 2026-02-19
CLI-4 Integrate secrets Claude hooks into CLI

kirill-knize-sonarsource:
Stop relying on global env vars to provide token to secret CLI. Take token from keystore and provide it as an env var to process spawn during call. Move secret hooks installation from `sonar install secrets` to `sonar onboard-agent claude`.
Tested on Mac - doesn't require keystore passwords on hooks execution.

Coverage is to be fixed in separate PR.


src/bootstrap/hooks.ts
```diff
@@ -184,239 +194,134 @@ export async function areHooksInstalled(projectRoot: string): Promise<boolean> {
...
-$toolName = $input.tool_name
-$filePath = $input.tool_input.file_path
+    // Create hooks directory
+    mkdirSync(hooksPath, { recursive: true });
 
-# Only analyze if Edit or Write was used
-if ($toolName -ne "Edit" -and $toolName -ne "Write") {
-    exit 0
-}
+    const sourceSecretsDir = join(homedir(), '.claude', 'hooks', 'sonar-secrets');
```
damien-urruty-sonarsource: At some point we discussed calling back the SQ CLI instead of secrets directly, is this part of a separate ticket?

src/commands/secret.ts
```diff
@@ -249,7 +243,7 @@ async function performFileScan(
  */
 export async function secretStatusCommand(): Promise<void> {
   const platform = detectPlatform();
-  const binDir = join(homedir(), '.sonarqube-cli', 'bin');
+  const binDir = join(homedir(), '.sonar-cli', 'bin');
```
damien-urruty-sonarsource: As per recent discussions, should we keep it as 'sonarqube-cli'?

damien-urruty-sonarsource (approved):
LGTM, there are a lot of changes so I might have missed some details

src/commands/secret.ts
kirill-knize-sonarsource: Probably yes. Changed back to `.sonarqube-cli`

src/bootstrap/hooks.ts
kirill-knize-sonarsource: It's an artifact of development. Real hook will use our command `sonar secret check`, but this logic was not removed. Good catch, thanks.

PR #10, opened 2026-02-19
CLI-11 command for searching projects

sophio-japharidze-sonarsource:


src/lib/types.ts
```diff
@@ -77,6 +77,27 @@ export interface IssuesSearchResponse {
   }>;
 }
 
+export interface SonarQubeProject {
+  key: string;
+  name: string;
+}
+
+export interface ProjectsSearchParams {
```
kirill-knize-sonarsource, line 85: analyzedBefore, onProvisionedOnly, qualifiers - there are some more parameters WebAPI accept. Maybe we want to add them as well? Possible improvement for later.

kirill-knize-sonarsource (approved):
We can implement more parameters, but let's not block on it now.

src/lib/types.ts
sophio-japharidze-sonarsource: I started with the same parameters that our MCP Server exposes. For now it should be enough. And indeed, we can always add more later

PR #11, opened 2026-02-20
CLI-18 Fix http vulnerabilities for loopback server

kirill-knize-sonarsource:


src/bootstrap/auth.ts
```diff
@@ -1,16 +1,22 @@
...
 
-const MIN_PORT = 64130;
-const MAX_PORT = 64140;
-const HTTP_STATUS_OK = 200;
 const PORT_TIMEOUT_MS = 50000;
+const HTTP_STATUS_OK = 200;
+const HTTP_STATUS_PAYLOAD_TOO_LARGE = 413;
+const TOKEN_DISPLAY_LENGTH = 20;
+const MAX_POST_BODY_BYTES = 4096;
+const SUCCESS_HTML_TITLE = 'SonarLint Authentication';
```
damien-urruty-sonarsource, line 16: So 2024 :)

src/bootstrap/auth.ts
```diff
@@ -307,10 +136,193 @@ function getSuccessHTML(): string {
...
+  // 4. Wait for user input
+  await waitForUserInput();
+
+  // 5. Open browser
+  await openBrowserWithFallback(authURL);
+
+  logger.info('\n   ⏳ Waiting for authorization (50 second timeout)...');
+  logger.debug(`Waiting for callback on http://127.0.0.1:${server.port}`);
+
+  // 6. Wait for token with timeout (50 seconds)
```
damien-urruty-sonarsource, line 298: I wanted to discuss with you about this point. I implemented the SQC frontend in the same way as for SQ-IDE, which means if the local server is not reachable, I fallback to showing the token and allow users to copy it, so they could paste it in the terminal. Do you think it's feasible to support this flow?

src/lib/loopback-server.ts
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * @param onRequest - Handler function for incoming requests
+ * @param options - Optional configuration (e.g., port range)
+ * @returns Promise with port and close function
+ */
+export async function startLoopbackServer(
+  onRequest: RequestHandler,
+  options?: { portRange?: [number, number] }
+): Promise<LoopbackServerResult> {
+  const portRange: [number, number] = options?.portRange ?? [MIN_PORT, MAX_PORT];
```
damien-urruty-sonarsource, line 92: Here I think we could use a simpler approach, without a port range. The port range is useful for SQ-IDE because the SQC/SQS frontend scans the range to find the IDE. Here we don't have this need. We could start a webserver on a random port (let it be assigned automatically by the OS) and pass this port to the frontend.

damien-urruty-sonarsource (approved):
LGTM, a couple of questions or suggestions for improvement (can be handled separately)

src/bootstrap/auth.ts
kirill-knize-sonarsource: Ops. Fixed.

src/lib/loopback-server.ts
kirill-knize-sonarsource: Done.

src/bootstrap/auth.ts
kirill-knize-sonarsource: Yes, we need to support it, I've made the ticket
https://sonarsource.atlassian.net/browse/CLI-27

PR #13, opened 2026-02-20
CLI-18 Update port allocation logic

kirill-knize-sonarsource:


src/bootstrap/auth.ts
```diff
@@ -13,7 +13,7 @@ const HTTP_STATUS_OK = 200;
 const HTTP_STATUS_PAYLOAD_TOO_LARGE = 413;
 const TOKEN_DISPLAY_LENGTH = 20;
 const MAX_POST_BODY_BYTES = 4096;
-const SUCCESS_HTML_TITLE = 'SonarLint Authentication';
+const SUCCESS_HTML_TITLE = 'Sonar CLI Authentication';
```
damien-urruty-sonarsource, line 16: `SonarQube CLI` is the official name I think?

tests/unit/auth-helpers.test.ts
```diff
@@ -50,7 +50,7 @@ describe('Auth Helper Functions', () => {
 
     it('should contain success title', () => {
       const html = getSuccessHTML();
-      expect(html).toContain('SonarLint Authentication');
+      expect(html).toContain('Sonar CLI Authentication');
```
damien-urruty-sonarsource, line 53: Same

src/bootstrap/auth.ts
```diff
@@ -267,12 +267,15 @@ export async function generateTokenViaBrowser(serverURL: string): Promise<string
   });
 
   // 1. Start embedded HTTP server with token extraction handler
+  // Allow the Sonar server origin so the OAuth callback POST is not blocked by DNS rebinding protection
```
damien-urruty-sonarsource, line 270: We are not really relying on OAuth, but on our homemade protocol. Also we only use GET when we send the token back. Not very academic but I don't think we will change that now

src/lib/loopback-server.ts
```diff
@@ -78,73 +76,67 @@ function mergeSecurityHeadersWithUserHeaders(
...
+      const isExternalAllowedOrigin = !!(origin && !isValidLoopbackOrigin(origin) && allowedOrigins.includes(origin));
 
       // Handle OPTIONS preflight requests
       if (req.method === 'OPTIONS') {
-        res.writeHead(HTTP_STATUS_OK, getSecurityHeaders());
+        const preflightHeaders: Record<string, string> = { ...getSecurityHeaders() };
+        // Add CORS headers for allowed external origins (e.g. SonarCloud OAuth callback)
+        if (origin && (isValidLoopbackOrigin(origin) || allowedOrigins.includes(origin))) {
+          preflightHeaders['Access-Control-Allow-Origin'] = origin;
+          preflightHeaders['Access-Control-Allow-Methods'] = 'GET, POST, OPTIONS';
```
damien-urruty-sonarsource: I think we can remove at least POST form this list

damien-urruty-sonarsource (approved):

src/bootstrap/auth.ts
kirill-knize-sonarsource: It's Sonar CLI for now everywhere. Not only here.
I suggest to globally rename in another PR related to branding and messaging.

tests/unit/auth-helpers.test.ts
kirill-knize-sonarsource: Same

src/lib/loopback-server.ts
kirill-knize-sonarsource: Done

src/bootstrap/auth.ts
kirill-knize-sonarsource: Yes, I will change this comment during cleanup.

PR #15, opened 2026-02-23
CLI-19, CLI-22, CLI-27, CLI-28 + some more

kirill-knize-sonarsource:


docs/error-messages-audit.md
```diff
@@ -0,0 +1,484 @@
+# Error Messages Audit: CLI Argument Validation
```
damien-urruty-sonarsource: What is the purpose of this file?

scripts/generate-docs.js
```diff
@@ -0,0 +1,133 @@
...
+function padEnd(str, width) {
+  return str + ' '.repeat(Math.max(0, width - displayWidth(str)));
+}
+
+function renderOptionsTable(options) {
+  if (options && options.length > 0) {
+    const headers = ['Option', 'Type', 'Required', 'Description', 'Default'];
+    const cells = options.map(opt => {
+      const flag = opt.alias ? `\`--${opt.name}\`, \`-${opt.alias}\`` : `\`--${opt.name}\``;
+      const required = opt.required ? '✅' : '❌';
```
damien-urruty-sonarsource, line 44: When you flashed the screen earlier this morning, the red-cross caught my eye as if something was wrong. I'm not sure those 2 icons are the most appropriate to represent if an argument is required or optional, these are more for 'pass-fail' types of situation. Maybe just 'Yes' and 'No' would be sufficient here for the doc since the column is named 'Required'?

src/ui/messages.ts
```diff
@@ -0,0 +1,48 @@
+// Inline terminal output — non-interactive, static messages
+
+import { isTTY, cyan, yellow, red, green } from './colors.js';
+import { isMockActive, recordCall } from './mock.js';
```
damien-urruty-sonarsource, line 4: Why do we need a mock? Is it for testing purposes?

damien-urruty-sonarsource (approved):
I's almost a full rewrite, I cannot review everything... Let's merge this, but I suggest moving forward we try to keep the PRs reviewable and focus on a single topic at a time

src/ui/messages.ts
kirill-knize-sonarsource: Yes, testing is still clumsy in some places. It's not easy to do something better, but ideally get rid of it before the release. It's not good.

scripts/generate-docs.js
kirill-knize-sonarsource: Changed. Good point

docs/error-messages-audit.md
kirill-knize-sonarsource: slipped through gitignore. Removed.

PR #16, opened 2026-02-24
CLI-7 Adjust parameters for token generation

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM

src/bootstrap/auth.ts
```diff
@@ -91,6 +91,10 @@ export function extractTokenFromQuery(host: string | undefined, url: string | un
  */
 export function buildAuthURL(serverURL: string, port: number): string {
   const cleanServerURL = serverURL.replace(/\/$/, '');
+  if (serverURL.includes("sonarcloud")) {
```
sophio-japharidze-sonarsource: What about `sonarqube.us`? Or do we not support it?

damien-urruty-sonarsource: Good point. Do we support SQC US with the CLI?

sophio-japharidze-sonarsource: @kirill-knize-sonarsource confirmed that we do

damien-urruty-sonarsource: OK cool, I will add it to this if. This is a temporary hack until SQS is implemented, the if will probably disappear at some point, or take another form.

PR #18, opened 2026-02-24
Improve auth flow

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM
Side note: we need to find a way to get rid of those `isMockActive()` calls

PR #20, opened 2026-02-24
CLI-10 Create install scripts

vnaskos-sonar:
1. Move `scripts` to `build-scripts`.
1. Create `user-scripts` to contain the install scripts and publish them along with the executables.
1. Create `install.sh` for Linux and Mac.
1. Create `install.ps1` for Windows.
1. Create dynamically on the pipeline `latest-version.txt` that will contain the latest version as text.
1. Download the latest version of the cli.
1. Move the cli executable to the correct path and rename it to `sonar` or `sonar.exe`.
1. Check the checksum <-- this step will change with a more secure method.
1. Add the directory to the `PATH`.
    > Windows: `C:\Users\<username>\AppData\Local\sonarqube-cli\bin`
    > Linux and Macos: `/home/<username>/.local/share/sonarqube-cli/bin`
1. Rename artifacts to follow naming convention `sonarqube-cli-{version}.{build-number}-{platform}.exe`

# Run scripts

Windows:
```ps1
irm http://localhost:8080/install.ps1 | iex
```

Linux:
```sh
curl -fsSL http://localhost:8080/install.sh | bash
```


kirill-knize-sonarsource (approved):
LGTM

PR #21, opened 2026-02-24
CLI-31 Set up developer tooling to ensure consistency

nquinquenel:
- Introduces eslint/prettier to ensure code quality & consistency. Some rules are based from other existing Sonar repositories.
- Add license header plugin.
- Add pre-commit hook with husky & lint-staged


src/bootstrap/auth.ts
```diff
@@ -17,17 +17,14 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-
-// Auth module - OAuth flow and token management
```
damien-urruty-sonarsource, line 21: Why are you removing the header comment on all files? Just want to understand, I'm not in favor or against such comments, I don't use them much personally but maybe they can help someone

CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
...
+npm run lint        # ESLint with TypeScript-aware rules
+npm run lint:fix    # Auto-fix safe issues (also inserts license headers)
+npm run typecheck   # TypeScript type checking
+npm run validate    # Validate spec.yaml and command registration
+```
+
+## Testing
+
+```bash
+npm test                  # Unit tests
```
damien-urruty-sonarsource: This should be `bun test` there are probably other commands that should be adjusted

damien-urruty-sonarsource (approved):
I think it's a good start. Shouldn't we also enforce checks in the CI? I'm not sure having a pre-commit hook is enough

src/bootstrap/auth.ts
nquinquenel: The plugin header is removing everything that is non code after the header, due to `trailingNewlines: 1`. We can remove this line to allow comments at the top if needed.

Edit: done and reverted

CONTRIBUTING.md
nquinquenel: You're right - I created this file because it was complicated to understand what to do on this repo at first glance, and I thought it was relying on npm at first. We should probably move everything to bun, at least in the doc, however in the CI we cannot make fully use of bun so we're still stuck in mix of the two

nquinquenel:
> I think it's a good start. Shouldn't we also enforce checks in the CI? I'm not sure having a pre-commit hook is enough

I will first open another PR that fixes all the issues and then add the check to the CI, because for now it will be in error everytime you open a PR

PR #22, opened 2026-02-24
CLI-16 CLI-23 state file documentation & cleanup

sophio-japharidze-sonarsource:


src/commands/issues.ts
```diff
@@ -95,6 +95,7 @@ export async function issuesSearchCommand(options: IssuesSearchOptions): Promise
 
     const params: IssuesSearchParams = {
       projects: options.project,
+      organization: resolved.orgKey,
```
vnaskos-sonar, line 98: Would this work for SQS?

src/lib/config-constants.ts
```diff
@@ -38,8 +38,8 @@ export const APP_NAME = 'sonarqube-cli';
 // CLI data directory
 // ---------------------------------------------------------------------------
 
-/** Root directory for all CLI data: ~/.sonarqube-cli (override via SONAR_CLI_DIR env var in test environment) */
-export const CLI_DIR = process.env.SONAR_CLI_DIR ?? join(homedir(), `.${APP_NAME}`);
+/** Root directory for all CLI data: ~/.sonar/sonarqube-cli (override via SONAR_CLI_DIR env var in test environment) */
+export const CLI_DIR = process.env.SONAR_CLI_DIR ?? join(homedir(), '.sonar', `${APP_NAME}`);
```
vnaskos-sonar, line 42: The install script currently installs the sonar cli on `~/.local/share/sonarqube-cli/bin` (recommended path for linux and macos). It also doesn't set the `SONAR_CLI_DIR`.

However, all the configuration is on the `~/.sonar/sonarqube-cli`. It looks like there are two CLI_DIRs.

Not sure what's the best way to proceed. The install scripts could change to the `~/.sonar/sonarqube-cli/bin`, this also means that sonar-secrets will be available on the PATH as well. Another option would be to have `CLI_INSTALL_DIR=~/.local/share/sonarqube-cli` and `CLI_CONF_DIR=~/.sonar/sonarqube-cli` for clarity.

build-scripts/setup.sh
```diff
@@ -96,7 +96,7 @@ echo "🔧 Configuring shell..."
 # Detect shell and config file
 SHELL_NAME="$(basename "$SHELL")"
 SHELL_CONFIG=""
-PATH_EXPORT="export PATH=\"\$HOME/.sonarqube-cli/bin:\$PATH\""
+PATH_EXPORT="export PATH=\"\$HOME/.sonar/sonarqube-cli/bin:\$PATH\""
```
vnaskos-sonar: Currently install script uses a different path `~/.local/share/sonarqube-cli`. Should this be aligned? 🤔 

The setup.sh is used for convenience of local development, we could keep it a use a different path and executable name (eg. sonar-dev).

src/commands/issues.ts
sophio-japharidze-sonarsource: Yes, it's an optional parameter.

build-scripts/setup.sh
sophio-japharidze-sonarsource: I will delete this script

src/lib/config-constants.ts
sophio-japharidze-sonarsource: Let's keep two dirs - one for the installation because it will be in the PATH automatically, and another one for everything else.

vnaskos-sonar (approved):

PR #23, opened 2026-02-24
CLI-32 Use non standard exit code for secret detection to avoid errors block agent

kirill-knize-sonarsource:
I have also cleaned up sonar verify hooks, that are leftovers form `sonar verify` A3S based Demos.

### DIFF
--- a/src/bootstrap/hook-templates.ts
+++ b/src/bootstrap/hook-templates.ts
@@ -1,244 +1,6 @@
 // Hook script templates for Claude Code integration
 
 /**
- * Unix hook template for prompt-based analysis (bash)
- */
-export function getHookPromptTemplateUnix(): string {
-  return String.raw`#!/bin/bash
-# SonarQube Analysis Hook for Claude Code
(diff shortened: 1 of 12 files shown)


src/bootstrap/hook-templates.ts
```diff
@@ -269,10 +31,10 @@ if [ -z "$file_path" ] || [ ! -f "$file_path" ]; then
 fi
 
 # Scan file for secrets
-sonar secret check --file "$file_path" > /dev/null 2>&1
+sonar analyze --file "$file_path" > /dev/null 2>&1
```
damien-urruty-sonarsource, line 34: Isn't it `sonar analyze secrets`?

damien-urruty-sonarsource (approved):

src/bootstrap/hook-templates.ts
kirill-knize-sonarsource: It's a shortcut for it, yes. It's the same now. As `sonar auth login` is the same as `sonar auth`. TBD it's what we want, but for now it's the case.

PR #25, opened 2026-02-24
CLI-35 Ensure spec validation is called on build, remove outdated commands, fix outdated messages

kirill-knize-sonarsource:
…mands, fix outdated messages


fpoulin (approved):
LGTM

PR #32, opened 2026-02-25
CLI-30 verify sonar-secrets signature upon installation

sophio-japharidze-sonarsource:
* Embeds SonarSource public signing key inside `src/lib/signatures.ts`
* Runs `npm run fetch:signatures` as part of build action & updates `src/lib/signatures.ts` with the fetched signatures
* Pins external dependency versions inside `package.json`
* During `sonar install secrets`, will compare the previously fetched signature against the downloaded binary signature
* Will delete downloaded file if signature check fails


build-scripts/fetch-signatures.js
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env node
```
vnaskos-sonar: This is probably not needed since we run this file using `bun run fetch:signatures`.

build-scripts/fetch-signatures.js
```diff
@@ -0,0 +1,148 @@
...
+  console.log(`Patched ${outputPath}`);
+}
+
+/** Returns { platform, armoredSignature } if distributed, null if skipped. */
+async function fetchAndVerifySignature(platform, version, distPrefix, verificationKey) {
+  const binaryName = distPrefix.split('/').at(-1);
+  const filename = `${binaryName}-${version}-${platform.os}-${platform.arch}.exe`;
+  const ascUrl = `${SONARSOURCE_BINARIES_URL}/${distPrefix}/${filename}.asc`;
+
+  process.stdout.write(`  ${platform.os}-${platform.arch} … `);
```
vnaskos-sonar:
```suggestion
  console.log(`  ${platform.os}-${platform.arch} … `);
```

build-scripts/fetch-signatures.js
```diff
@@ -0,0 +1,148 @@
...
+  let totalFailures = 0;
+
+  for (const [binaryName, { version, binaryPath, platforms }] of Object.entries(pkg.externalBinaries)) {
+    console.log(`Fetching signatures for ${binaryName} ${version}\n`);
+
+    const results = await Promise.allSettled(
+      platforms.map(platform => fetchAndVerifySignature(platform, version, binaryPath, verificationKey))
+    );
+
+    let failures = 0;
```
vnaskos-sonar: The logic in the script seems to be working, it's quite complicated, but can be refactored later 😕

vnaskos-sonar (approved):

build-scripts/fetch-signatures.js
sophio-japharidze-sonarsource: I have pushed the simplification

PR #34, opened 2026-02-25
CLI-36 update package version with build number; get rid of version.ts

sophio-japharidze-sonarsource:


.github/workflows/build.yml
```diff
@@ -40,13 +40,24 @@ jobs:
       - uses: SonarSource/ci-github-actions/get-build-number@148774f456203f228b7bd1bd68ed0c22254d9cd1
         id: get-build-number
 
+      - name: Update package.json version with build number
+        env:
+          BUILD_NUMBER: ${{ steps.get-build-number.outputs.BUILD_NUMBER }}
+        run: |
```
sophio-japharidze-sonarsource, line 46: cannot call `npm version ...` because `M.m.p.buildNumber` does not comply with SemVer

src/commands/auth.ts
```diff
@@ -241,7 +241,7 @@ export async function authLoginCommand(options: {
 
     await saveToken(server, token, org);
 
-    const state = loadState(VERSION);
+    const state = loadState(CLI_VERSION);
```
damien-urruty-sonarsource: A bit unrelated, but I noticed everywhere we call loadState, we pass the version. Couldn't we read it from loadState directly? Maybe a separate thing

damien-urruty-sonarsource (approved):
LGTM

PR #47, opened 2026-02-26
Fix shelldre:S7688 on hook-templates by replacing [] test command with [[]]

jonas-wielage-sonarsource:


src/bootstrap/hook-templates.ts
jonas-wielage-sonarsource, line 1: Changed my hooks locally to have the same format, it still works :)

sophio-japharidze-sonarsource (approved):

PR #49, opened 2026-02-26
CLI-3 Add telemetry for CLI

damien-urruty-sonarsource:


spec.yaml
```diff
@@ -272,3 +272,26 @@ commands:
 
           - command: cat .env | sonar analyze secrets --stdin
             description: Scan stdin for hardcoded secrets
+
+  # Configure CLI
```
vnaskos-sonar, line 285: missing README update 😉 

```sh
bun run ./build-scripts/generate-docs.js
```

tests/unit/configure.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      .filter(c => c.method === 'success')
+      .map(c => String(c.args[0]));
+    expect(messages.some(m => m.includes('disabled'))).toBe(true);
+  });
+
+  // -------------------------------------------------------------------------
+  // Validation: neither flag
+  // -------------------------------------------------------------------------
+
+  it('throws InvalidOptionError when neither --enabled nor --disabled is passed', async () => {
```
vnaskos-sonar: Currently I think the only way for a user to know if telemetry is enabled is manually open the state file.

In this case, should it report the status of telemetry 🤔

src/telemetry/user.ts
```diff
@@ -0,0 +1,61 @@
...
+  // Fast path — file already exists.
+  try {
+    return readFileSync(USER_FILE, 'utf-8').trim();
+  } catch {
+    // Fall through to create.
+  }
+
+  const id = randomUUID();
+  try {
+    // O_CREAT | O_EXCL guarantees that only one process creates the file.
```
vnaskos-sonar: Maybe I am missing something, in the file `(O_CREAT | O_EXCL)` are only mentioned in the comments.

Not sure if these comment are still relevant.

src/telemetry/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+import {randomUUID} from 'node:crypto';
+import {getActiveConnection, loadState, saveState,} from '../lib/state-manager.js';
+import {StoredTelemetryEvent, TelemetryEventPayload} from '../lib/state.js';
+import {getOrCreateUserId} from './user.js';
+
+export const TELEMETRY_FLUSH_MODE_ENV = '__SQ_CLI_TELEMETRY_FLUSH__';
+
+const TELEMETRY_ENDPOINT = 'https://events.sonardata.io/cli';
+const TELEMETRY_API_KEY = 'hJPRohLsOsasZeOhSCSNDiL4h2yR96S5fOWJqRch';
```
vnaskos-sonar, line 30: The API key could be stored in Vault, and get injected only during build time to avoid having it available in source code.

src/index.ts
```diff
@@ -168,6 +170,35 @@ analyze
...
+
+// Hidden flush command — only registered when running as a telemetry worker.
+if (process.env[TELEMETRY_FLUSH_MODE_ENV]) {
+  program
+      .command('flush-telemetry', { hidden: true })
+      .action(flushTelemetry);
+}
+
+// Collect a telemetry event after every command action.
+program.hook('postAction', async (_thisCommand, actionCommand) => {
```
vnaskos-sonar, line 178: From my understanding we always spawn a new process for telemetry. And if telemetry is disabled, it doesn't send any data just exits. Should this be wrapped in an `if (telemetry.enabled)` to stop spawning the new process if it's not necessary?

src/telemetry/index.ts
damien-urruty-sonarsource: We do it this way for SLCORE too, I double checked with the Data team and they are OK to follow the same approach. What you propose would add another layer of security, but in the end the secret would still be there somewhere in the executable. I don't think it adds much value. I will track it as a minor improvement ticket if that's OK with you?

src/index.ts
damien-urruty-sonarsource: This is done at the beginning of storeEvent, that is called at the end of this method. I could move it a bit earlier to avoid some computations, but in any case the process will not be spawned

spec.yaml
damien-urruty-sonarsource: Good catch! Fixed

src/telemetry/user.ts
damien-urruty-sonarsource: The explain what happens at low-level, but no they are probably not relevant. Removed

tests/unit/configure.test.ts
damien-urruty-sonarsource: This is more a functional/UX question. I will ask Clifford about it

src/telemetry/index.ts
damien-urruty-sonarsource: This is the message from Martin the other day:
> I think that's what you did for the IDE. I hope we don't think we need to over engineer this. If someone wants to taint the data they'd find a way. The telemetry receiver for SQIde and SQS is key-less and so far there hasn't been targeted abuse, so I guess we're good

I suggest we keep this approach

tests/unit/configure.test.ts
cliffordgoh-sonar: @vnaskos-sonar @damien-urruty-sonarsource 

I recommend the below. I think this is same as Vasileos suggestion
Command | Action | Output
-- | -- | --
sonar config telemetry --enabled | Enable telemetry | Telemetry enabled.
sonar config telemetry --disabled | Disables telemetry | Telemetry disabled.
sonar config telemetry | Returns the current status | Telemetry is currently: [enabled/disabled]

damien-urruty-sonarsource: Implemented as discussed

src/index.ts
damien-urruty-sonarsource: I moved the computation of the command to have the telemetry enablement check earlier

vnaskos-sonar (approved):

PR #55, opened 2026-02-27
CLI-52 Fix typos in help and error messages

kirill-knize-sonarsource:
- Fix wrong command hints: 'sonar secret install' → 'sonar install secrets'
- Fix wrong command hints: 'sonar secret status' → 'sonar install secrets --status'
- Fix auth error message: replace env var instructions with 'sonar auth login'
- Fix 'Press any key' prompt: rename to pressEnterKeyPrompt, implement raw stdin so only Enter advances and all other keys are silently consumed
- Fix misleading 404 error: map HTTP 403/404 to 'Access denied' message (SonarQube servers return 404 in place of 403 for auth failures)

### DIFF
--- a/src/sonarqube/client.ts
+++ b/src/sonarqube/client.ts
@@ -62,7 +62,7 @@ export class SonarQubeClient {
     if (!response.ok) {
       if (response.status === HTTP_STATUS_FORBIDDEN || response.status === HTTP_STATUS_NOT_FOUND) {
         throw new Error(
-          `Access denied (HTTP ${response.status}). Check your token and organization permissions.`,
+          `Access denied (HTTP ${response.status}). Check that the supplied token and organization are valid.`,
         );
       }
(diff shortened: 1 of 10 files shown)


src/sonarqube/client.ts
```diff
@@ -58,6 +60,11 @@ export class SonarQubeClient {
     });
 
     if (!response.ok) {
+      if (response.status === HTTP_STATUS_FORBIDDEN || response.status === HTTP_STATUS_NOT_FOUND) {
+        throw new Error(
+          `Access denied (HTTP ${response.status}). Check your token and organization permissions.`,
```
damien-urruty-sonarsource: > Check your token and organization permissions

The wording is a bit strange. Maybe:

> Check that the supplied token and organization are valid

damien-urruty-sonarsource (approved):

src/sonarqube/client.ts
kirill-knize-sonarsource: Changed

PR #58, opened 2026-03-02
CLI-33 Generate the documentation from the source code

damien-urruty-sonarsource:


build-scripts/generate-docs.ts
```diff
@@ -0,0 +1,139 @@
...
+const ROOT = join(__dirname, '..');
+
+// Wide characters used in option tables (display as 2 columns in monospace fonts)
+const WIDE_CHARS = new Set(['Yes', 'No']);
+
+/**
+ * Visual display width of a string in monospace fonts.
+ * Characters in WIDE_CHARS count as 2 columns, all others as 1.
+ */
+function displayWidth(str: string): number {
```
kirill-knize-sonarsource, line 47: This is taken from previous generator. There's a mistake in this check. We need to check like so:
```
if (WIDE_CHARS.has(str)) return 2;
return str.length;
```

kirill-knize-sonarsource (approved):
Minor error in doc generator. Else LGTM.

build-scripts/generate-docs.ts
damien-urruty-sonarsource: I don't see any problem with the rendering. Let's keep that for later

PR #67, opened 2026-03-03
CLI-75 Fix page-size options

damien-urruty-sonarsource:


src/cli/commands/list.ts
```diff
@@ -65,13 +63,9 @@ export async function listIssues(options: ListIssuesOptions): Promise<void> {
...
-  if (options.pageSize !== undefined) {
-    const ps = options.pageSize;
-    if (!Number.isInteger(ps) || ps < 1 || ps > DEFAULT_PAGE_SIZE) {
-      throw new Error(
-        `Invalid --page-size: '${options.pageSize}'. Must be an integer between 1 and 500`,
-      );
-    }
+  const ps = options.pageSize;
+  if (ps < 1 || ps > MAX_PAGE_SIZE) {
+    throw new Error(`Invalid --page-size option: '${ps}'. Must be an integer between 1 and 500`);
```
damien-urruty-sonarsource, line 68: Ideally I would like to move such checks upper in the command-tree. But I would need some integration tests to validate the behavior, keeping them here for now

tests/unit/projects-command.test.ts
```diff
@@ -210,33 +237,12 @@ describe('projectsSearchCommand', () => {
         return makeProjectsResponse([]);
       });
 
-      await listProjects({ pageSize: 50 });
-
-      expect(capturedParams?.ps).toBe(50);
-    });
-
-    it('defaults to page 1 when page option is not provided', async () => {
```
damien-urruty-sonarsource, line 218: Similarly, this should be tested at an upper layer, removed for now as the method always receives a correct argument

src/cli/command-tree.ts
kirill-knize-sonarsource: Re-export chain end.

src/cli/commands/list.ts
```diff
@@ -194,3 +193,5 @@ export async function listProjects(options: ListProjectsOptions): Promise<void>
     }),
   );
 }
+
+export { MAX_PAGE_SIZE as DEFAULT_PAGE_SIZE } from '../../sonarqube/projects';
```
kirill-knize-sonarsource: Re-export chain

src/cli/commands/list.ts
```diff
@@ -163,20 +157,25 @@ export async function listProjects(options: ListProjectsOptions): Promise<void>
...
-  if (pageSize <= 0 || pageSize > MAX_PAGE_SIZE) {
+  const pageSize = options.pageSize;
+  if (pageSize < 1 || pageSize > MAX_PAGE_SIZE) {
     throw new Error(
       `--page-size must be greater than 0 and less than or equal to ${MAX_PAGE_SIZE}`,
     );
   }
 
+  const page = options.page;
+  if (page < 1) {
```
kirill-knize-sonarsource, line 173: Shall we check the page number for `list issues` same way as well?

src/cli/commands/list.ts
kirill-knize-sonarsource, line 68: Message is inconsistent with listProjects: "--page-size must be greater than 0 and...".
This one is better, I think listProjects message shall be aligned.

kirill-knize-sonarsource (approved):
Some non critical suggestions.

src/cli/commands/list.ts
damien-urruty-sonarsource: Good catch

PR #68, opened 2026-03-03
Remove token from interactive login command

andreaguarino:
While following the documentation, it looks like the command below had a typo, i.e. if the login is interactive we shouldn't need to put the token

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ sonar auth login
 Interactive login for SonarCloud with browser
 
 ```bash
-sonar auth login -o my-org -t squ_abc123
+sonar auth login -o my-org
 ```
 Non-interactive login with direct token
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ sonar auth login
 Interactive login for SonarCloud with browser
 
 ```bash
-sonar auth login -o my-org -t squ_abc123
+sonar auth login -o my-org
```
vnaskos-sonar, line 45: This part of the README is autogenerated, and it's layout may be confusing at the moment. The description of the command is at the bottom.

Layout:
```
`<sonar command>`
<description>
```

This is actually the non-interactive example. I understand the confusion and we note it, to switch the layout.

PR #70, opened 2026-03-04
CLI-88 Use GitHub releases to fetch latest version

vnaskos-sonar:
Replace hardcoded version with dynamic resolution in install scripts
                                                                                                                                                                                                                                                                                                                    
The install scripts (install.sh, install.ps1) previously hardcoded the latest version number as a workaround after the `latest-version.txt` approach failed.                                                                                                                                     

Now that the project is public on GitHub, the scripts resolve the latest version dynamically by following the GitHub releases redirect (/releases/latest → /releases/tag/VERSION) and extracting the version from the final URL. Alternatively we could use the GitHub API, however, it has the drawbacks of requiring API token after the unauthenticated rate limit is reached which makes it less ideal.

The `latest-version.txt` generation step has also been removed from the build pipeline as it is no longer used.

### DIFF
--- a/user-scripts/install.ps1
+++ b/user-scripts/install.ps1
@@ -10,8 +10,14 @@ $BaseUrl    = 'https://binaries.sonarsource.com/Distribution/sonarqube-cli'
 $Platform   = 'windows-x86-64'
 
 function Resolve-LatestVersion {
-    $Version = (Invoke-WebRequest -Uri "$BaseUrl/latest-version.txt" -UseBasicParsing).Content.Trim()
-    if (-not $Version) {
+    $ReleasesUrl = 'https://github.com/SonarSource/sonarqube-cli/releases/latest'
+    $Request = [System.Net.WebRequest]::Create($ReleasesUrl)
+    $Request.AllowAutoRedirect = $true
+    $Response = $Request.GetResponse()
+    $FinalUrl = $Response.ResponseUri.AbsoluteUri
+    $Response.Close()
+    $Version = ($FinalUrl -split '/')[-1] -replace '^v', ''
+    if (-not $Version -or $Version -eq 'latest') {
         Write-Error 'Could not determine the latest version.'
         exit 1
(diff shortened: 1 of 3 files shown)


user-scripts/install.ps1
```diff
@@ -10,8 +10,14 @@ $BaseUrl    = 'https://binaries.sonarsource.com/Distribution/sonarqube-cli'
 $Platform   = 'windows-x86-64'
 
 function Resolve-LatestVersion {
-    $Version = (Invoke-WebRequest -Uri "$BaseUrl/latest-version.txt" -UseBasicParsing).Content.Trim()
-    if (-not $Version) {
+    $ReleasesUrl = 'https://github.com/SonarSource/sonarqube-cli/releases/latest'
```
damien-urruty-sonarsource, line 13: I don't think this can work reliably. In our release process, the GH release is created at some point, and the artifacts are published on binaries a bit later. So there is a short period of time where artifact lookup from the latest GH release will fail for users trying to install. It's even worse if the GH release creation works but the actual release fails for any reason, as it wouldn't be temporary

vnaskos-sonar: For projects without the release automation, the GitHub Release remains as Draft until the Release pipeline finishes successfully, in case the pipeline fails the GitHub Release stays as a draft.

With the release automation, isn't it the same? Does the GitHub Release becomes public (not Draft)?

`https://github.com/SonarSource/sonarqube-cli/releases/latest` excludes Draft releases.

damien-urruty-sonarsource: I think you're right, but the approach is fragile. We couple the ability for users to install our tool to some details of our release pipeline. I would be reluctant to use GH as the install gateway. Another reason is: what if it's down (as it happens at least every month)? It would block our users from installing (or updating when we support it). Using GitHub was fine at the beginning to bootstrap the distribution, but I think we need to find a better approach

PR #79, opened 2026-03-05
CLI-109 Build all artifacts from a single runner

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Won't do, see https://sonarsource.atlassian.net/browse/CLI-13?focusedCommentId=877697

PR #85, opened 2026-03-09
CLI-120 Implement command to self-update

vnaskos-sonar:
Create a new command `sonar self-update` to update the sonar cli itself.

# How
* The self-update is checking the version from the latest install.ps1 or install.sh on Github (the same that the users will use to fresh install)
* If the current version is lower than the fetched then calls the install.ps1 or install.sh and the cli exits
* The install script is working as usual (no change)
* On Windows the install scripts runs on a new terminal window so that the progress is visible to the users (no need for that on linux, it already works on the same terminal)

# Options
* `--status` will report if there is a new version
* `--force` will proceed to call the install script, no matter what the current version is

### DIFF
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -181,6 +181,7 @@ configure
 COMMAND_TREE.command('self-update')
   .description('Update sonar CLI to the latest version')
   .option('--status', 'Check for a newer version without installing')
+  .option('--force', 'Install the latest version even if already up to date')
   .action((options: SelfUpdateOptions) => runCommand(() => selfUpdate(options)));
 
 // Hidden flush command — only registered when running as a telemetry worker.
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -33,7 +33,7 @@ import { integrateClaude, type IntegrateClaudeOptions } from './commands/integra
 import { analyzeSecrets, type AnalyzeSecretsOptions } from './commands/analyze/secrets';
 import { flushTelemetry, storeEvent, TELEMETRY_FLUSH_MODE_ENV } from '../telemetry';
 import { configureTelemetry, type ConfigureTelemetryOptions } from './commands/config/telemetry';
-import { selfUpdate, type SelfUpdateOptions } from './commands/self-update';
+import { selfUpdate, type SelfUpdateOptions } from './commands/self-update/self-update';
 import { parseInteger } from './commands/_common/parsing';
 import { MAX_PAGE_SIZE } from '../sonarqube/projects';
 
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -33,6 +33,7 @@ import { integrateClaude, type IntegrateClaudeOptions } from './commands/integra
 import { analyzeSecrets, type AnalyzeSecretsOptions } from './commands/analyze/secrets';
 import { flushTelemetry, storeEvent, TELEMETRY_FLUSH_MODE_ENV } from '../telemetry';
 import { configureTelemetry, type ConfigureTelemetryOptions } from './commands/config/telemetry';
+import { selfUpdate } from './commands/self-update';
 import { parseInteger } from './commands/_common/parsing';
 import { MAX_PAGE_SIZE } from '../sonarqube/projects';
 
@@ -175,6 +176,11 @@ configure
   .option('--enabled', 'Enable collection of anonymous usage statistics')
   .option('--disabled', 'Disable collection of anonymous usage statistics')
   .action((options: ConfigureTelemetryOptions) => runCommand(() => configureTelemetry(options)));
+
+// Update the CLI to the latest version
+COMMAND_TREE.command('self-update')
+  .description('Update sonar CLI to the latest version')
+  .action(() => runCommand(() => selfUpdate()));
 
 // Hidden flush command — only registered when running as a telemetry worker.
 if (process.env[TELEMETRY_FLUSH_MODE_ENV]) {
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -33,7 +33,7 @@ import { integrateClaude, type IntegrateClaudeOptions } from './commands/integra
 import { analyzeSecrets, type AnalyzeSecretsOptions } from './commands/analyze/secrets';
 import { flushTelemetry, storeEvent, TELEMETRY_FLUSH_MODE_ENV } from '../telemetry';
 import { configureTelemetry, type ConfigureTelemetryOptions } from './commands/config/telemetry';
-import { selfUpdate } from './commands/self-update';
+import { selfUpdate, type SelfUpdateOptions } from './commands/self-update';
 import { parseInteger } from './commands/_common/parsing';
 import { MAX_PAGE_SIZE } from '../sonarqube/projects';
 
@@ -180,7 +180,8 @@ configure
 // Update the CLI to the latest version
 COMMAND_TREE.command('self-update')
   .description('Update sonar CLI to the latest version')
-  .action(() => runCommand(() => selfUpdate()));
+  .option('--status', 'Check for a newer version without installing')
+  .action((options: SelfUpdateOptions) => runCommand(() => selfUpdate(options)));
 
 // Hidden flush command — only registered when running as a telemetry worker.
 if (process.env[TELEMETRY_FLUSH_MODE_ENV]) {
--- a/src/cli/commands/self-update.ts
+++ b/src/cli/commands/self-update.ts
@@ -49,7 +49,8 @@ export function extractVersion(scriptContent: string): string | null {
  * the CLI version from package.json only has three segments ("0.5.0").
  */
 export function stripBuildNumber(version: string): string {
-  return version.split('.').slice(0, 3).join('.');
+  const SEMVER_SEGMENTS = 3;
+  return version.split('.').slice(0, SEMVER_SEGMENTS).join('.');
 }
 
 /** Returns true when `candidate` is strictly newer than `current` (semver, numeric comparison). */
--- a/src/cli/commands/self-update.ts
+++ b/src/cli/commands/self-update.ts
@@ -97,6 +97,7 @@ export async function checkForUpdate(): Promise<UpdateCheckResult> {
 
 export interface SelfUpdateOptions {
   status?: boolean;
+  force?: boolean;
 }
 
 async function selfUpdateStatus(): Promise<void> {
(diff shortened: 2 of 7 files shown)


src/cli/commands/self-update/self-update.ts
```diff
@@ -0,0 +1,180 @@
...
+import { info, success, warn, text, blank } from '../../ui';
+
+const VERSION_PATTERNS = [
+  // Shell:       version="1.2.3"  or  version='1.2.3'
+  /\bversion\s*=\s*["'](\d+\.\d+\.\d+(?:\.\d+)?)["']/,
+  // PowerShell:  $SonarVersion = "1.2.3"
+  /\$SonarVersion\s*=\s*["'](\d+\.\d+\.\d+(?:\.\d+)?)["']/i,
+];
+
+/** Extract the pinned version from an install script. Returns null if not found. */
```
vnaskos-sonar, line 36: Not sure if this is the best way to find the latest version 😕

src/cli/command-tree.ts
```diff
@@ -184,6 +185,13 @@ configure
   .option('--disabled', 'Disable collection of anonymous usage statistics')
   .action((options: ConfigureTelemetryOptions) => runCommand(() => configureTelemetry(options)));
 
+// Update the CLI to the latest version
+COMMAND_TREE.command('self-update')
+  .description('Update sonar CLI to the latest version')
+  .option('--status', 'Check for a newer version without installing')
+  .option('--force', 'Install the latest version even if already up to date')
```
damien-urruty-sonarsource, line 184: Is this option like a re-install or a repair? If it's the case, then `self-update --force` does not make it obvious. Otherwise I don't understand the purpose

src/cli/commands/self-update/self-update.ts
```diff
@@ -0,0 +1,180 @@
...
+    const match = pattern.exec(scriptContent);
+    if (match) return match[1];
+  }
+  return null;
+}
+
+/**
+ * Strips the build number (4th segment) from a version string.
+ * The install script version may include a build number (e.g. "0.5.0.241") while
+ * the CLI version from package.json only has three segments ("0.5.0").
```
damien-urruty-sonarsource, line 48: For artifacts that will be installed in the wild, the version in the package.json does include the build number, it's injected from the CI. Maybe that doesn't change anything

src/cli/commands/self-update/self-update.ts
```diff
@@ -0,0 +1,180 @@
+/*
```
damien-urruty-sonarsource, line 1: This file would need to be moved to a dedicated folder now, could you rebase and mimic other commands?

src/cli/commands/self-update.ts
```diff
@@ -0,0 +1,180 @@
...
+    info('Starting update in a new terminal window...');
+    // The ComSpec environment variable (always points to the system cmd.exe)
+    const cmdExe = process.env.ComSpec ?? String.raw`C:\Windows\System32\cmd.exe`;
+    const child = spawn(
+      cmdExe,
+      ['/c', 'start', 'powershell', '-NoExit', '-ExecutionPolicy', 'Bypass', '-File', tempPath],
+      { detached: true, stdio: 'ignore' },
+    );
+    child.unref();
+    throw new CommandFailedError('', 0);
```
damien-urruty-sonarsource: Why are we throwing from here? This looks a bit strange

src/cli/commands/self-update/self-update.ts
```diff
@@ -0,0 +1,180 @@
...
+  const tempPath = join(tmpdir(), scriptName);
+
+  if (process.platform === 'win32') {
+    // On Windows the running binary is file-locked, so the parent must exit
+    // before the script can overwrite it. Open PowerShell in a new window so
+    // it has its own console and the user can see the output.
+    writeFileSync(tempPath, scriptContent, 'utf8');
+    info('Starting update in a new terminal window...');
+    // The ComSpec environment variable (always points to the system cmd.exe)
+    const cmdExe = process.env.ComSpec ?? String.raw`C:\Windows\System32\cmd.exe`;
```
damien-urruty-sonarsource, line 163: Are we sure the Windows installation is always on the C drive? Couldn't we figure this out by other means? Or is it too unlikely that `ComSpec` is missing?

src/cli/commands/self-update/self-update.ts
```diff
@@ -0,0 +1,180 @@
...
+    info(`Updating v${currentVersion} → v${latestVersion}...`);
+  } else {
+    info(`Force installing v${latestVersion}...`);
+  }
+
+  const tempPath = join(tmpdir(), scriptName);
+
+  if (process.platform === 'win32') {
+    // On Windows the running binary is file-locked, so the parent must exit
+    // before the script can overwrite it. Open PowerShell in a new window so
```
damien-urruty-sonarsource: Is it really before? If so, how can we make sure the current program finished before the new one is started? Could there be a race condition where the other process is faster than the termination of this program?

src/lib/config-constants.ts
```diff
@@ -68,6 +68,8 @@ export const BIN_DIR = join(CLI_DIR, 'bin');
 export const SONARSOURCE_BINARIES_URL =
   process.env.SONAR_CLI_BINARIES_URL ?? 'https://binaries.sonarsource.com';
 export const SONAR_SECRETS_DIST_PREFIX = 'CommercialDistribution/sonar-secrets';
+export const UPDATE_SCRIPT_BASE_URL =
+  'https://raw.githubusercontent.com/SonarSource/sonarqube-cli/refs/heads/master/user-scripts';
```
damien-urruty-sonarsource, line 72: We discussed that earlier, I'm not a big fan of relying on GitHub to check for updates, but for now we probably don't have a better way

CLAUDE.md
```diff
@@ -25,6 +25,19 @@ Each command lives in `src/cli/commands/`. The command tree is defined in `src/c
...
+|---------|------|-------------|
+| `sonar install secrets` | `commands/install.ts` | Install sonar-secrets binary |
+| `sonar integrate claude` | `commands/integrate.ts` | Setup SonarQube integration for Claude Code |
+| `sonar list issues` | `commands/list.ts` | List issues from SonarQube |
+| `sonar list projects` | `commands/list.ts` | List projects from SonarQube |
+| `sonar auth login/logout/purge/status` | `commands/auth.ts` | Manage authentication tokens |
+| `sonar analyze secrets` | `commands/analyze.ts` | Analyze code for hardcoded secrets |
+| `sonar config telemetry` | `commands/config.ts` | Configure telemetry settings |
+| `sonar self-update` | `commands/self-update.ts` | Update sonar CLI to the latest version |
+
```
damien-urruty-sonarsource: I am a bit reluctant to have this here because we will very likely forget to update it in the future. Is it really necessary? From my experience Claude was able to navigate the codebase easily already

src/cli/command-tree.ts
```diff
@@ -184,6 +185,13 @@ configure
   .option('--disabled', 'Disable collection of anonymous usage statistics')
   .action((options: ConfigureTelemetryOptions) => runCommand(() => configureTelemetry(options)));
 
+// Update the CLI to the latest version
+COMMAND_TREE.command('self-update')
+  .description('Update sonar CLI to the latest version')
+  .option('--status', 'Check for a newer version without installing')
```
damien-urruty-sonarsource, line 183: It's not obvious that `self-update --status` checks if an update is available. We could also understand it as "let me check the status of a self-update that would be going on". What about `self-update --check`, `self-update --dry-run`, `self-update --up-to-date` ? I'm not good at naming things

damien-urruty-sonarsource (commented):
A first round of comments, it's mostly looking good to me. The current solution would not allow for self-update for dogfooding builds, how do we plan to achieve that? Also as I shared homebrew doesn't like self-update, so we might need a way to disable this command, but we can do that later

src/cli/commands/self-update.ts
vnaskos-sonar: It was confusion of not using the `process.exit(0)`, but actually it's not necessary at all.

src/cli/commands/self-update/self-update.ts
vnaskos-sonar: Originally I used `cmd` but it raises a PATH injection issue, so the other option is to rely on ComSpec which in most cases should be available and fallback to C drive. It should cover more than 90% of the cases.

src/cli/commands/self-update/self-update.ts
vnaskos-sonar: 1. The current running executable has to exit (which happens instantaneously) 
2. then the script is downloading the new version on a temp file 
3. and when it's completed downloading, it replaces the temp with the old version.

Problems could occur
* if 1. gets stuck and the running executable doesn't exit
    > there isn't any additional action between starting the script and the exit of the executable, that could cause it to hang.
* [race condition] the download was too fast, if 1. is still exiting
    > The download on Windows takes quite some time (mainly caused by the size of the artifact and the progress bar on the terminal)

So, generally should be safe and worst case scenario the user can retry to update.

src/cli/commands/self-update/self-update.ts
vnaskos-sonar: I think this is not true, package.json doesn't get the build number injected on CI (at least not at the moment). It's not a problem though, as new versions should not rely on the build number based on our processes.

src/cli/commands/self-update/self-update.ts
kirill-knize-sonarsource: Looks like an important point. Probably worth putting a comment that we reasonably expect that 1 will exit fast, but if not - race is possible.

kirill-knize-sonarsource (approved):
LGTM

src/cli/command-tree.ts
vnaskos-sonar: As discussed with Kirill, this will be addressed separately with UX involvement to unblock the dependent PRs.

src/cli/command-tree.ts
vnaskos-sonar: As discussed with Kirill, this will be addressed separately with UX involvement to unblock the dependent PRs

src/cli/commands/self-update/self-update.ts
damien-urruty-sonarsource: In theory it should be the case, we do change the version with the build number [here](https://github.com/SonarSource/sonarqube-cli/blob/master/.github/workflows/build.yml#L43), the only thing is I don't think this works now, because of some CI changes that happened recently. 

I don't know if we should fix that, or simply remove this step in the job

PR #87, opened 2026-03-09
CLI-123 Move commands to dedicated folders

damien-urruty-sonarsource:
Move the content of the bootstrap folder to more adapted places


kirill-knize-sonarsource (approved):
Great!

PR #88, opened 2026-03-09
CLI-122 Add support for Mac x64

damien-urruty-sonarsource:


sophio-japharidze-sonarsource (approved):
Do we need to fix something on secrets CLI installation side as well? Looks like it should work as-is but just to be sure

damien-urruty-sonarsource:
@sophio-japharidze-sonarsource You're right, the install script needs to be modified. Good catch, thank you!

damien-urruty-sonarsource:
@sophio-japharidze-sonarsource I updated the installation script, but I will probably ask Kevin to test it

sophio-japharidze-sonarsource (approved):

damien-urruty-sonarsource:
@sophio-japharidze-sonarsource I just understood your comment. I thought you were talking about our install script, that needed fixing, but you are talking about the Secrets CLI. I checked and indeed they don't publish Mac x64 on binaries. If this is really needed, we would need to request the security squad to add support for a new platform. I will check with @vnaskos-sonar if we can try to make a first publication without this platform

damien-urruty-sonarsource:
Closing for now. This PR is outdated, we don't do cross-compilation anymore in the CI since CLI-150. At the moment there is no Mac x64 available CI runner, so we cannot build. The same applies for secrets CLI, and on top of that there is no GraalVM Mac x64 support

PR #89, opened 2026-03-10
CLI-40 simplify org selection

sophio-japharidze-sonarsource:


src/cli/commands/auth/login.ts
```diff
@@ -143,6 +143,68 @@ async function getOrGenerateToken(
...
+  }
+
+  if (choice === MANUAL_ENTRY) {
+    const manualOrg = await textPrompt('Enter organization key');
+    if (manualOrg === null) {
+      throw new Error('Organization selection cancelled');
+    }
+    if (!manualOrg.trim()) {
+      throw new Error('Organization key is required');
+    }
```
vnaskos-sonar: I think this is technically the same case, but currently there are different messages 🤔

src/lib/config-constants.ts
```diff
@@ -73,9 +73,10 @@ export const SONAR_SECRETS_DIST_PREFIX = 'CommercialDistribution/sonar-secrets';
 // SonarCloud
 // ---------------------------------------------------------------------------
 
-export const SONARCLOUD_HOSTNAME = 'sonarcloud.io';
-export const SONARCLOUD_URL = 'https://sonarcloud.io';
-export const SONARCLOUD_API_URL = 'https://api.sonarcloud.io';
+export const SONARCLOUD_URL = process.env.SONARQUBE_CLI_SONARCLOUD_URL ?? 'https://sonarcloud.io';
+export const SONARCLOUD_HOSTNAME = new URL(SONARCLOUD_URL).hostname;
+export const SONARCLOUD_API_URL =
+  process.env.SONARQUBE_CLI_SONARCLOUD_API_URL ?? 'https://api.sonarcloud.io';
```
vnaskos-sonar: So far most of the sonarqube-cli env variables have the `SONAR_CLI_` prefix, we can follow the same here for consistency (Maybe even drop the _SONAR_ part from _SONARCLOUD_)


```suggestion
export const SONARCLOUD_URL = process.env.SONAR_CLI_CLOUD_URL ?? 'https://sonarcloud.io';
export const SONARCLOUD_HOSTNAME = new URL(SONARCLOUD_URL).hostname;
export const SONARCLOUD_API_URL =
  process.env.SONAR_CLI_CLOUD_API_URL ?? 'https://api.sonarcloud.io';
```

src/cli/commands/auth/login.ts
```diff
@@ -143,6 +143,68 @@ async function getOrGenerateToken(
...
+    { value: MANUAL_ENTRY, label: 'Enter organization key manually' },
+  ];
+
+  const choice = await selectPrompt<string>('Select an organization', orgOptions);
+  if (choice === null) {
+    throw new Error('Organization selection cancelled');
+  }
+
+  if (choice === MANUAL_ENTRY) {
+    const manualOrg = await textPrompt('Enter organization key');
```
vnaskos-sonar, line 201: There is a buggy behavior and I am not able to enter any characters on the manual organization

Steps to reproduce:
1. `sonar auth purge`
2. `sonar auth login`
3. 
```
 ›  Press Enter to open the browser
  ⏳  Waiting for authorization... or paste token and press Enter:
  › █
```
4. Generate the token as usual on the browser
5. Select to enter an organization manually
6. Type and nothing happens

On step 3. if I click enter while the browser window is still pending I get the following
```
URL: https://sonarcloud.io/auth?product=cli&port=64121
  ›  Press Enter to open the browser
  ✓  Token accepted
```
and then I can type manually my organization

src/cli/commands/auth/login.ts
```diff
@@ -143,6 +143,68 @@ async function getOrGenerateToken(
...
+    if (manualOrg === null) {
+      throw new Error('Organization selection cancelled');
+    }
+    if (!manualOrg.trim()) {
+      throw new Error('Organization key is required');
+    }
+    return manualOrg.trim();
+  }
+
+  return choice;
```
vnaskos-sonar, line 208: When I enter a non-existing organization I still get successful message (eg. `hgfjhgfhj`)

src/sonarqube/client.ts
```diff
@@ -162,6 +162,25 @@ export class SonarQubeClient {
...
+   */
+  async listUserOrganizations(): Promise<{
+    organizations: Array<{ key: string; name: string }>;
+    total: number;
+  }> {
+    try {
+      const result = await this.get<{
+        organizations: Array<{ key: string; name: string }>;
+        paging: { total: number };
+      }>('/api/organizations/search', { member: true, ps: 100 });
```
vnaskos-sonar: Very minor comment:  It's not a common scenario to have a ton of organizations, and because we don't provide a search functionality we could reduce from 100 to maybe 10 to be manageable on the terminal screen 🤔

vnaskos-sonar (changes requested):

src/cli/commands/auth/login.ts
sophio-japharidze-sonarsource: I have created https://sonarsource.atlassian.net/browse/CLI-129 to track this separately.
It was already the case before :D

src/cli/commands/auth/login.ts
sophio-japharidze-sonarsource: I have pushed a fix that works but looks ugly

vnaskos-sonar (approved):

PR #90, opened 2026-03-10
CLI-121 Create mechanism for post-update actions

vnaskos-sonar:
On startup (index.ts) compare the cliVersion from the state.json with the current version. If the state exists and the current version is newer than the state then run some post-update actions.


src/cli/commands/self-update.ts
```diff
@@ -178,3 +155,5 @@ export async function selfUpdate(options: SelfUpdateOptions = {}): Promise<void>
     }
   }
 }
+
+export { isNewerVersion, stripBuildNumber } from '../../lib/version';
```
kirill-knize-sonarsource: I don't think we need this re-export. We import those form `lib/version` everywhere.

kirill-knize-sonarsource (approved):
One comment, else LGTM

PR #93, opened 2026-03-10
CLI-55 Forward logs from the secrets CLI to the user

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Closing for now, we will reconsider later

PR #95, opened 2026-03-11
CLI-130 fix installation on old macos

sophio-japharidze-sonarsource:


damien-urruty-sonarsource (approved):
LGTM, tested on Kevin's personal Mac. The install script fails later because arm64 is detected instead of x64, this is for another ticket, but the mktemp works now!

PR #99, opened 2026-03-12
CLI-9 integrate pre commit hook

sophio-japharidze-sonarsource:


src/cli/commands/integrate/git/git-husky.ts
```diff
@@ -0,0 +1,84 @@
...
+import { info, success } from '../../../../ui';
+import { HOOK_MARKER } from './git-scripts';
+import type { GitHookType } from '.';
+
+function getHuskyPreCommitSnippet(): string {
+  return String.raw`
+# ${HOOK_MARKER}
+FILES=$(git diff --cached --name-only --diff-filter=ACMR)
+if [ -n "$FILES" ]; then
+  CLEAN_PATH=$(echo "$PATH" | tr ':' '\n' | grep -v node_modules | tr '\n' ':')
```
sophio-japharidze-sonarsource: Image

src/cli/commands/integrate/git/git-shell-fragments.ts
```diff
@@ -0,0 +1,130 @@
...
+export function getHookScript(hook: GitHookType): string {
+  return hook === 'pre-commit' ? getPreCommitHookScript() : getPrePushHookScript();
+}
+
+// ─── Husky snippets ───────────────────────────────────────────────────────────
+// Fragments appended to existing .husky/pre-commit or .husky/pre-push files.
+// Husky prepends node_modules/.bin to PATH, so we strip those entries before
+// looking up `sonar` to avoid accidentally running a project-local package.
+
+export function getHuskyPreCommitSnippet(): string {
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: Will fix in a separate PR

src/cli/commands/analyze/secrets.ts
```diff
@@ -299,29 +299,30 @@ function displayScanResults(scanResult: {
...
   scanDurationMs: number,
   exitCode: number,
 ): void {
   blank();
-  error('Scan found secrets');
-  text(`  Exit code: ${exitCode}`);
-  text(`  Duration: ${scanDurationMs}ms`);
 
-  if (result.stderr) {
+  const output = [result.stderr, result.stdout].filter(Boolean).join('\n');
```
damien-urruty-sonarsource, line 307: What does this `filter(Boolean)` do?

src/cli/commands/analyze/secrets.ts
```diff
@@ -299,29 +299,30 @@ function displayScanResults(scanResult: {
...
+  if (exitCode === EXIT_CODE_SECRETS_FOUND) {
+    error(`Secrets found (${scanDurationMs}ms)`);
+    logger.error(`Secrets found, exit code: ${exitCode}`);
+    throw new CommandFailedError(`Secrets found (${scanDurationMs}ms)`, exitCode);
   }
-  blank();
-  throw new CommandFailedError(`Scan failed with exit code: ${exitCode}`, exitCode);
+
+  error(`Scan error (exit code ${exitCode}, ${scanDurationMs}ms)`);
+  logger.error(`Scan failed with exit code: ${exitCode}`);
```
damien-urruty-sonarsource: This should already managed by the global catch in run-command.ts

src/cli/commands/analyze/secrets.ts
```diff
@@ -299,29 +299,30 @@ function displayScanResults(scanResult: {
...
-    print(result.stderr);
   }
 
-  if (result.stdout) {
-    blank();
-    text('Output:');
-    print(result.stdout);
+  if (exitCode === EXIT_CODE_SECRETS_FOUND) {
+    error(`Secrets found (${scanDurationMs}ms)`);
+    logger.error(`Secrets found, exit code: ${exitCode}`);
```
damien-urruty-sonarsource: Same

src/cli/commands/integrate/git/index.ts
```diff
@@ -0,0 +1,368 @@
...
+  );
+  const usePreCommitConfig = existsSync(join(projectInfo.rootDir, PRE_COMMIT_CONFIG_FILE));
+
+  const hook = await resolveHookType(options);
+  text(`Hook: ${hook}`);
+  blank();
+
+  await ensureSonarSecrets();
+
+  const huskyHookPath = join(projectInfo.rootDir, '.husky', hook);
```
damien-urruty-sonarsource: Part of this path is accessed above, you could maybe use a local variable

src/cli/commands/integrate/git/git-husky.ts
```diff
@@ -0,0 +1,42 @@
...
+import { HOOK_MARKER, getHuskySnippet } from './git-shell-fragments';
+import type { GitHookType } from '.';
+
+export async function installViaHusky(huskyHookPath: string, hook: GitHookType): Promise<void> {
+  const fs = await import('node:fs/promises');
+  let content: string;
+  try {
+    content = await fs.readFile(huskyHookPath, 'utf-8');
+  } catch {
+    content = '';
```
damien-urruty-sonarsource: This is I guess to catch when the file does not exist, but there are maybe other exception, e.g. if the file is not readable for permission reasons. There is maybe a way to check the error and rethrow or log something when the error is not caused by a missing file

src/cli/commands/integrate/git/git-husky.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  if (content.includes(HOOK_MARKER)) {
+    info(`Secrets check already present in .husky/${hook}.`);
+    return;
```
damien-urruty-sonarsource, line 42: Is this sufficient? What if we update the script? Should the marker contains some kind of version, so we can install a new version post-update? Or is it the purpose of the force option?

src/cli/commands/integrate/git/git-husky.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+// Husky integration: appends a secrets-scan snippet to an existing .husky hook file.
+
+import { info, success } from '../../../../ui';
+import { HOOK_MARKER, getHuskySnippet } from './git-shell-fragments';
+import type { GitHookType } from '.';
+
+export async function installViaHusky(huskyHookPath: string, hook: GitHookType): Promise<void> {
+  const fs = await import('node:fs/promises');
```
damien-urruty-sonarsource: Any reason to not declare it at the top level?

src/cli/commands/integrate/git/git-shell-fragments.ts
```diff
@@ -0,0 +1,130 @@
...
+  # Branch deletion — nothing to scan
+  [ "$local_sha" = '0000000000000000000000000000000000000000' ] && continue
+  if [ "$remote_sha" = '0000000000000000000000000000000000000000' ]; then
+${newBranchPushBlock('files')}
+  else
+    files=$(git diff --name-only --diff-filter=ACMR "$remote_sha" "$local_sha")
+  fi
+  [ -z "$files" ] && continue
+  echo "$files" | tr '\n' '\0' | xargs -0 "$SONAR_BIN" analyze secrets -- || exit 1
+done
```
damien-urruty-sonarsource, line 80: I would like to suggest a shift of approach. Those scripts are quite complex, and also everytime we will fix something in them, they will need to be reinstalled post-update. What if we move as much as we can to the CLI? We could introduce new options like `sonar analyze secrets --pre-commit` and `sonar analyze secrets --pre-push`. This way the install scripts would remain simple, and we won't have that duplication across them. And some users might find it useful to run this command on their own. Just an idea

src/cli/commands/integrate/git/index.ts
```diff
@@ -0,0 +1,368 @@
...
+    await resolveAuth({});
+  } catch {
+    throw new CommandFailedError('Not authenticated. Please run: sonar auth login');
+  }
+
+  if (options.global) {
+    return integrateGitGlobal(options);
+  }
+
+  const projectInfo = await discoverProject(process.cwd());
```
damien-urruty-sonarsource: Instead of relying on discoverProject, which does a bunch of things, you could directly use findGitRoot which is more adapted to our context

src/cli/commands/integrate/git/index.ts
```diff
@@ -0,0 +1,368 @@
...
+      throw new CommandFailedError('Installation cancelled');
+    }
+  }
+  blank();
+
+  const installation = await detectSonarHookInstallation(projectInfo.rootDir);
+  const useHusky = toForwardSlash(installation.hooksDir).startsWith(
+    toForwardSlash(join(projectInfo.rootDir, '.husky')),
+  );
+  const usePreCommitConfig = existsSync(join(projectInfo.rootDir, PRE_COMMIT_CONFIG_FILE));
```
damien-urruty-sonarsource: Instead of filling all variables at once, you could maybe find a way to check one only if the other is not true, e.g. with some functions. I think we could even introduce our own `GitRepo` type to get some clarity and better code distribution. Pseudo-code:

```
const gitRepo = findGitRoot(); //maybe errors if null

if (gitRepo.usesPreCommitFramework()) {
    await installViaPreCommitFramework(projectInfo.rootDir, hook);
  } else if (gitRepo.usesHusky()) {
    await installViaHusky(huskyHookPath, hook);
  } else {
    await installViaGitHooks(installation.hooksDir, hook, options.force);
  }
```

There are maybe more things we could move to GitRepo

src/cli/commands/integrate/git/index.ts
```diff
@@ -0,0 +1,368 @@
...
+  force?: boolean,
+): Promise<void> {
+  mkdirSync(hooksDir, { recursive: true });
+  const hookPath = join(hooksDir, hook);
+  const fs = await import('node:fs/promises');
+  if (existsSync(hookPath)) {
+    const existing = await fs.readFile(hookPath, 'utf-8');
+    if (!existing.includes(HOOK_MARKER) && !force) {
+      warn(`A different ${hook} hook already exists at ${hookPath}.`);
+      text('  Use --force to replace it, or add the secrets check manually.');
```
damien-urruty-sonarsource: > or add the secrets check manually

We could maybe give the exact command

src/cli/commands/integrate/git/index.ts
```diff
@@ -0,0 +1,368 @@
...
+  try {
+    gitResult = await spawnProcess('git', [
+      'config',
+      '--global',
+      'core.hooksPath',
+      toForwardSlash(GLOBAL_HOOKS_DIR),
+    ]);
+  } catch {
+    const msg = 'git is not installed or not on PATH';
+    throw new CommandFailedError(msg);
```
damien-urruty-sonarsource: I would at least log the exception in case it's something else

src/cli/commands/integrate/git/index.ts
```diff
@@ -0,0 +1,368 @@
...
+  }
+  if (gitResult.exitCode !== 0) {
+    const detail = [gitResult.stderr, gitResult.stdout].filter(Boolean).join('\n');
+    const msg = `git config --global core.hooksPath failed (exit code ${gitResult.exitCode}): ${detail}`;
+    logger.error(msg);
```
damien-urruty-sonarsource: No need to log from here. Maybe review all exceptions that are thrown

damien-urruty-sonarsource (commented):
A first round of comments, I will try the feature now

tests/unit/integrate-git.test.ts
```diff
@@ -0,0 +1,708 @@
...
+    );
+  });
+
+  afterEach(() => {
+    setMockUi(false);
+    resolveAuthSpy.mockRestore();
+    performSecretInstallSpy.mockRestore();
+  });
+
+  it('returns without throwing when the user cancels the global install confirmation', () => {
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: Will fix in a separate PR.

src/cli/commands/analyze/secrets.ts
sophio-japharidze-sonarsource: It just filters out '' or `undefined` values.

src/cli/commands/integrate/git/index.ts
sophio-japharidze-sonarsource: I created CLI-174 for this

src/cli/commands/integrate/git/git-husky.ts
sophio-japharidze-sonarsource: https://sonarsource.atlassian.net/browse/CLI-177

src/cli/commands/integrate/git/git-shell-fragments.ts
sophio-japharidze-sonarsource: https://sonarsource.atlassian.net/browse/CLI-176

src/cli/commands/integrate/git/index.ts
sophio-japharidze-sonarsource: I'm not sure it can be a simple command... Will remove the manual option for now.

src/cli/commands/integrate/git/git-shell-fragments.ts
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: fixed in https://github.com/SonarSource/sonarqube-cli/pull/119

tests/unit/integrate-git.test.ts
```diff
@@ -0,0 +1,709 @@
...
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: fixed in https://github.com/SonarSource/sonarqube-cli/pull/119

tests/unit/integrate-git.test.ts
```diff
@@ -0,0 +1,709 @@
...
+    setMockUi(true);
+    queueMockResponse(null);
+    try {
+      expect(resolveHookType({})).rejects.toThrow('Installation cancelled');
+    } finally {
+      setMockUi(false);
+    }
+  });
+
+  it('propagates the error when secrets installation fails after the user confirms', () => {
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: fixed in https://github.com/SonarSource/sonarqube-cli/pull/119

damien-urruty-sonarsource (approved):
LGTM!

PR #103, opened 2026-03-13
CLI-142 CLI-143 CLI-144 CLI-148 Dogfooding bug fixes

kirill-knize-sonarsource:
* CLI-143 sonar-a3s hook must always be project-level, even with -g flag
* CLI-142 Fix incorrect analyze subcommand in hook scripts
* CLI-144 Fix bash hook JSON parsing to handle pretty-printed format
* CLI-148 Fix global integrate overwriting project-level agentExtensions

### DIFF
--- a/src/cli/commands/integrate/claude/hook-templates.ts
+++ b/src/cli/commands/integrate/claude/hook-templates.ts
@@ -32,15 +32,15 @@ if ! command -v sonar &> /dev/null; then
   exit 0
 fi
 
-# Read JSON from stdin and extract fields using grep/cut (no external runtimes required)
+# Read JSON from stdin and extract fields using sed (handles both compact and pretty-printed JSON)
 stdin_data=$(cat)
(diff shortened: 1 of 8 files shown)


vnaskos-sonar:
Related to CLI-143
When running `sonar integrate claude -g`
1. it installs the secrets hooks on the home directory
2. it installs the a3s hook on the current directory

If 2. is expected (and not a bug), it's very confusing because `-g` could run from a totally random directory which may not even be project.

I was expecting that `-g` will skip the a3s installation and print a warning or info message to explain to the user why this happened.

vnaskos-sonar:
There is a issue with the state being wrongly overwritten when
1. `cd my-project`
1. `sonar integrate claude`
1. from the same directory `sonar integrate claude -g`

on the `state.json` file both refer to the same `projectRoot` which means that the last `integrate` will overwrite the previous one with a different `global` value. Eventually one of the two installations is lost from the state. In the example above the secrets hooks from the `my-project` with `global: false` is lost.

Probably the `integrate -g` should have `projectRoot: "<homedir>"` 🤔

vnaskos-sonar (approved):

PR #107, opened 2026-03-15
CLI-101 CLI integrate claude resolves org from project rather than auth

vnaskos-sonar:
# Behavior changes
1. `sonar integrate claude` now relies on the authentication flow (via `resolveAuth`)
2. the resolution of the organization follows
    * `-o --org` option
    * fallback to `auth login` configuration
    * fallback to local project discovery (`sonar-project.properties` and `connectecMode.json`)
3. similar flow for `serverUrl` and `projectKey`

# Expect
1. no other change in behavior
2. cleanup on the `integrate/` code (removed duplicated code)
3. extensive unit tests
4. refactoring for testability
5. extraction of reusable functions

### DIFF
--- a/tests/unit/auth-org-selection.test.ts
+++ b/tests/unit/auth-org-selection.test.ts
@@ -57,7 +57,7 @@ describe('authLogin: org selection', () => {
     clearMockUiCalls();
     loadStateSpy = spyOn(stateManager, 'loadState').mockReturnValue(getDefaultState('test'));
     saveStateSpy = spyOn(stateManager, 'saveState').mockImplementation(() => undefined);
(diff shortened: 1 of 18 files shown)


tests/unit/auth-org-selection.test.ts
```diff
@@ -153,7 +153,7 @@ describe('authLogin: org selection', () => {
 
     try {
       await authLogin({ server: 'https://sonarcloud.io' });
-    } catch (error) {
+    } catch (error: any) {
```
kirill-knize-sonarsource, line 156: Maybe we want to do something with this broad catch later. Let's not block on it now, since it's pre-existing.

kirill-knize-sonarsource (approved):
LGTM, broad catch is to be fixed later

PR #113, opened 2026-03-17
CLI-165 Automate install scripts version update on release

kirill-knize-sonarsource:


.github/workflows/full-release.yml
```diff
@@ -61,6 +61,18 @@ jobs:
       - uses: jdx/mise-action@c37c93293d6b742fc901e1406b8f764f6fb19dac # v2.4.4
         with:
           version: 2025.9.12
+      - name: Get released version
+        id: released-version
+        run: |
+          git fetch --tags
+          RELEASED_VERSION=$(git tag --merged HEAD --sort=-version:refname | head -1)
```
damien-urruty-sonarsource: Instead of checking tags, a cleaner way would be to have our common release action return the released version, the same way it returns the new version that is used below. I think this could also benefit MCP since we also need to hardcode the released version in some files

damien-urruty-sonarsource (approved):
I think it should work but there is a better way to do it IMO

.github/workflows/full-release.yml
kirill-knize-sonarsource: We will do it as a separate improvement

PR #115, opened 2026-03-18
CLI-128 Enforce 'auth' for feature commands

damien-urruty-sonarsource:


src/lib/auth-resolver.ts
```diff
@@ -20,18 +20,19 @@
...
 // Centralized auth resolver - resolves token + serverUrl from env vars, state, or keychain
 
-import { SONARCLOUD_HOSTNAME, SONARCLOUD_US_HOSTNAME } from './config-constants';
+import { SONARCLOUD_HOSTNAME, SONARCLOUD_URL, SONARCLOUD_US_HOSTNAME } from './config-constants';
 import { getToken } from './keychain.js';
 import { loadState, getActiveConnection } from './state-manager.js';
 import { warn } from '../ui';
 import logger from './logger.js';
 
 export const ENV_TOKEN = 'SONAR_CLI_TOKEN';
```
sophio-japharidze-sonarsource, line 29: Should we rename these to `SONARQUBE_CLI_*` before it's too late?

src/lib/auth-resolver.ts
```diff
@@ -20,18 +20,19 @@
...
 export const ENV_SERVER = 'SONAR_CLI_SERVER';
+export const ENV_ORG = 'SONAR_CLI_ORG';
 
 export interface ResolvedAuth {
   token: string;
-  serverUrl: string;
+  serverUrl?: string;
   orgKey?: string;
   /** Type of the active connection; undefined when resolved from env vars alone */
   connectionType?: 'cloud' | 'on-premise';
```
sophio-japharidze-sonarsource: It would be nice if we can make this non-optional field & set it even inside `resolveFromEnv`. Then any command with `ResolvedAuth` can easily determine what connection type they are dealing with.
Similarly, `serverUrl` can be made non-optional, and default to `https://sonarcloud.io` in case resolved from env & it's not explicitly provided.

src/cli/commands/analyze/sqaa.ts
```diff
@@ -35,55 +35,53 @@ export interface AnalyzeSqaaOptions {
...
 
 /**
- * Resolve auth and validate that the connection is SonarQube Cloud.
- * Returns null when SQAA should be silently skipped (no auth / on-premise without --project).
+ * Validate that the resolved auth is for SonarQube Cloud.
+ * Returns null when SQAA should be silently skipped (on-premise or missing orgKey without --project).
  * Throws CommandFailedError when --project is set but the connection is not Cloud.
  */
-async function resolveCloudAuth(
+function resolveCloudAuth(
```
sophio-japharidze-sonarsource, line 74: I think this method can be simplified/removed.

If `connectionType` becomes mandatory on `ResolvedAuth`, we can completely drop it. Otherwise, there should be no need to check the existence of the token at least.

src/cli/commands/analyze/sqaa.ts
```diff
@@ -35,55 +35,53 @@ export interface AnalyzeSqaaOptions {
...
-
+): { serverUrl: string; token: string; orgKey: string } | null {
   if (!auth.token || !auth.orgKey || auth.connectionType === 'on-premise') {
     if (explicitProject) {
       throw new CommandFailedError(
         'SQAA analysis requires a SonarQube Cloud connection. Run: sonar auth login',
       );
     }
-    logger.debug('SQAA analysis skipped: no auth, missing orgKey, or on-premise server');
+    logger.debug('SQAA analysis skipped: missing orgKey or on-premise server');
```
sophio-japharidze-sonarsource, line 84: SQAA is not supported on on-premise server ?? 🤔 I know it's not related to this PR :D

sophio-japharidze-sonarsource (approved):
I love it! 😍 With some nice-to-have suggestions

src/cli/commands/analyze/sqaa.ts
damien-urruty-sonarsource: No, it's not for now!

src/lib/auth-resolver.ts
```diff
@@ -20,102 +20,112 @@
...
 
-  // 1. Both env vars present → use them immediately
+  // 1. Both SONAR_CLI_TOKEN + SONAR_CLI_SERVER env vars present → use them immediately
   if (envToken && envServer) {
     logger.debug('Using environment variable authentication');
     return {
       token: envToken,
       serverUrl: envServer,
-      orgKey: options.org,
+      connectionType: 'on-premise',
```
sophio-japharidze-sonarsource: Unfortunately here we will need to check that server URL does not equal to `sonarqube.us`. And if it does, require org key too.

src/lib/auth-resolver.ts
damien-urruty-sonarsource: Done. I think the ResolvedAuth could be improved further, we also have other types that are very similar, I would like to keep only one. But I will keep that for later

src/cli/commands/analyze/sqaa.ts
damien-urruty-sonarsource: I kept it because there are some checks that are made on top of the auth, like explicitProject. But I don't like we return a separate object, I would like to improve it later

sophio-japharidze-sonarsource (changes requested):
I think with the current implementation of auth-resolver, SQC US support will not work properly.

sophio-japharidze-sonarsource (approved):

src/lib/auth-resolver.ts
damien-urruty-sonarsource: I will bring this to Edgar/Clifford, it's not only for us to decide

PR #119, opened 2026-03-19
CLI-173 E2E test for pre-commit & pre-push hooks

sophio-japharidze-sonarsource:


tests/integration/specs/integrate/git.test.ts
```diff
@@ -21,9 +21,78 @@
...
 
 import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
-import { mkdirSync, writeFileSync } from 'node:fs';
+import { mkdirSync, symlinkSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { TestHarness } from '../../harness';
+import { BINARY_PATH } from '../../harness/cli-runner.js';
+
+const PATH_DELIM = process.platform === 'win32' ? ';' : ':';
+function pathWithoutNodeModules(envPath: string | undefined): string {
```
sophio-japharidze-sonarsource, line 30: This is again needed to avoid calling sonar-scanner JS dependency

src/cli/commands/integrate/git/index.ts
```diff
@@ -103,13 +103,24 @@ export async function detectSonarHookInstallation(root: string): Promise<HookIns
...
+/** Rejects invalid `--hook` when it is set */
+export function validateHookOption(hook: string | undefined): void {
+  if (hook !== undefined && !isGitHookType(hook)) {
+    throw new InvalidOptionError('--hook must be pre-commit or pre-push');
+  }
+}
+
+/**
+ * Returns explicit `--hook`, or `pre-commit` when non-interactive with no hook, or prompts.
+ * Call `validateHookOption` first for CLI-driven options.
```
kirill-knize-sonarsource: Not sure what happened here. It says "Call `validateHookOption` first for CLI-driven options." and then we call `validateHookOption` inside of the function. There are two places where we call this function from - one caller calls `validateHookOption` before, another doesn't. Looks like it's not coherent and something is off.

src/cli/commands/_common/git-repo.ts
```diff
@@ -24,9 +24,8 @@ import { existsSync, statSync } from 'node:fs';
 import { isAbsolute, join } from 'node:path';
 import { CommandFailedError } from './error';
 import { spawnProcess } from '../../../lib/process';
-
-const PRE_COMMIT_CONFIG_FILE = '.pre-commit-config.yaml';
-export const toForwardSlash = (p: string) => p.replaceAll('\\', '/');
+import { PRE_COMMIT_CONFIG_FILE } from '../integrate/git/git-precommit-framework';
+export const toForwardSlash = (p: string): string => p.replaceAll('\\', '/');
```
kirill-knize-sonarsource: This would be great to fetch to some common place, because we will need it for https://sonarsource.atlassian.net/browse/CLI-199.
Path normalization is not git specific thing.

kirill-knize-sonarsource (changes requested):

kirill-knize-sonarsource (approved):

PR #121, opened 2026-03-19
CLI-179 add network test for install.sh script

kirill-knize-sonarsource:


damien-urruty-sonarsource (commented):
I think it's a good start, but I have couple of comments:
* how do you ensure the test is run only on linux/mac? If someone on Windows checks out the project, the test will fail
* I would maybe put this test in a separate folder, it's yet different than integration tests. Maybe `end-to-end`, or `install`?

Also some more observations or nice-to-have:
* I think the test is doing the bare minimum, it's fine for now, we will probably include more tests soon
* At some point if we have more tests like this one, we could also build a "install tests harness", similar to what we have for integration tests. I can already see a lot of setup boilerplate can be factorized in a single place

damien-urruty-sonarsource (approved):
LGTM, do we need to create a separate ticket to test the Windows install script?

kirill-knize-sonarsource:
> LGTM, do we need to create a separate ticket to test the Windows install script?

https://sonarsource.atlassian.net/browse/CLI-180

PR #126, opened 2026-03-24
CLI-194 rename a3s to sqaa in code, methods, hooks and templates

kirill-knize-sonarsource:


src/cli/commands/integrate/claude/hooks.ts
```diff
@@ -143,6 +144,40 @@ async function installHook(params: HookInstallParams): Promise<void> {
   await fsPromises.writeFile(settingsPath, JSON.stringify(settings, null, 2), 'utf-8');
 }
 
+/**
+ * Remove all hook entries matching any of the given markers from settings.json.
+ * Used to clean up obsolete hook entries when hook names change across versions.
+ */
+async function removeObsoleteHookEntries(installDir: string, markers: string[]): Promise<void> {
```
sophio-japharidze-sonarsource: We should also remove the `sonar-a3s` folder under `hooks`

sophio-japharidze-sonarsource: Also I think this does not remove or replace an entry in the `state.json`

kirill-knize-sonarsource: Addresed.

src/lib/migration.ts
sophio-japharidze-sonarsource, line 1: This whole file looks way too complex to maintain. We can maybe refactor it before we go to GA.

sophio-japharidze-sonarsource (changes requested):
It still does not clean up state.json properly. I think we are missing a call to `saveState` somewhere.

src/lib/migration.ts
kirill-knize-sonarsource: Yes, ticket created.
https://sonarsource.atlassian.net/browse/CLI-201

kirill-knize-sonarsource:
> It still does not clean up state.json properly. I think we are missing a call to `saveState` somewhere.

Pushed more tests around it. I can't find it. We have ITs that test exactly this scenario.

sophio-japharidze-sonarsource (approved):

PR #129, opened 2026-03-26
CLI-186 refer to SONARQUBE_CLI consistently

sophio-japharidze-sonarsource:


kirill-knize-sonarsource (approved):
Commit message doesn't really reflect what is the goal of the ticket. I think ideally it should match the ticket title if it's not too long, or at least rephrase it without losing the point.
Like "Rename env vars to SONARQUBE_CLI_<NAME>" or similar.

kirill-knize-sonarsource (approved):
LGTM!

PR #131, opened 2026-03-26
CLI-127 Codesign MacOS binary

kirill-knize-sonarsource:


.github/workflows/build.yml
```diff
@@ -124,6 +124,56 @@ jobs:
...
+        id: apple-secrets
+        if: matrix.os == 'macos' && (github.ref_name == github.event.repository.default_branch || startsWith(github.ref_name, 'branch-'))
+        uses: SonarSource/vault-action-wrapper@320bd31b03e5dacaac6be51bbbb15adf7caccc32 # 3.1.0
+        with:
+          secrets: |
+            development/kv/data/sign/sonarqube-cli CERTIFICATE | CERTIFICATE;
+            development/kv/data/sign/sonarqube-cli CSC_KEY_PASSWORD | CSC_KEY_PASSWORD;
+            development/kv/data/sign/sonarqube-cli APPLE_TEAM_ID | APPLE_TEAM_ID;
+            development/kv/data/sign/sonarqube-cli APPLE_ID | APPLE_ID;
+            development/kv/data/sign/sonarqube-cli APPLE_APP_SPECIFIC_PASSWORD | APPLE_APP_SPECIFIC_PASSWORD;
```
damien-urruty-sonarsource, line 137: This syntax is a bit strange, usually the first fragment is the token key, I'm surprised we have many times the same here, is it a different syntax?

build-scripts/codesign-local.ts
```diff
@@ -0,0 +1,52 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { spawnSync } from 'node:child_process';
```
damien-urruty-sonarsource, line 20: Is this script automatically called after the binary is built? It's my understanding as it's called post*. If yes, couldn't we remove things from the CI script?

CONTRIBUTING.md
```diff
@@ -79,6 +79,27 @@ Obtain a token from **sonarcloud.io → Account → Security → Generate token*
 
 If the variables are not set, the relevant tests are skipped automatically — this is expected for local development.
 
+## macOS code signing (optional)
```
damien-urruty-sonarsource, line 82: Is it the right place to document this? To me CONTRIBUTING is general guidance for external contributors, but I might be wrong

damien-urruty-sonarsource (approved):
LGTM, I left a couple of questions

.github/workflows/build.yml
kirill-knize-sonarsource: I'd say syntax is the same, it's just the names of the vars in vault this time exactly the same as the vars we use in workflow. Normally they differ a bit - you can see lines 94-98 above.

build-scripts/codesign-local.ts
kirill-knize-sonarsource: It will not do anything on CI - it checks if it's on mac, env var APPLE_TEAM is set and certificate is in keychain.
If one of conditions is not met - script will exit not doing anything. I don't think we need to explicitly disable it for CI. It's not possible by coincidence meet all conditions on pipeline.

CONTRIBUTING.md
kirill-knize-sonarsource: Not sure what the other place should be. I've marked it as optional. Maybe we can decide where to move it later.

PR #135, opened 2026-03-27
CLI-212 Sonar CLI UX demonstrate value in sonar get started menu

kirill-knize-sonarsource:


tests/unit/command-tree-help.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+describe('getCustomRootHelp', () => {
+  it('includes the box banner with version', () => {
+    const help = getCustomRootHelp();
+    expect(help).toContain('┌');
+    expect(help).toContain('SonarQube CLI');
+    expect(help).toMatch(/v\d+\.\d+\.\d+/);
+  });
+
+  it('contains all required sections and content', () => {
```
sophio-japharidze-sonarsource: Why do we need to test it in both places ? i.e. we have the same in ITs.

src/cli/command-tree.ts
```diff
@@ -39,31 +39,24 @@ import { parseInteger } from './commands/_common/parsing';
...
-    '',
-  ].join('\n');
-}
 
 export const COMMAND_TREE = new SonarCommand();
 
 COMMAND_TREE.name('sonar')
   .description('SonarQube CLI')
   .version(VERSION, '-v, --version', 'display version for command')
-  .addHelpText('beforeAll', getHelpBanner())
```
sophio-japharidze-sonarsource, line 65: We used to display the version & brand banner on all help commands (for example `sonar auth help` too). We might want to check this is intended change.

kirill-knize-sonarsource: It was not discussed with UX. Returned it back. Good catch.

tests/unit/command-tree-help.test.ts
kirill-knize-sonarsource: Redundant unit test removed.

sophio-japharidze-sonarsource (approved):

PR #148, opened 2026-04-02
CLI-125 add caller_agent field to telemetry

sophio-japharidze-sonarsource:


src/lib/agent-detector.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Claude Code markers take precedence when both families could be set.
+ *
+ * @param env - Defaults to `process.env`; inject a custom object for tests.
+ */
+export function detectCallerAgent(env: NodeJS.ProcessEnv = process.env): CallerAgent {
+  if (isClaudeCodeAgentEnv(env)) return 'claude';
+  if (isCursorAgentEnv(env)) return 'cursor';
+  return '';
```
damien-urruty-sonarsource: I would double check with the DPO what should be the right value here, empty string seems a bit strange. We could have null, or omit the field in the payload

damien-urruty-sonarsource (approved):
LGTM, I would just double check what the payload should be when no agent is detected

src/lib/agent-detector.ts
sophio-japharidze-sonarsource: I have changed it to use the same pattern as other fields - `string | null`

PR #149, opened 2026-04-02
CLI-242 CLI-243 Fix pre-commit hook when CLI not installed and onboard with renovate

nquinquenel:


tests/unit/git-husky.test.ts
```diff
@@ -93,3 +105,35 @@ describe('installViaHusky', () => {
     expect(getMockUiCalls().some((c) => c.method === 'success')).toBe(false);
   });
 });
+
+describe('git-shell-fragments (sonar resolution)', () => {
+  it.each([
```
sophio-japharidze-sonarsource, line 136: Should we not test for the actual shell fragments that will be installed instead of hardcoding them here? i.e. do these tests actually test the implementation?

sophio-japharidze-sonarsource (approved):
Thanks for the change! Please have a look at the comment about tests

tests/unit/git-husky.test.ts
nquinquenel: I think it's probably better, you're right. I did an actual integration test where we git add a file, run hook and check for the output

PR #150, opened 2026-04-07
CLI-278 review auth tests; remove GET option for token submission; remove success HTML

sophio-japharidze-sonarsource:


src/cli/commands/_common/token.ts
```diff
@@ -236,30 +162,16 @@ export function handlePostRequest(
...
-
 /**
  * Create request handler for loopback server
  */
 export function createRequestHandler(onToken: (token: string) => void) {
   return (req: IncomingMessage, res: ServerResponse) => {
     if (req.method === 'POST') {
       handlePostRequest(req, res, onToken);
-    } else if (req.method === 'GET') {
-      handleGetRequest(req, res, onToken);
```
damien-urruty-sonarsource, line 259: It's surprising, but requests coming from SQC/SQS are in fact GET requests. I think we implemented it like this back then because of CORS limitations. I'm not even sure we have POST requests, we would need to check

src/cli/commands/_common/token.ts
```diff
@@ -236,30 +162,16 @@ export function handlePostRequest(
...
     } else {
-      res.writeHead(HTTP_STATUS_OK);
-      res.end('OK');
+      res.writeHead(HTTP_STATUS_BAD_REQUEST);
+      res.end('Bad Request');
```
damien-urruty-sonarsource: Nitpick: I think there is a dedicated status for unsupported method

damien-urruty-sonarsource (commented):
I think this breaks the embedded server functionality. Also, but maybe it's a separate topic, but we have many auth-* test files. I wonder if we should consolidate them into a single file, or if we could have integration tests for some of them. But this is OK for now

src/cli/commands/_common/token.ts
sophio-japharidze-sonarsource: Are you sure? SQC sends POST request when sending the token

Screenshot 2026-04-14 at 10 46 14

And in SLCORE for the token handler we reject any non-post request. See [source](https://github.com/SonarSource/sonarlint-core/blob/6e5cb6ee030e2990d9ce061ace5603755c4837b2/backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/handler/GeneratedUserTokenHandler.java#L53).

Manual test with this removal also succeeds.

damien-urruty-sonarsource: Ah yes, what I tried was the "Open in IDE" feature. We probably won't need it in the CLI, so it's probably OK to remove support for GET

damien-urruty-sonarsource (approved):

PR #151, opened 2026-04-07
CLI-209 ProjectKey not detected from connected mode file

nquinquenel:
Project detection

- SonarQube Cloud: reads sonarCloudOrganization + projectKey, optional region (EU vs US base URL).
- SonarQube Server: reads sonarQubeUri + projectKey only (no org).
- Looks only under .sonarlint/: connectedMode.json (any casing), then other *.json (e.g. solution bindings).
- Removed: settings.json, connected-mode.json, serverId, connectionId - not part of the SQ:IDE model we support.

Code

- Discovery logic now lives in src/lib/project-workspace/: sonarlint-connected-mode.ts (load/parse .sonarlint), project-info.ts, index.ts barrel.

Tests

- tests/unit/project-workspace/ mirrors the lib layout: sonarlint-connected-mode.test.ts and project-info.test.ts.
- tests/integration/specs/project-workspace/sonarlint-connected-mode.test.ts - same loader against TestHarness project trees.
- Removed: monolithic discovery.test.ts and the old mock.module('node:fs/promises') test file; I/O edge cases use the real filesystem in the unit file where needed.


src/lib/project-workspace/project-info.ts
```diff
@@ -139,7 +133,7 @@ export async function discoverProjectInfo(startDir: string): Promise<ProjectInfo
   }
 
   const sonarProps = await loadSonarProperties(projectRoot);
-  const sonarLintConfig = await loadSonarLintConfig(projectRoot);
+  const sonarLintLoaded = await loadSonarLintConfig(projectRoot);
```
sophio-japharidze-sonarsource, line 136: I feel this name is confusing, `sonarlintConfig` was clearer.

src/lib/project-workspace/project-info.ts
```diff
@@ -18,13 +18,14 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-// Discovery module - discovers project information
+// Project workspace: git root, sonar-project.properties, SonarLint connected mode
```
sophio-japharidze-sonarsource, line 21: What is "project workspace" ? 😅 It sounds very cool and abstract, but maybe we can have a more specfic name - like `project discovery`?

src/lib/project-workspace/sonarlint-connected-mode.ts
```diff
@@ -0,0 +1,165 @@
...
+    if ('sonarQubeUri' in generic) {
+      const serverURL = parseStringValue(generic.sonarQubeUri).trim();
+      const projectKey = parseStringValue(generic.projectKey).trim();
+      if (!serverURL || !projectKey) {
+        return null;
+      }
+      return {
+        serverURL,
+        projectKey,
+        organization: '',
```
sophio-japharidze-sonarsource: Why do we need to set an empty organization? Can we make the field optional instead?

src/lib/project-workspace/sonarlint-connected-mode.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    return null;
+  } catch {
+    return null;
+  }
+}
+
+async function tryLoadSonarLintFile(configPath: string): Promise<SonarLintConfig | null> {
+  const fs = await import('node:fs/promises');
```
sophio-japharidze-sonarsource: Let's move this import to top level

src/lib/project-workspace/sonarlint-connected-mode.ts
```diff
@@ -0,0 +1,165 @@
...
+function parseStringValue(value: unknown): string {
+  return typeof value === 'string' ? value : '';
+}
+
+function parseSonarLintConfig(data: string): SonarLintConfig | null {
+  try {
+    const generic = JSON.parse(data) as Record<string, unknown>;
+
+    // SonarQube Server: sonarQubeUri + projectKey
+    if ('sonarQubeUri' in generic) {
```
sophio-japharidze-sonarsource: In VS config files (`<solutionName>.json`), the names will be in PascalCase - `SonarQubeUri`, `SonarCloudOrganization`, `ProjectKey`

tests/integration/specs/project-workspace/sonarlint-connected-mode.test.ts
```diff
@@ -0,0 +1,134 @@
...
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'bun:test';
+import { mkdirSync, writeFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { SONARCLOUD_URL } from '../../../../src/lib/config-constants';
+import { loadSonarLintConfig } from '../../../../src/lib/project-workspace';
+import { TestHarness } from '../../harness';
+
+describe('SonarLint connected mode (harness)', () => {
```
sophio-japharidze-sonarsource: These tests look more like unit tests, checking for `loadSonarLintConfiguration` method specifically. It would be great if we can call project discovery method instead, and assert that sonarlint config will be taken into account.

tests/unit/project-workspace/sonarlint-connected-mode.test.ts
```diff
@@ -0,0 +1,227 @@
...
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { SONARCLOUD_URL, SONARCLOUD_US_URL } from '../../../src/lib/config-constants';
+import { loadSonarLintConfig } from '../../../src/lib/project-workspace';
+
+function tempProject(name: string): string {
+  return join(tmpdir(), `sonarlint-ut-${name}-${Date.now()}`);
+}
+
+describe('loadSonarLintConfig', () => {
```
sophio-japharidze-sonarsource, line 32: How are these tests different from the ITs in the other file?

sophio-japharidze-sonarsource (changes requested):
Looks good, can be improved ⏬ 

I don't think this will work with VS configuration files. I think tests can be improved.

src/lib/project-workspace/project-info.ts
nquinquenel: I think `sonarLintConfig` is misleading, the object returned contains a config and a relative path, and you would end up with `sonarLintConfig.config` which is redundant.

Maybe `loadedSonarLint` or `loadedSonarLintConfig` ?

sophio-japharidze-sonarsource: How about `loadSonarLintClues` & `sonarLintClues`?

src/lib/project-workspace/project-info.ts
nquinquenel: `discovery` is what you do, not what you're describing. The APIs in this file still uses the discovery-style names like `dicoverProjectInfo`, but the package groups the source of truth - it describes the project workspace which is a local picture of the repo (git root, sonar-project, .sonarlint, etc.). IMO the name is fine

src/lib/project-workspace/sonarlint-connected-mode.ts
nquinquenel: No strong reason, it was already like that before (the file appears as modified because I re-organized the folders), I think it's worth making it optional, I'll do it 👍🏻

src/lib/project-workspace/sonarlint-connected-mode.ts
nquinquenel: Good point - so it was already broken before? (I kept this part as well)

I'll fix it 👍🏻

src/lib/project-workspace/sonarlint-connected-mode.ts
nquinquenel: Generally speaking, I think it's fine to use dynamic imports, it can defer loading or break circular dependency sometimes - it's also used this way in the code base already from what I've seen.

However, I agree that here the benefit is minimal, there's already a static node:fs at the top, so I'll change it 👍🏻

src/lib/project-workspace/sonarlint-connected-mode.ts
sophio-japharidze-sonarsource: Yes I think it never worked 😬 Thank you! 🙇

tests/unit/project-workspace/sonarlint-connected-mode.test.ts
nquinquenel: I think the difference is essentially the temp folder that is different by using the harness, I agree the difference is very thin 😅 

I'll try to update the tests so that:
- Integration tests are using discover_xxx methods
- Unit tests will be focused on specific methods, such as the loading/parsing methods

But I must admit that getting the coverage was challenging!

src/lib/project-workspace/project-info.ts
nquinquenel: `ResolvedSonarLintConfig`? 😅

sophio-japharidze-sonarsource (approved):

PR #157, opened 2026-04-08
CLI-244 CLI-245 callback infrastructure

kirill-knize-sonarsource:


src/cli/commands/hook/claude-pre-tool-use.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (payload.tool_name !== 'Read') return;
+
+  const filePath = payload.tool_input?.file_path;
+  if (!filePath || !existsSync(filePath)) return;
+
+  const auth = await resolveAuth().catch(() => null);
+  if (!auth) return; // not authenticated — allow gracefully
+
+  const binaryPath = resolveSecretsBinaryPath();
+  if (!binaryPath) return; // binary not installed — allow gracefully
```
sophio-japharidze-sonarsource: Has this been agreed with PM & UX? I agree it's good not to fail the Claude process completely, but at the same time somehow we need to let the users know that their integration is broken, so that they don't have a false sense of security.

sophio-japharidze-sonarsource: Maybe it's possible to put something in Claude status bar at this point?

src/cli/commands/hook/secrets-scan.ts
```diff
@@ -0,0 +1,85 @@
...
+  const platform = detectPlatform();
+  const binaryPath = join(BIN_DIR, buildLocalBinaryName(platform));
+  return existsSync(binaryPath) ? binaryPath : null;
+}
+
+/**
+ * Run secrets scan on the given files. Returns the binary exit code.
+ * May throw on timeout; other errors are surfaced as non-zero exit codes.
+ */
+export async function scanFiles(
```
sophio-japharidze-sonarsource: Let's call this `scanFilesForSecrets` maybe ? So that when it's imported in other files it's very clear what it does.

src/cli/commands/hook/secrets-scan.ts
```diff
@@ -0,0 +1,85 @@
+/*
```
sophio-japharidze-sonarsource: I think this logic should completely live under `analyze/secrets.ts` & `install/secrets.ts`. We should not need to re-implement scanning for secrets differently for analyze vs hook invocation. If **really** needed, this file should simply call common analysis methods & format output in a way that is adapted for hooks.

src/cli/commands/integrate/claude/hook-templates.ts
```diff
@@ -20,270 +20,42 @@
...
-
-if ($exitCode -eq 51) {
-    $reason = "Sonar detected secrets in prompt"
-    $response = @{
-        decision = "block"
-        reason = $reason
-    } | ConvertTo-Json
-    Write-Host $response
+export function getSqaaPostToolTemplateUnix(projectKey: string): string {
+  return unixTemplate(`sonar hook agent-post-tool-use --project ${projectKey}`);
```
sophio-japharidze-sonarsource: I guess implementation of these hooks are in a separate PR ? Just a note that we must not merge this until the command actually exists.

src/cli/commands/integrate/claude/hook-templates.ts
```diff
@@ -20,270 +20,42 @@
...
-  return String.raw`param(
-    [Parameter(ValueFromPipeline = $true)]
-    [string]$InputData
-)
-
-try {
-    $input = $InputData | ConvertFrom-Json -ErrorAction Stop
-} catch {
-    exit 0
+  return windowsTemplate('sonar hook agent-prompt-submit');
```
sophio-japharidze-sonarsource: Should we keep the `claude-prompt-submit` naming for all hook commands ?

src/cli/command-tree.ts
```diff
@@ -228,6 +229,44 @@ COMMAND_TREE.command('self-update')
   .option('--force', 'Install the latest version even if already up to date')
   .anonymousAction((options: SelfUpdateOptions) => selfUpdate(options));
 
+// Hidden callback command — internal handlers for agent and git hooks.
+// Shell hook scripts call `sonar hook <event>` to delegate all business logic to TypeScript.
+export const callbackCommand = COMMAND_TREE.command('hook', { hidden: true })
```
sophio-japharidze-sonarsource: let's call it `hookCommand` for clarity

src/cli/command-tree.ts
```diff
@@ -228,6 +229,44 @@ COMMAND_TREE.command('self-update')
...
+  .anonymousAction(function (this: Command) {
+    this.outputHelp();
+  });
+
+callbackCommand
+  .command('claude-pre-tool-use')
+  .description('PreToolUse handler: scan files for secrets before agent reads them')
+  .anonymousAction(() => claudePreToolUse());
+
+// codex-pre-tool-use reuses the same handler (same hook event, different agent name)
```
sophio-japharidze-sonarsource: Let's remove this for now. This is not in scope of the ticket, and it should not use the same handler.

src/cli/command-tree.ts
```diff
@@ -228,6 +229,44 @@ COMMAND_TREE.command('self-update')
...
+callbackCommand
+  .command('codex-pre-tool-use')
+  .description('PreToolUse handler for Codex: scan files for secrets before agent reads them')
+  .anonymousAction(() => claudePreToolUse());
+
+// agent-prompt-submit and agent-post-tool-use are installed by `sonar integrate claude`.
+// They are implemented in subsequent PRs; stubs here ensure scripts don't fail with
```
sophio-japharidze-sonarsource: Let's remove these PR-specific comments.

src/cli/command-tree.ts
```diff
@@ -228,6 +229,44 @@ COMMAND_TREE.command('self-update')
...
+// "unknown command" on a CLI that only has CLI-244/245.
+callbackCommand
+  .command('agent-prompt-submit')
```
sophio-japharidze-sonarsource: Naming should consistently include agent name in the command

tests/integration/specs/hook/hook-claude-pre-tool-use.test.ts
sophio-japharidze-sonarsource, line 1: 🔥

tests/integration/specs/hook/hook.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      expect(result.stdout).toContain('hook');
+    },
+    { timeout: 15000 },
+  );
+
+  it(
+    'sonar hook exits 0 (shows help)',
+    async () => {
+      const result = await harness.run('hook');
+      expect(result.exitCode).toBe(0);
```
sophio-japharidze-sonarsource, line 54: I'm not sure this assertion is enough. It is also kind of duplicate of the previous test. Let's only keep one of them and assert command description is shown.

sophio-japharidze-sonarsource (changes requested):
I love the direction this is going 😍

I left a few comments & suggestions, mainly about duplication of secrets scanning logic. I will test it now.

src/cli/command-tree.ts
```diff
@@ -229,6 +230,35 @@ COMMAND_TREE.command('self-update')
   .option('--force', 'Install the latest version even if already up to date')
   .anonymousAction((options: SelfUpdateOptions) => selfUpdate(options));
 
+// Hidden callback command — internal handlers for agent and git hooks.
+// Shell hook scripts call `sonar hook <event>` to delegate all business logic to TypeScript.
+export const hookCommand = COMMAND_TREE.command('hook', { hidden: true })
+  .description('Internal callback handlers for agent and git hooks')
```
sophio-japharidze-sonarsource:
```suggestion
  .description('Internal hook handlers for agent and git hooks')
```

src/cli/commands/analyze/secrets.ts
sophio-japharidze-sonarsource, line 1: I think it's a good start and there could be even more simplification. There is still quite a lot of duplication between `runSecretsBinary` / `runScanFromStdin` .

Also the failure scenarios in performStdinScan / performPathsScan look the same

kirill-knize-sonarsource: Worked on removing duplications.

src/cli/commands/analyze/secrets.ts
```diff
@@ -127,72 +152,21 @@ async function performPathsScan(
...
-            : {}),
-        },
-      }),
-      new Promise<never>((_resolve, reject) => {
-        timeoutId = setTimeout(() => {
-          reject(new Error(`Scan timed out after ${SCAN_TIMEOUT_MS}ms`));
-        }, SCAN_TIMEOUT_MS);
-      }),
-    ]);
+    return await spawnWithTimeout(binaryPath, [tempFile], {
```
sophio-japharidze-sonarsource: Why not call the `runSecretsBinary` here?

kirill-knize-sonarsource: Good catch! Fixed

sophio-japharidze-sonarsource (approved):

src/cli/commands/hook/claude-pre-tool-use.ts
kirill-knize-sonarsource: I think adding something to the status bar for Claude should be a separate ticket. I don't think we should introduce a new UI component as a part of migration of hooks. At least it was not agreed on.

PR #158, opened 2026-04-08
CLI-246 Move prompt submit hook from shell scripts

kirill-knize-sonarsource:


src/cli/commands/hook/agent-prompt-submit.ts
```diff
@@ -0,0 +1,62 @@
...
+interface PromptSubmitPayload {
+  prompt?: string;
+}
+
+export async function agentPromptSubmit(): Promise<void> {
+  let payload: PromptSubmitPayload;
+  try {
+    payload = await readStdinJson<PromptSubmitPayload>();
+  } catch {
+    return; // unparseable stdin — allow
```
eray-felek-sonarsource, line 39: Maybe we could log something here

eray-felek-sonarsource (approved):
Tested and LGTM

src/cli/commands/hook/agent-prompt-submit.ts
```diff
@@ -0,0 +1,62 @@
+/*
```
eray-felek-sonarsource, line 1: There is duplicated logic between this file and `claude-pre-tool-use.ts`. Like, auth, binary and parsing check, Maybe we could use a helper method to avoid duplication

eray-felek-sonarsource (approved):

PR #160, opened 2026-04-08
CLI-248 git hooks

kirill-knize-sonarsource:


src/cli/commands/hook/git-pre-commit.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (!auth) return; // not authenticated — skip gracefully
+
+  const binaryPath = resolveSecretsBinaryPath();
+  if (!binaryPath) return; // binary not installed — skip gracefully
+
+  try {
+    const result = await runSecretsBinary(binaryPath, stagedFiles, auth);
+    const exitCode = result.exitCode ?? 1;
+    if (exitCode === EXIT_CODE_SECRETS_FOUND) {
+      throw new CommandFailedError('Secrets detected in staged files');
```
eray-felek-sonarsource: As a user I only see this generic message, maybe we could return more comprehensive error message that includes what caused the failure

tests/integration/specs/hook/hook.test.ts
```diff
@@ -39,7 +39,8 @@ describe('sonar hook', () => {
     async () => {
       const result = await harness.run('');
       expect(result.exitCode).toBe(0);
-      expect(result.stdout).not.toContain('callback');
+      // 'hook' must not appear as a listed command entry (indented under COMMANDS)
```
eray-felek-sonarsource: Maybe comment could be slightly more descriptive about why the regex pattern matters

eray-felek-sonarsource (commented):
Small comments, otherwise tested and LGTM

PR #164, opened 2026-04-09
CLI-180 Test Windows install script

tomshafir-sonarsource:


kirill-knize-sonarsource (changes requested):
We can't confirm that it works from this PR since branch name is `task/ts/CLI-180-test-windows-install-script `. I suggest to open another PR for branch name starting with `branch-` , validate we have installation e2e tests running and green and then merge.
Than we can close this one without merging.

PR #172, opened 2026-04-10
CLI-288 Document how to uninstall the CLI

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
It is technically correct.
At the same time I think people want to delete it when something doesn't work. For example hooks are erroring out.
Do you think it makes sense to also explain how to check/remove hooks?

damien-urruty-sonarsource:
@kirill-knize-sonarsource as we just discussed, ideally we should have commands for that (reversibility). I suggest merging this PR, it's already useful on its own, and I don't have time to expand on it

PR #175, opened 2026-04-14
CLI-298 git pre push

kirill-knize-sonarsource:


src/lib/process.ts
```diff
@@ -79,6 +79,11 @@ export async function spawnProcess(
       proc.stdin.end();
     }
 
+    if (options.stdinData !== undefined && proc.stdin) {
```
eray-felek-sonarsource: This code block is duplicated

eray-felek-sonarsource (commented):
1 comment otherwise tested and LGTM

src/lib/process.ts
kirill-knize-sonarsource: Good catch, thanks!

PR #178, opened 2026-04-14
CLI-269 CLI-303 Implement Bun.secrets and remove keytar

nquinquenel:
First step - add Bun.secrets and remove keytar. Adapt current tests.

Not covered yet:
- Account index listing
- keytar clean up procedure
- CI runners refactoring for building
- Documentation


src/lib/keychain.ts
```diff
@@ -18,38 +18,64 @@
...
+
+const bunSecretsBackend: KeychainBackend = {
+  getPassword: (service, account) =>
+    wrapBunSecrets(() => Bun.secrets.get({ service, name: account })),
+  setPassword: (service, account, password) =>
+    wrapBunSecrets(() => Bun.secrets.set({ service, name: account, value: password })),
+  deletePassword: (service, account) =>
+    wrapBunSecrets(() => Bun.secrets.delete({ service, name: account })),
+  // Bun.secrets has no list/enumerate API - CLI-270 will add an account index
+  findCredentials: () => Promise.resolve([]),
```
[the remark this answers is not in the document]
nquinquenel: It will be done in scope of https://github.com/SonarSource/sonarqube-cli/pull/180

tests/unit/keychain/keychain-test-handle.ts
```diff
@@ -0,0 +1,120 @@
...
+
+      delete process.env.SONARQUBE_CLI_KEYCHAIN_FILE;
+      delete process.env.SONARQUBE_CLI_DISABLE_KEYCHAIN;
+      process.env.SONARQUBE_CLI_KEYCHAIN_SERVICE = serviceName;
+
+      clearTokenCache();
+    },
+
+    async teardown() {
+      const allAccounts = new Set([...trackedAccounts, ..._testGetCachedAccountKeys()]);
```
kirill-knize-sonarsource: Why do we need _testGetCachedAccountKeys?
It looks like we need it in case we call `saveToken()` from tests directly, bypassing handle. That's the only way we can have untracked accounts in cache. But we never do so - we call `handle.saveToken` or `handle.seedToken`, so `trackedAccounts` are always `allAccounts`.
Do I miss something?

kirill-knize-sonarsource (approved):
LGTM, one question on tests teardown.

tests/unit/keychain/keychain-test-handle.ts
nquinquenel: I think you're right - I was afraid that in some cases we could leak the token in some tests that would directly call the save method. I re-checked and it seems safe in the end. I'll remove it to clean up the code from this test-specific method

PR #180, opened 2026-04-14
CLI-270 CLI-311 Account index for purge and credential listing

nquinquenel:


src/lib/config-constants.ts
```diff
@@ -46,6 +46,7 @@ export const CLI_DIR = join(homedir(), '.sonar', APP_NAME);
 // ---------------------------------------------------------------------------
 
 export const STATE_FILE = join(CLI_DIR, 'state.json');
+export const ACCOUNT_INDEX_FILE = join(CLI_DIR, 'keychain-accounts.json');
```
kirill-knize-sonarsource: We can put it into `state.json`. No need to create separate file.

tests/integration/harness/environment-builder.ts
kirill-knize-sonarsource, line 1: I think in previous PR we lost the isolation of ITs from real OS keystore. It's critical on Macs, we should restore the isolation. Tests touching real keystore shall live in `e2e` dir and they need to exclude MacOS, because we get a lot of keystore prompts during execution.

kirill-knize-sonarsource (changes requested):

.github/workflows/build.yml
```diff
@@ -370,10 +353,31 @@ jobs:
...
+          mkdir -p ~/.local/share/keyrings
+          # Empty password is intentional — CI keyring is ephemeral and
+          # only used to give Bun.secrets a working backend for e2e tests.
+          eval "$(printf '\n' | gnome-keyring-daemon --unlock --components=secrets)"
+          if [ -n "$GNOME_KEYRING_CONTROL" ]; then
+            echo "GNOME_KEYRING_CONTROL=$GNOME_KEYRING_CONTROL" >> "$GITHUB_ENV"
+          fi
+
+      - name: Run e2e tests (Windows)
+        if: matrix.os == 'windows'
```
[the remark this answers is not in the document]
nquinquenel: It's intentional

src/lib/keychain.ts
```diff
@@ -207,13 +221,26 @@ export async function deleteToken(serverURL: string, org?: string): Promise<void
...
+  if (filePath) {
+    const store = readFileStore(filePath);
+    return Object.entries(store.tokens).map(([account, password]) => ({ account, password }));
+  }
+
+  const accounts = deriveAccountsFromConnections();
+  const service = getServiceName();
+  const results: Array<{ account: string; password: string }> = [];
+  for (const account of accounts) {
+    const password = await bunSecretsBackend.getPassword(service, account);
```
kirill-knize-sonarsource: We have a function `getBackend()`. Shouldn't we use it here?

kirill-knize-sonarsource (approved):
Minor comment, else LGTM

src/lib/keychain.ts
nquinquenel: Good catch, there's no reason not to

PR #177, opened 2026-04-14
Add linux-arm64 build target

mcfedr:
## Summary
- Add architecture detection (`uname -m`) to `install.sh` and `install-prerelease.sh` so they resolve `linux-arm64` on ARM hosts and `linux-x86-64` on x86 hosts
- Add `linux-arm64` matrix entry to the CI build pipeline (`build.yml`) with `bun-linux-arm64` compile target
- Add artifact download step and publish entry for the new `linux-arm64` binary

## Test plan
- [ ] Verify the `github-ubuntu-latest-arm64-m` runner is available in CI
- [ ] Confirm the `bun-linux-arm64` compile target produces a working binary
- [ ] Run `install.sh` on a linux-arm64 host and verify correct platform detection
- [ ] Run `install.sh` on a linux-x86-64 host and verify no regression
- [ ] Run `install.sh` on macOS and verify no regression

PR #184, opened 2026-04-14
Add macos-x86-64 build target

Ghostavio:
## Summary
- Add architecture detection (`uname -m`) to `install.sh` and `install-prerelease.sh` so they resolve `macos-x86-64` on Intel hosts and `macos-arm64` on Apple Silicon hosts
- Add `macos-x86-64` matrix entry to the CI build pipeline (`build.yml`) with `bun-darwin-x64` compile target
- Add artifact download step and publish entry for the new `macos-x86-64` binary
- Update macOS-specific CI conditions to use `startsWith(matrix.os, 'macos')` so code-signing and notarization apply to both macOS variants

## Test plan
- [x] Verify `detect_arch` returns `x86-64` on an Intel Mac (`uname -m` → `x86_64`)
- [x] Verify `detect_platform` returns `macos-x86-64` on an Intel Mac
- [x] Build binary locally with `bun run build:binary` on Intel Mac — produces `Mach-O 64-bit executable x86_64`
- [x] Install and run `sonar --version` / `sonar --help` on Intel Mac — works correctly
- [ ] Confirm the `macos-13` runner produces a working binary in CI
- [ ] Run `install.sh` on macOS ARM64 and verify no regression
- [ ] Run `install.sh` on Linux x86-64 and verify no regression

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,13 +64,17 @@ jobs:
             runner: macos-latest-xlarge
             target: bun-darwin-arm64
             platform: macos-arm64
+          - os: macos-x86-64
+            runner: macos-13
+            target: bun-darwin-x64
+            platform: macos-x86-64
           - os: windows
             runner: warp-custom-windows-2022-s
             target: bun-windows-x64
             platform: windows-x86-64
     steps:
       - name: Setup Cloudflare WARP (macOS only)
-        if: matrix.os == 'macos'
+        if: startsWith(matrix.os, 'macos')
         uses: SonarSource/gh-action_setup-cloudflare-warp@v1
 
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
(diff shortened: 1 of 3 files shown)


Ghostavio:
Actually upon further testing, for this to work a `macos-x86-64` binary of https://github.com/SonarSource/sonar-secrets-pre-commit would also be necessary.

sophio-japharidze-sonarsource:
Thanks a lot for putting time into this @Ghostavio 🙇 Unfortunately, as mentioned in #177, `macos-x88-64` support comes with a few caveats, one of which you also observed [here](https://github.com/SonarSource/sonarqube-cli/pull/184#issuecomment-4245130457) :) 

Given that the effort is not trivial AND
> As of April 2026, the x86-64 (Intel) Mac platform is officially in its “sunset” phase. 

> Apple has officially confirmed that macOS Tahoe (released in 2025) is the final major version of macOS to support Intel-based Macs.

We've decided not to accept this contribution.

We still have the ticket in our backlog and might revisit it later on. 
I hope this makes sense and thanks again for using our products!

Sophio

PR #192, opened 2026-04-17
CLI-181 Adjust token generation URL for SQS 2026.2+

Krosovok:


src/lib/server-info.ts
```diff
@@ -0,0 +1,80 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { version as VERSION } from '../../package.json';
+import { CommandFailedError } from '../cli/commands/_common/error';
```
kirill-knize-sonarsource: `lib` module shouldn't depend on `cli`. `lib` supposed to be a leaf dependency, that doesn't depend on anything. We need to move error type here, if we thro it from here.

kirill-knize-sonarsource (changes requested):
Potential circular dependency later.

src/lib/server-info.ts
Krosovok: `CommandFailedError` seems to be very CLI-specific. How about throwing a more generic error in `lib` and then translate it to `CommandFailedError` on CLI level? That's how I'd approach it in Java, not sure if it makes sense from TS perspective.

kirill-knize-sonarsource: I think it's OK.

kirill-knize-sonarsource (approved):
LGTM

PR #194, opened 2026-04-17
CLI-198 Check if token is valid during `sonar auth status`

Krosovok:


tests/integration/harness/fake-sonarqube-server.ts
```diff
@@ -193,6 +193,12 @@ export class FakeSonarQubeServerBuilder {
         const bearerToken = authHeader?.startsWith('Bearer ') ? authHeader.slice(7) : null;
         const isAuthorized = !validToken || bearerToken === validToken;
 
+        if (path === '/api/authentication/validate' && req.method === 'GET') {
+          return new Response(JSON.stringify({ valid: isAuthorized }), {
+            headers: { 'Content-Type': 'application/json' },
+          });
```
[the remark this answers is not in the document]
Krosovok: It still handles POST for the same endpoint.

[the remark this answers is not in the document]
Krosovok: there's an existing test in api.test.ts that does POST /api/authentication/validate with a wrong token and expects a 401. Without the GET guard, our change would make the fake server return { valid: false } for that POST too, breaking the test.

src/cli/commands/auth/status.ts
```diff
@@ -37,15 +37,43 @@ export async function authStatus(): Promise<void> {
...
 
-  const lines = [`Server  ${conn.serverUrl}`, ...(conn.orgKey ? [`Org     ${conn.orgKey}`] : [])];
+  const serverLine = `Server  ${conn.serverUrl}`;
+  const orgLines = conn.orgKey ? [`Org     ${conn.orgKey}`] : [];
 
   if (token === null) {
-    note([...lines, '', 'Run "sonar auth login" to restore the token'], '✗ Token missing', {
-      borderColor: red,
-      titleColor: red,
+    note(
```
kirill-knize-sonarsource: It's a UI code, it should not be mixed with business logic.
Also not sure if it's correct to write essentially error text into stdout instead of stderr, but this one is TBD with UX.

src/cli/commands/auth/status.ts
```diff
@@ -37,15 +37,43 @@ export async function authStatus(): Promise<void> {
...
+      },
+    );
+    return;
+  }
+
+  const status = await withSpinner('Verifying token...', () =>
+    checkTokenStatus(conn.serverUrl, token),
+  );
+
+  if (status === 'valid') {
```
kirill-knize-sonarsource: Same

src/cli/commands/auth/status.ts
```diff
@@ -37,15 +37,43 @@ export async function authStatus(): Promise<void> {
...
+    note([serverLine, ...orgLines], '✓ Connected', {
+      borderColor: green,
+      titleColor: green,
       contentColor: dim,
     });
+  } else if (status === 'invalid') {
+    note(
```
kirill-knize-sonarsource: Same

src/cli/commands/auth/status.ts
```diff
@@ -37,15 +37,43 @@ export async function authStatus(): Promise<void> {
...
+      [serverLine, ...orgLines, '', 'Run "sonar auth login" to reauthenticate'],
+      '✗ Token invalid',
+      { borderColor: red, titleColor: red, contentColor: dim },
+    );
   } else {
-    note(lines, '✓ Connected', { borderColor: green, titleColor: green, contentColor: dim });
+    note(
```
kirill-knize-sonarsource: Same

kirill-knize-sonarsource (changes requested):
UI code in business logic. Writing error messages into stdout.

src/cli/commands/auth/status.ts
Krosovok: Moved it out. Also moved NoteOptions presets to colors.ts (which we might rename to "styles" or something as it already contained STATUS_ICONS for example).

src/cli/commands/auth/status.ts
```diff
@@ -18,34 +18,73 @@
...
+    throw new CommandFailedError('Token missing');
   }
+
+  const status = await withSpinner('Verifying token...', () =>
+    checkTokenStatus(conn.serverUrl, token),
+  );
+  displayTokenStatus(conn.serverUrl, conn.orgKey, status);
+
+  if (status === 'invalid') throw new CommandFailedError('Token invalid');
+  if (status === 'unreachable') throw new CommandFailedError('Server unreachable');
```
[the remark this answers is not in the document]
Krosovok: Yes. It adds some duplication, but it makes sense to me. The status display is consistent (we render it the same way for all results) and we add a message for non-0 exit code to be consistent with the rest of exceptional flows. That's how it looks like:
image
@kirill-knize-sonarsource your take?

tests/integration/harness/fake-sonarqube-server.ts
```diff
@@ -197,6 +197,12 @@ export class FakeSonarQubeServerBuilder {
...
+        }
```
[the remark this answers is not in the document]
Krosovok: there's an existing test in api.test.ts that does POST /api/authentication/validate with a wrong token and expects a 401. Without the GET guard, our change would make the fake server return { valid: false } for that POST too, breaking the test.

src/cli/commands/auth/status.ts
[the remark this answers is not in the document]
kirill-knize-sonarsource: I think it's a question to @cliffordgoh-sonar

src/cli/commands/auth/status.ts
```diff
@@ -20,32 +20,69 @@
...
 }
```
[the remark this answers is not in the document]
Krosovok: For now we don't distinguish much between failure codes.

src/cli/commands/_common/token.ts
```diff
@@ -35,16 +35,26 @@ const HTTP_STATUS_METHOD_NOT_ALLOWED = 405;
...
+
+export async function checkTokenStatus(serverURL: string, token: string): Promise<TokenStatus> {
+  try {
+    const client = new SonarQubeClient(serverURL, token);
+    return await client.checkTokenValidity();
+  } catch (err) {
+    // checkTokenValidity() lets HTTP errors propagate — any non-200 response or
+    // network failure is treated as a connectivity issue rather than an auth failure,
+    // because /api/authentication/validate always returns HTTP 200 per the SonarQube API contract.
+    logger.debug(`Token validation failed for ${serverURL}: ${(err as Error).message}`);
```
kirill-knize-sonarsource, line 48: In theory `err` is not always `Error`. I think here's it's safe, but in general we can get an undefined here.
So clean way is to check the type before casting. Not a huge problem because in our flow it never happen in reality.

kirill-knize-sonarsource (approved):
LGTM

src/cli/commands/_common/token.ts
Krosovok: Then I will merge as is.

PR #195, opened 2026-04-20
CLI-154 Make sure that we don't duplicate hooks

eray-felek-sonarsource:


src/cli/commands/integrate/claude/index.ts
```diff
@@ -188,6 +199,39 @@ function validateConfiguration(project: DiscoveredProject, config: Configuration
...
+  }
+}
+
+/**
+ * Print the scope-aware outcome after hook installation completes.
+ * When project-level setup was skipped, the earlier info() already communicated
+ * the outcome, so nothing more is printed here.
+ */
+function reportHookInstallationOutcome(isGlobal: boolean, globalHookPreexists: boolean): void {
+  if (globalHookPreexists) return;
```
kirill-knize-sonarsource: Silent return. We need a message here.
```
else if (globalHookPreexists) {
    success('Claude integration configured. Secrets scanning will use the existing global hook.');
} 
```
Exact message and if we need to print the hook location - TBD with UX.

src/cli/commands/integrate/claude/hooks.ts
```diff
@@ -158,19 +162,42 @@ export async function areHooksInstalled(hooksRoot: string): Promise<boolean> {
...
       ),
     );
+
+    if (!hasSettingsEntry) {
+      return false;
+    }
+
+    // Verify the backing script directory still exists; a missing directory means the
+    // install was partially deleted and should not block a fresh project-level install.
+    const secretsHookDir = join(
```
kirill-knize-sonarsource: It means if the global hook directory is deleted but the `settings.json` entry remains, areHooksInstalled returns false, triggering a project-level install — which results in the exact duplication the PR is trying to prevent.
Shall we at least report broken state to the user? Was it discussed with UX?

kirill-knize-sonarsource (changes requested):

src/cli/commands/integrate/claude/index.ts
cliffordgoh-sonar: Lets print hook location
`Claude integration configured. Secrets scanning will use the existing global hook at: ${globalHookPath}`

src/lib/migration.ts
```diff
@@ -177,11 +187,17 @@ export async function runMigrations(
...
+    // When skipSecretsHooks is set, a global hook already exists — leave the
+    // project-level scripts alone rather than re-materializing them.
+    if (!skipSecretsHooks) {
+      migrateHookScripts(projectRoot, globalDir);
+    }
 
-    // Install new PostToolUse hook
-    await installHooks(projectRoot, globalDir, installSqaa, projectKey);
+    // Install new PostToolUse hook (and refresh secrets hooks unless skipped)
+    await installHooks(projectRoot, globalDir, installSqaa, projectKey, { skipSecretsHooks });
```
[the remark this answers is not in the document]
eray-felek-sonarsource: The purpose of this ticket is to prevent new hooks being written if global exist. I'd consider this case out of scope

tests/unit/cli/commands/self-update/migration.test.ts
```diff
@@ -266,12 +280,14 @@ describe('runMigrations — migration execution', () => {
     await runMigrations('/some/project');
 
     const projectExts = state.agentExtensions.filter(
-      (e) =>
+      (e): e is HookExtension =>
+        e.kind === 'hook' &&
         e.name === 'sonar-secrets' &&
         e.kind === 'hook' &&
         e.hookType === 'PreToolUse' &&
         e.projectRoot === '/some/project',
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: This comment is valid.

src/cli/commands/integrate/claude/hooks.ts
```diff
@@ -145,33 +145,94 @@ async function installHook(params: HookInstallParams): Promise<void> {
...
   }
 }
 
+/**
+ * Returns the absolute path of the installed sonar-secrets hook directory,
+ * or `null` when no complete installation is detected.
+ *
+ * Thin wrapper around {@link detectSecretsHook}.
+ */
+export async function findInstalledSecretsHookPath(hooksRoot: string): Promise<string | null> {
```
kirill-knize-sonarsource: Why do we need this function? We only call it from tests.

tests/unit/cli/commands/integrate/claude/hooks.test.ts
```diff
@@ -77,25 +92,89 @@ describe('areHooksInstalled', () => {
...
+
+  afterEach(() => {
+    existsSyncSpy.mockRestore();
+    readFileSpy.mockRestore();
+  });
+
+  it('returns null when settings.json does not exist', async () => {
+    existsSyncSpy.mockReturnValue(false);
+
+    const result = await findInstalledSecretsHookPath(PROJECT_ROOT);
```
kirill-knize-sonarsource: It is not an integration test, IT supposed to call binary not a function from the code. It's a unit test that is testing function that we never call from production code. Something is strange here.

kirill-knize-sonarsource (changes requested):
We need some changes here.

src/cli/commands/integrate/claude/hooks.ts
eray-felek-sonarsource: It is a wrapper that was not needed. Removed now.

tests/unit/cli/commands/integrate/claude/hooks.test.ts
eray-felek-sonarsource: Tests are removed since they are no longer valid for a wrapper that doesn't exist

src/cli/commands/integrate/claude/index.ts
```diff
@@ -71,7 +71,22 @@ export async function integrateClaude(
...
+  const globalSecretsHookState = isGlobal ? null : await detectSecretsHook(homedir());
+  if (globalSecretsHookState?.kind === 'orphaned') {
+    warn(
+      `WARNING: Global hook configuration detected, but the source files are missing at ${globalSecretsHookState.hookDir}. Falling back to local project installation`,
+    );
+  }
+  const globalSecretsHookPath =
+    globalSecretsHookState?.kind === 'installed' ? globalSecretsHookState.hookDir : null;
+  const globalSecretsHookPreexists = globalSecretsHookPath !== null;
+  const skipSecretsHooks = globalSecretsHookPreexists;
```
kirill-knize-sonarsource: It's a strange chain of aliases for one boolean `globalSecretsHookPath !== null`

kirill-knize-sonarsource (approved):
Minor comment, else LGTM

src/cli/commands/integrate/claude/hooks.ts
eray-felek-sonarsource: Discussed and updated according to UX feedback

PR #208, opened 2026-04-21
CLI-307 support linux ARM64

sophio-japharidze-sonarsource:
Add support for Linux ARM64 architecture

### DIFF
--- a/user-scripts/install-prerelease.sh
+++ b/user-scripts/install-prerelease.sh
@@ -12,20 +12,32 @@ trap cleanup EXIT
 
 BASE_URL="https://repox.jfrog.io/artifactory/sonarsource-public-builds/org/sonarsource/cli/sonarqube-cli"
 
-detect_platform() {
+detect_os() {
   local os
   os="$(uname -s)"
   case "$os" in
-    Linux*)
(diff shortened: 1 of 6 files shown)


user-scripts/install-prerelease.sh
```diff
@@ -17,7 +17,14 @@ detect_platform() {
...
     Linux*)
-      echo "linux-x86-64"
+      case "$(uname -m)" in
+        aarch64 | arm64) echo "linux-arm64" ;;
+        x86_64 | amd64) echo "linux-x86-64" ;;
+        *)
+          echo "Unsupported Linux architecture: $(uname -m)" >&2
+          exit 1
+          ;;
+      esac
```
nquinquenel, line 38: Not sure if it's important, but the implementation of `detect_platform` in this file is different than in `install.sh`, is it ok?

nquinquenel (approved):
We're not running ARM64 in the CI AFAIK, it's worth checking if the binary works 👀

user-scripts/install-prerelease.sh
sophio-japharidze-sonarsource: Updated to use the same logic in both places.

PR #210, opened 2026-04-21
CLI-75 Revoke token when doing 'auth logout'

nquinquenel:


nquinquenel:
Done via https://github.com/SonarSource/sonarqube-cli/pull/214

PR #214, opened 2026-04-23
CLI-75 Revoke token when doing auth logout

eray-felek-sonarsource:


eray-felek-sonarsource:
Tested both on SQC and SQS

src/cli/commands/integrate/claude/repair.ts
```diff
@@ -47,6 +51,40 @@ export async function repairToken(serverURL: string, organization?: string): Pro
 
   // Save to keychain
   await saveToken(serverURL, newToken, organization);
+
+  // Persist the fresh tokenName on the active connection so that the
```
kirill-knize-sonarsource: This description and similar ones are too long. I think we can do them more brief, if we need them at all.

src/cli/commands/auth/login.ts
```diff
@@ -49,13 +57,22 @@ export async function authLogin(options: AuthLoginOptions): Promise<void> {
...
 
   const connection = addOrUpdateConnection(state, server, isCloud ? 'cloud' : 'on-premise', {
     orgKey: org,
     region: cloudRegionFromUrl(server),
+    // Manual `--with-token` logins are out of scope for server-side revocation
+    // (we have no reliable way to know the token's server-side name), so never
+    // inherit a stale tokenName from a prior browser-OAuth session — that would
+    // re-arm the revoke path against a name that no longer matches the secret
+    // now stored in the keychain.
+    tokenName: isNonInteractive ? undefined : (tokenName ?? existingConnection?.tokenName),
```
kirill-knize-sonarsource: The comment explains why --with-token must not inherit a stale tokenName, but the `?? existingConnection?.tokenName` fallback does exactly that for the browser-OAuth path: if the server generates a new token but omits name from the callback, logout will attempt to revoke the old name while the new token stays active on the server.           
                                          
Is this fallback intentional? If the server always includes name in the сallback for new tokens, this branch is dead code and could be simplified to just tokenName.

kirill-knize-sonarsource (approved):
Minor comments, else LGTM

src/cli/commands/auth/login.ts
eray-felek-sonarsource: I think it's safe to assume for server to always include name, updated accordingly

PR #223, opened 2026-04-28
CLI-321 introduce 'run mcp' command

sophio-japharidze-sonarsource:


src/lib/project-workspace/project-info.ts
```diff
@@ -154,10 +156,11 @@ export async function discoverProject(startDir: string): Promise<DiscoveredProje
   const config: DiscoveredProject = {
     rootDir: projectInfo.root,
     isGitRepo: projectInfo.isGitRepo,
+    configSources: [],
   };
 
   if (projectInfo.hasSonarProps && projectInfo.sonarPropsData) {
-    print('Found sonar-project.properties');
```
sophio-japharidze-sonarsource, line 160: This change is needed so that project discovery does not pollute `stdio` for mcp

src/cli/commands/integrate/claude/index.ts
```diff
@@ -67,6 +67,9 @@ export async function integrateClaude(
   blank();
 
   const project = await discoverProject(process.cwd());
+  for (const configSource of project.configSources) {
+    print(`Found ${configSource}`);
+  }
```
nquinquenel, line 72: Is this wanted in scope of this PR? Was it for debug only?

src/cli/commands/run/mcp.ts
```diff
@@ -0,0 +1,74 @@
...
+  await new Promise<void>((resolve, reject) => {
+    const child = spawn(config.command, config.args, {
+      stdio: 'inherit',
+      env: { ...process.env, ...config.env },
+    });
+
+    child.on('error', reject);
+    child.on('exit', (code) => {
+      process.exitCode = code ?? 1;
+      resolve();
```
nquinquenel, line 71: I think it should not `resolve()` for non 0 exit code, maybe reject with a `CommandFailedError` message?

src/lib/mcp/server-config.ts
```diff
@@ -0,0 +1,98 @@
...
+    'SONARQUBE_TOKEN',
+    '-e',
+    'SONARQUBE_URL',
+  ];
+  const env: Record<string, string> = { SONARQUBE_TOKEN: token, SONARQUBE_URL: serverUrl };
+
+  if (auth.connectionType === 'cloud') {
+    args.push('-e', 'SONARQUBE_ORG');
+    env.SONARQUBE_ORG = org ?? '';
+  }
```
nquinquenel, line 68: If we want to optimize, `SONARQUBE_URL` is not needed in scope of SQC (it defaults to sonarcloud.io)

src/lib/mcp/server-config.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  if (context.discoveredProjectKey) {
+    args.push('-e', 'SONARQUBE_PROJECT_KEY');
+    env.SONARQUBE_PROJECT_KEY = context.discoveredProjectKey;
+  }
+
+  if (context.withFsMount) {
+    const hostPath = normalizePath(context.projectRoot);
+    args.push('-v', `${hostPath}:/app/mcp-workspace:ro`);
```
nquinquenel, line 77: I know it's not ideal but CAG requires to have `:rw` access (write included). Worth checking what we want to do, it's maybe not a good practice to put a write access via a wrapper command that is doing opaque things.

nquinquenel, line 77: I know it's not ideal but CAG requires to have `:rw` access (write included). Worth checking what we want to do, it's maybe not a good practice to put a write access via a wrapper command that is doing opaque things.

src/lib/mcp/server-config.ts
```diff
@@ -0,0 +1,98 @@
...
+    '-e',
+    'SONARQUBE_TOKEN',
+    '-e',
+    'SONARQUBE_URL',
+  ];
+  const env: Record<string, string> = { SONARQUBE_TOKEN: token, SONARQUBE_URL: serverUrl };
+
+  if (auth.connectionType === 'cloud') {
+    args.push('-e', 'SONARQUBE_ORG');
+    env.SONARQUBE_ORG = org ?? '';
```
nquinquenel, line 67: I think it should probably fail if we don't have an org defined. Currently, we won't be able to start the server properly.

tests/integration/specs/run/mcp.test.ts
```diff
@@ -0,0 +1,232 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// Integration tests for `sonar run mcp`
```
nquinquenel, line 21: Should we test read only/toolsets options here?

nquinquenel (commented):
Looks good overall, few comments

src/cli/commands/integrate/claude/index.ts
sophio-japharidze-sonarsource: No, it's needed. Because before this print line was inside `discoverProject` method that is now also called from `runMcp`. And any print lines would interfere with the stdio.

Additionally, these prints do not belong in the `discoverProject` as it is an utility method that should not affect the CLI UX.

src/lib/mcp/server-config.ts
sophio-japharidze-sonarsource: I think it's ok to keep it because like this we will always explicitly set the URL and will not need extra checks saying something like `isCloud && region==US`, etc.

src/lib/mcp/server-config.ts
sophio-japharidze-sonarsource: I have changed auth-resolver to NOT resolve auth if connection type is `cloud` and org key is not defined.

src/cli/commands/run/mcp.ts
sophio-japharidze-sonarsource: Rejecting here would add another redundant error message to stdout (through generic CommandFailedError handler). When container exists with non-0 exit code, this command will exit too, with the same exit code as the continer.

src/lib/mcp/server-config.ts
sophio-japharidze-sonarsource: I don't have answer from them yet but I think it's safe to keep `:ro` for now - this is anyway the existing behaviour with Claude integration. If we want to change it later, it is straightforward.

nquinquenel (approved):

PR #224, opened 2026-04-28
CLI-367 Refactor secrets binary download to be reusable

georgii-borovinskikh-sonarsource:


damien-urruty-sonarsource (approved):
LGTM, I wonder if some other pieces will need to move. I think our file structure is still a bit blurry: we now have `cli/commands/_common/install/binary.ts`, but on the other hand we also have lib/sonarsource-releases.ts and signatures.ts, which are all related to the integration with binaries.sonarsource.com. I think this PR prepares well the ground for SCA, but after it's integrated we can think about some reorganization

PR #225, opened 2026-04-28
CLI-359 CLI-313 Analyze local change set with SQAA

nquinquenel:
This PR includes:
- Analyze with local change set
- Standardization of some of the error code (e.g., wrong flag should always be 2)

### DIFF
--- a/src/cli/commands/analyze/sqaa-changeset.ts
+++ b/src/cli/commands/analyze/sqaa-changeset.ts
@@ -21,7 +21,7 @@
 // Resolves the set of local files to analyze from Git, honouring .gitignore.
 
 import { closeSync, openSync, readSync, statSync } from 'node:fs';
-import { join } from 'node:path';
+import { join, resolve } from 'node:path';
 
 import { spawnProcess } from '../../../lib/process';
 import { CommandFailedError } from '../_common/error';
@@ -87,7 +87,7 @@ export async function resolveChangeSet(
  */
 async function resolveRepoRoot(cwd: string): Promise<string> {
   const out = await runGit(['rev-parse', '--show-toplevel'], cwd);
-  return out.trim();
+  return resolve(out.trim());
 }
 
 async function getDiffFiles(
--- a/src/cli/commands/analyze/sqaa-changeset.ts
+++ b/src/cli/commands/analyze/sqaa-changeset.ts
@@ -46,11 +46,17 @@ export interface IgnoredFile {
 export interface ChangeSetResult {
   files: string[];
   ignored: IgnoredFile[];
+  /** Repository root used to resolve git output. Useful for computing relative paths. */
+  repoRoot: string;
 }
 
 /**
  * Resolves the list of absolute file paths that belong to the local change set,
  * filtering out git-ignored paths and binary files, capped at SQAA_MAX_FILE_BYTES per file.
+ *
+ * All git commands are run from the repository top-level (resolved via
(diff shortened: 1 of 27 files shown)


src/cli/commands/analyze/sqaa-changeset.ts
```diff
@@ -0,0 +1,139 @@
...
+      return false;
+    }
+  });
+}
+
+/** Excludes files that exceed the per-file size limit. */
+function enforceMaxSize(files: string[]): string[] {
+  return files.filter((f) => {
+    try {
+      return statSync(f).size <= SQAA_MAX_FILE_BYTES;
```
damien-urruty-sonarsource: Small concern with this file is that we silently remove files from the changeset, and users are not aware. Now that you have a nice viz with file names in the terminal, we could display them and simply mention 'IGNORED' or 'BINARY', but I guess the most surprising lack of information for users would be too big files

src/cli/commands/analyze/sqaa.ts
```diff
@@ -30,54 +30,372 @@ import type { HookExtension } from '../../../lib/state';
...
+const COUNTDOWN_TICK_MS = 1000;
+
+/** Cloud authentication context required for SQAA API calls. */
+interface CloudAuth {
+  serverUrl: string;
+  token: string;
+  orgKey: string;
+}
 
 export interface AnalyzeSqaaOptions {
```
damien-urruty-sonarsource, line 54: Don't we have a JSON output format? I think it would be a nice addition for later, to be able to pipe the output of this command to something else

src/cli/commands/analyze/sqaa.ts
```diff
@@ -30,54 +30,372 @@ import type { HookExtension } from '../../../lib/state';
...
+
+  if (!process.stdout.isTTY) {
+    return true;
+  }
+
+  blank();
+  const confirmed = await confirmPrompt('Do you wish to proceed?');
+  if (!confirmed) {
+    blank();
+    text('Analysis cancelled. Use --force to skip this prompt.');
```
damien-urruty-sonarsource: I don't understand the "skip this prompt." part. Is it about not having the check for the number of files? Maybe something like "Analysis cancelled. Use --force to skip/bypass the check for the number of files.". I think my concern is about the word 'prompt' here

src/cli/commands/analyze/sqaa.ts
```diff
@@ -30,54 +30,372 @@ import type { HookExtension } from '../../../lib/state';
...
+  const progress = new SqaaProgress({ files: allPaths });
+  const ctx: BatchContext = { files, allPaths, cloudAuth, projectKey, branch, progress };
+  const tally = await runBatches(ctx);
+
+  progress.finish(tally.allResults.length);
+  printFileDetails(tally.allResults);
+  printSummary(tally.totalIssues, tally.totalErrors, tally.totalFailures);
+}
+
+async function runBatches(ctx: BatchContext): Promise<BatchTally> {
```
damien-urruty-sonarsource: Nitpick: I wonder if you could extract the batching-related code to a separate file. I think it makes the interesting logic about sqaa invocation a bit buried in the middle. Also this batching functionality might be reusable in other contexts? But maybe it's a bit too soon to make it generic

damien-urruty-sonarsource (approved):
LGTM some suggestions

src/cli/commands/analyze/sqaa.ts
nquinquenel: Added

PR #228, opened 2026-04-29
CLI-158 Replace legacy pre-commit configuration when encountered.

Krosovok:
I split `.pre-commit-config.yaml` update into separate `read`->`checkLegacy`->`addHook`->`write` so that it's easier to separate UI from actions. We needed to notify the user that the old hook was deleted and I didn't want to neither do it deep inside logic methods nor read-write twice.
That led to quite a shake up on tests so now they mostly test without the file outside of few high level scenarios.

### DIFF
--- a/src/cli/commands/integrate/git/git-precommit-framework.ts
+++ b/src/cli/commands/integrate/git/git-precommit-framework.ts
@@ -26,12 +26,13 @@ import { join } from 'node:path';
 import yaml from 'js-yaml';
 
 import { spawnProcess } from '../../../../lib/process';
-import { success } from '../../../../ui';
+import { success, text } from '../../../../ui';
 import { CommandFailedError } from '../../_common/error';
 import type { GitHookType } from '.';
 
 export const PRE_COMMIT_CONFIG_FILE = '.pre-commit-config.yaml';
 const PRE_COMMIT_SONAR_HOOK_ID = 'sonar-secrets';
+export const PRE_COMMIT_LEGACY_REPO = 'https://github.com/SonarSource/sonar-secrets-pre-commit';
 
 interface PreCommitHookEntry {
   id: string;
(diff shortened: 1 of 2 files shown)


src/cli/commands/integrate/git/git-precommit-framework.ts
```diff
@@ -83,35 +83,48 @@ function isSonarHookEntry(hookEntry: unknown): hookEntry is PreCommitHookEntry {
...
+    localRepo.hooks[idx] = sonarHook;
   } else {
     config.repos.push({ repo: 'local', hooks: [sonarHook] });
   }
-  writeFileSync(configPath, yaml.dump(config, { lineWidth: -1 }), 'utf-8');
+}
+
+/** Return true if .pre-commit-config.yaml already contains the sonar-secrets local hook. */
+export function hasSonarHookInPreCommitConfig(root: string): boolean {
+  if (!existsSync(join(root, PRE_COMMIT_CONFIG_FILE))) return false;
```
sophio-japharidze-sonarsource: I think we try to avoid inline ifs like this, without braces.

Krosovok: Good catch! Do we have a code style somewhere? I found only instructions about [imports and stuff](https://github.com/SonarSource/sonar-developer-toolset#manual-javascripttypescript-code-style).

sophio-japharidze-sonarsource: the link you shared is also the one I know :D

sophio-japharidze-sonarsource (commented):
The implementation looks good and will work, but I think it can be simplified. Why not consider legacy hook directly inside `hasSonarHookInPreCommitConfig` & `isSonarHooksEntry`? Is it because you want to explicitly log that legacy config was removed?

Krosovok:
> The implementation looks good and will work, but I think it can be simplified. Why not consider legacy hook directly inside `hasSonarHookInPreCommitConfig` & `isSonarHooksEntry`? Is it because you want to explicitly log that legacy config was removed?

1. Yes, according to our principles we want to notify user when we change something, especially when it's deletion. And not to mix UI concerns with logic it should be somewhere on top level at least.
2. Specifically for `hasSonarHookInPreCommitConfig` & `isSonarHooksEntry`: they are both boolean-returning checks, I would expect them to be pure functions that do the check only, without any side-effects. Changing a file, even if for removing legacy stuff, should be exclusively intentional in code. If somebody else would want to reuse any of this checks in the future, they would not expect it to modify the file checked. If for example we would want it to do in some `status` command. User wouldn't expect it to change a file.

sophio-japharidze-sonarsource (approved):

PR #229, opened 2026-04-29
CLI-323 configure MCP for Copilot integration; Refactor Claude MCP

sophio-japharidze-sonarsource:


src/cli/commands/integrate/claude/index.ts
```diff
@@ -140,7 +140,12 @@ export async function integrateClaude(
   });
   reportHookInstallationOutcome(isGlobal, globalSecretsHookPath);
 
-  await setupMcpServer('claude', project.rootDir, isGlobal, auth, project.projectKey);
+  await setupMcpServerForAgent(
+    'claude',
+    project.rootDir,
+    isGlobal,
+    options.project ?? project.projectKey,
+  );
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: This observation is correct. At the same time, we are not yet GA and breaking changes are expected. I think we should aim for our code to reflect the state that is relevant for v1.0. If we put complex migrations now, it will stay there for a long time.

Additionally, if MCP config is present in both `~/.claude` & inside `.mcp.json`, only one of them is being applied, not both.

src/cli/commands/integrate/_common/mcp.ts
```diff
@@ -0,0 +1,52 @@
...
+import { CLI_COMMAND } from '../../../../lib/config-constants';
+import {
+  getMcpConfig,
+  getMcpConfigFilePath,
+  writeMcpServerEntry,
+} from '../../../../lib/mcp/mcp-helper';
+import { error, info, success } from '../../../../ui';
+
+export async function setupMcpServerForAgent(
+  agent: 'claude' | 'copilot',
```
damien-urruty-sonarsource: I don't think having a common method for all MCP configs will work in the long run. Each Agent comes with different syntax or subtleties in their MCP config files. But for now it should be fine

src/cli/command-tree.ts
```diff
@@ -139,7 +139,7 @@ integrateCommand
   .description(
     'Setup SonarQube integration for Claude Code. This will install secrets scanning hooks, configure SonarQube Agentic Analysis and MCP Server.',
   )
-  .option('-p, --project <project>', 'Project key')
+  .option('-p, --project <project>', 'Project key. Ignored when --global is used.')
```
damien-urruty-sonarsource, line 142: Maybe to discuss with Clifford, but I know for some other commands, we fail if some options are incompatible. I think it's less confusing because as a user you don't know which option 'wins'

damien-urruty-sonarsource (approved):
LGTM, I just foresee that this common MCP config will not be common for all agents, but for now it's fine. Another thing that bothers me is that we have 2 different folders that are used for shared helper functionality: `lib/` and `src/cli/commands/_common/`. It's a bit confusing, and maybe we should group them. But it's out of scope of this PR

PR #237, opened 2026-04-30
CLI-378 Run Bun unit tests in parallel

nquinquenel:


damien-urruty-sonarsource (approved):
That's great! Could we generalize it to other kind of tests?

nquinquenel:
> That's great! Could we generalize it to other kind of tests?

As mentioned in the ticket, I don't feel comfortable adding this for integration tests or E2E tests. I did not notice an improve, and I wonder if it could even decrease the performance due to the sub-processes and other I/O operations.

PR #238, opened 2026-04-30
CLI-317 Show a warning instead of silently skipping for non-Cloud

Krosovok:


src/cli/commands/analyze/sqaa.ts
```diff
@@ -61,7 +61,12 @@ export async function runSqaaAnalysis(
   command?: Command,
 ): Promise<void> {
   const cloudAuth = resolveCloudAuth(auth, explicitProject);
-  if (!cloudAuth) return;
+  if (!cloudAuth) {
+    warn(
+      'SonarQube Agentic Analysis skipped: a SonarQube Cloud connection is required. Run: sonar auth login (ensure you connect to SonarQube Cloud)',
+    );
+    return;
```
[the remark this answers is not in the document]
Krosovok: It's not a realistic user scenario as `orgKey` is supposed to be there. Much more likely user is just connected to a server. And what's more, if they run `sonar auth login` they might just reconnect to the server if a project configs point to a server instead of cloud. So that message is important to show user what to pay attention to.

src/cli/commands/analyze/sqaa.ts
```diff
@@ -77,7 +82,7 @@ export async function runSqaaAnalysis(
 
 /**
  * Validate that the resolved auth is for SonarQube Cloud.
- * Returns null when SQAA should be silently skipped (on-premise or missing orgKey without --project).
+ * Returns null when the connection is not Cloud and --project is not set.
```
damien-urruty-sonarsource, line 85: `and` -> `or`?

tests/unit/cli/commands/analyze/analyze-sqaa.test.ts
```diff
@@ -116,11 +116,32 @@ describe('analyzeSqaa: input validation', () => {
 });
 
 describe('analyzeSqaa: auth resolution', () => {
-  it('skips SQAA when orgKey is missing from auth', async () => {
+  it('warns and skips SQAA when orgKey is missing from auth', async () => {
```
damien-urruty-sonarsource: I see little value in those unit tests if the same thing is tested via an integration test above. Do you think it's needed? I would remove those 2 tests

damien-urruty-sonarsource (approved):
LGTM, only nitpicks

src/cli/commands/analyze/sqaa.ts
Krosovok: `and` is the right one, it throws when `--project` is set to non-cloud.

tests/unit/cli/commands/analyze/analyze-sqaa.test.ts
Krosovok: You are right. We can remove them.

src/cli/commands/analyze/sqaa.ts
damien-urruty-sonarsource: Ah yes, got it! Thanks

PR #239, opened 2026-05-01
CODEFIX-1426 Interactively list and assign issues to Remediation Agent

vnaskos-sonar:
## Summary

Introduces the `sonar remediate` command (CODEFIX-1426): an interactive flow that lets developers pick from project issues fixable by the SonarQube Remediation Agent and submit them as a single job. Includes a pre-flight entitlement check so unsupported orgs fail early with a clear message instead of hitting an opaque API error mid-flow. 

## User-visible flow

1. Validate connection is SonarQube **Cloud** with an active org (the agent is not supported on Server).
2. Pre-flight `checkAiRemediationEntitlement` → exits early with one of:
 - `not_eligible` - org is not on a plan that includes the agent. 
 - `not_enabled` - eligible but `delegateIssuesEnabled` is off. 
 - `unknown` - entitlement service unreachable; user is asked to retry. 
3. Fetch open + confirmed issues filtered by `fixableByAgent=true`, sorted by severity (BLOCKER → INFO). 
4. Interactive multi-select (new `multiSelectPrompt`) → user picks one or more. 
5. Submit a single job to `/fix-suggestions/ai-agent-scheduled-jobs` with `triggerSource: "CLI"` and print the activity URL for tracking.


src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+import { CommandFailedError } from '../_common/error';
+
+export interface RemediateOptions {
+  project?: string;
+}
+
+const SEVERITY_ORDER = ['BLOCKER', 'CRITICAL', 'MAJOR', 'MINOR', 'INFO'] as const;
+
+const AI_REMEDIATION_DOCS_URL =
+  'https://docs.sonarsource.com/sonarqube-cloud/administering-sonarcloud/ai-features/sonarqube-remediation-agent';
```
nquinquenel: If we start referencing docs or links, I'd centralize everything in one place, it will help to maintain them

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  if (!auth.orgKey) {
+    throw new CommandFailedError('Cannot verify the Remediation Agent entitlements.');
+  }
+  const { status: entitlement } = await client.checkAiRemediationEntitlement(auth.orgKey);
+  if (entitlement === 'not_eligible') {
+    print(`The Remediation Agent is not available for your organization (${auth.orgKey}).`);
+    print(`Learn more: ${AI_REMEDIATION_DOCS_URL}`);
+    blank();
+    throw new CommandFailedError('The Remediation Agent is not available for this organization.');
```
nquinquenel: In all branches you print a message, and throw an error with the same message that will end up rendered twice in `sonar-command.ts` (we output the error message there). I think it's better to be consistent and let our error handler to print the error message.

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  const issuesClient = new IssuesClient(client);
+
+  // Step 1: Fetch eligible issues
+  process.stdout.write(`  Fetching issues for project ${projectKey}...`);
+
+  const issues = await fetchEligibleIssues(issuesClient, auth.orgKey, projectKey);
+
+  process.stdout.write(`  ${green('✓')}  ${issues.length} eligible issues found\n`);
```
nquinquenel: I think we already have a spinner for this kind of waiting time -> `withSpinner()`, that would probably be better

Also, why we do we write to `stdout` here and not use `print()`?

docs/data/commands.json
```diff
@@ -521,6 +522,30 @@
       ],
       "children": []
     },
+    {
+      "id": "sonar-remediate",
+      "name": "remediate",
+      "fullName": "sonar remediate",
```
nquinquenel, line 561: As I understand, the command is not agent/CI friendly, there's no `--non-interative` mode. We might want to double check with our UX if that's acceptable. Otherwise, perhaps a `--all` `--severity` flag could be interesting (just ideas)

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+
+const SEVERITY_COLORS: Record<string, (s: string) => string> = {
+  BLOCKER: red,
+  CRITICAL: red,
+  MAJOR: yellow,
+  MINOR: cyan,
+  INFO: dim,
+};
+
+export async function remediate(options: RemediateOptions, auth: ResolvedAuth): Promise<void> {
```
nquinquenel, line 49: If we're not running in TTY, the command will hang forever. We may want at least to fail fast with a `!process.stdin.isTTY` check

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+    blank();
+    print(`0 jobs created · 0 issues queued · 1 job failed`);
+    throw new CommandFailedError('Remediation job submission failed.');
+  }
+
+  const issueWord = selectedKeys.length === 1 ? 'issue' : 'issues';
+  print(`  job/${taskId}  ·  ${selectedKeys.length} ${issueWord}  ·  submitted`);
+  blank();
+  print(`1 job created · ${selectedKeys.length} ${issueWord} queued · 0 skipped`);
```
nquinquenel: I think this "dot" style is new and does not match the existing commands style, for success I think we use `discreetSuccess` for example or `success`

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+  const elispsisLength = 3;
+  const msg =
+    issue.message.length > maxDescriptionLength
+      ? `${issue.message.slice(0, maxDescriptionLength - elispsisLength)}...`
+      : issue.message;
+  return `${severity}  ${rule}  ${path} - ${msg}`;
+}
+
+function mapErrorMessage(raw: string, displayOrg: string): string[] {
+  if (raw.includes('Organization does not have allowance for AI agent jobs')) {
```
nquinquenel, line 200: Looks very fragile 😅 
Couldn't we rely on checking `err.status === 403`?

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    print(`0 jobs created · 0 issues queued · 1 job failed`);
+    throw new CommandFailedError('Remediation job submission failed.');
+  }
+
+  const issueWord = selectedKeys.length === 1 ? 'issue' : 'issues';
+  print(`  job/${taskId}  ·  ${selectedKeys.length} ${issueWord}  ·  submitted`);
+  blank();
+  print(`1 job created · ${selectedKeys.length} ${issueWord} queued · 0 skipped`);
+  blank();
+  const activityUrl = `${auth.serverUrl}/project/agent_activity?id=${projectKey}`;
```
nquinquenel: You may want to use `encodeURIComponent` here for the `projectKey`, not sure what it can contains

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    throw new CommandFailedError(
+      'sonar remediate requires SonarQube Cloud - The Remediation Agent is not supported on SonarQube Server.',
+    );
+  }
+
+  const client = new SonarQubeClient(auth.serverUrl, auth.token);
+
+  if (!auth.orgKey) {
+    throw new CommandFailedError('Cannot verify the Remediation Agent entitlements.');
+  }
```
nquinquenel: I think this is technically impossible, for `connectionType === 'cloud'`, `resolveAuth` enforces `orgKey`

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+}
+
+function formatIssueLabel(issue: SonarQubeIssue, projectKey: string): string {
+  const severityColor = SEVERITY_COLORS[issue.severity] ?? dim;
+  const severity = severityColor(issue.severity.padEnd(8));
+  const maxRuleLineLength = 14;
+  const rule = dim(issue.rule.padEnd(maxRuleLineLength));
+  const path = issue.component.replace(`${projectKey}:`, '');
+  const maxDescriptionLength = 50;
+  const elispsisLength = 3;
```
nquinquenel:
```suggestion
  const ellipsisLength = 3;
```

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    p: 1,
+  });
+  return result.issues;
+}
+
+function formatIssueLabel(issue: SonarQubeIssue, projectKey: string): string {
+  const severityColor = SEVERITY_COLORS[issue.severity] ?? dim;
+  const severity = severityColor(issue.severity.padEnd(8));
+  const maxRuleLineLength = 14;
+  const rule = dim(issue.rule.padEnd(maxRuleLineLength));
```
nquinquenel: Not 100% sure about the flow of this method, but I think the rule can be longer than 14, wouldn't that be an issue with this `padEnd`?

src/sonarqube/client.ts
```diff
@@ -308,6 +309,33 @@ export class SonarQubeClient {
...
+      const configEndpoint = `/fix-suggestions/organization-configs/${org.id}`;
+      const config = await this.get<{
+        codeReviewAgent: { organizationEligible: boolean; delegateIssuesEnabled?: boolean };
+      }>(configEndpoint, undefined, resolveFromEndpoint(this.serverURL, configEndpoint));
+
+      if (!config.codeReviewAgent.organizationEligible) return { status: 'not_eligible' };
+      if (!config.codeReviewAgent.delegateIssuesEnabled) return { status: 'not_enabled' };
+      return { status: 'ok' };
+    } catch (err) {
+      logger.debug('AI remediation entitlement check failed', err);
```
nquinquenel: Should it be warn or error instead?

nquinquenel (commented):
Looks good! Some comments but overall it's nicely done

src/sonarqube/client.ts
vnaskos-sonar: Changed to warn 👍

src/cli/commands/remediate/index.ts
vnaskos-sonar: That was removed. The issue is now two lines
image

src/cli/commands/remediate/index.ts
vnaskos-sonar: Changed to be uniform with the other commands
image

src/cli/commands/remediate/index.ts
vnaskos-sonar: I've added the check and an env variable for the tests

src/cli/commands/remediate/index.ts
vnaskos-sonar: Moved to the config-constants.ts as CLAUDE.md suggests, good catch!

docs/data/commands.json
vnaskos-sonar: There will be another ticket to add `--issues` that will skip the issue selection and act as non-interactive mode

src/cli/commands/remediate/index.ts
vnaskos-sonar: Unfortunately, the agent endpoint returns 403 for several different conditions (non-member of org, insufficient privileges, no AI-agent allowance), each with a different body string. Switching to a pure `err.status === 403` check would lose the ability to distinguish billing/allowance from access-control failures, so the user message would regress. I'll add a code comment explaining the cascade so the body-match isn't surprising. WDYT?

nquinquenel: Okay, if the limitation is documented near the code I guess it's acceptable

nquinquenel (approved):
Nice 🚀

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,211 @@
...
+    return [
+      `Your organization plan does not include the Remediation Agent (${displayOrg}).`,
+      `Learn more: ${AI_REMEDIATION_DOCS_URL}`,
+    ];
+  }
+
+  return [
+    `The Remediation Agent is not enabled for your organization (${displayOrg}).`,
+    `Learn more: ${AI_REMEDIATION_DOCS_URL}`,
+  ];
```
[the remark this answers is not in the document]
vnaskos-sonar: This will be fixed on a later PR

src/cli/commands/remediate/index.ts
```diff
@@ -0,0 +1,209 @@
...
+): Promise<SonarQubeIssue[]> {
+  const result = await issuesClient.searchIssues({
+    projects: projectKey,
+    organization: orgKey,
+    issueStatuses: 'OPEN,CONFIRMED',
+    fixableByAgent: true,
+    ps: MAX_PAGE_SIZE,
+    p: 1,
+  });
+  return result.issues;
```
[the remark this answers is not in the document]
vnaskos-sonar: This is intentional and documented

PR #241, opened 2026-05-03
CLI-364 `integrate copilot` - pre tool use hook and custom instructions for secrets

tomshafir-sonarsource:


src/cli/commands/integrate/_common/types.ts
```diff
@@ -0,0 +1,25 @@
+/*
```
sophio-japharidze-sonarsource, line 1: Small suggestion - should we call this file `types.ts` instead?

src/cli/commands/integrate/_common/state.ts
```diff
@@ -0,0 +1,122 @@
...
+import logger from '../../../../lib/logger';
+import { loadState, saveState } from '../../../../lib/repository/state-repository';
+import type { BaseAgentExtension, CliState, HookType } from '../../../../lib/state';
+import { markAgentConfigured, upsertAgentExtension } from '../../../../lib/state-manager';
+import { warn } from '../../../../ui';
+
+type ExtensionAttrs = Partial<Pick<BaseAgentExtension, 'projectKey' | 'orgKey' | 'serverUrl'>>;
+
+export type AgentExtensionSpec =
+  | {
```
sophio-japharidze-sonarsource: Let's create dedicated types for `HookExtension` & `InstructionExtension` & make `AgentExtensionSpec` be union of those two

src/cli/commands/integrate/copilot/instructions.ts
```diff
@@ -0,0 +1,88 @@
...
+  const filePath = join(dir, INSTRUCTIONS_FILENAME);
+
+  mkdirSync(dir, { recursive: true });
+  await writeFile(filePath, buildInstructionsBody(), 'utf-8');
+
+  success(`Prompt-secrets instructions installed (${filePath})`);
+}
+
+function buildInstructionsBody(): string {
+  return `# SonarQube prompt-secrets protocol
```
sophio-japharidze-sonarsource, line 101: how would we be able to update these instructions in future if we needed to? Maybe we can add a start/end markers to this section so that if needed, in future, we can easily find the text range and replace it.

tests/integration/specs/hook/hook-copilot-pre-tool-use.test.ts
```diff
@@ -0,0 +1,249 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// Integration tests for `sonar hook copilot-pre-tool-use`:
+// JSON stdin parsing, graceful skips, and end-to-end secret detection in files
+// that GitHub Copilot CLI is about to read.
+//
+// Behaviour contract (differs from the Claude hook):
```
sophio-japharidze-sonarsource, line 25: This is cool. Would be great to have it in the source file.

sophio-japharidze-sonarsource (commented):
Preliminary review

src/cli/commands/integrate/_common/state.ts
```diff
@@ -0,0 +1,126 @@
...
+export function recordAgentExtensions(
+  state: CliState,
+  agentId: string,
+  projectRoot: string,
+  isGlobal: boolean,
+  specs: AgentExtensionSpec[],
+): void {
+  const effectiveRoot = isGlobal ? homedir() : projectRoot;
+  const now = new Date().toISOString();
+  for (const spec of specs) {
```
sophio-japharidze-sonarsource: Can we rename it to `agentExtension` please?

src/cli/commands/integrate/claude/index.ts
```diff
@@ -212,35 +198,19 @@ function validateConfiguration(project: DiscoveredProject, config: Configuration
...
-function reportHookInstallationOutcome(isGlobal: boolean, globalHookPath: string | null): void {
-  if (globalHookPath !== null) {
+function reportHookInstallationOutcome(
+  isGlobal: boolean,
+  existingGlobalHookPath: string | undefined,
+): void {
+  if (existingGlobalHookPath) {
     success(
-      `Claude integration configured. Secrets scanning will use the existing global hook at: ${globalHookPath}`,
+      `Claude integration configured. Secrets scanning will use the existing global hook at: ${existingGlobalHookPath}`,
```
sophio-japharidze-sonarsource: I think this should say `Claude Code`

src/cli/commands/integrate/copilot/index.ts
```diff
@@ -57,12 +40,72 @@ export async function integrateCopilot(_auth: ResolvedAuth, options: IntegrateCo
...
+    : detectGlobalPromptSecretsInstructions();
+  const skipInstructions = !!existingGlobalInstructionsPath;
 
-  // TODO setup hooks
+  // ============
+  // Installation
+  // ============
+  await installSecretsBinary();
+  if (!skipHookInstall) {
+    await installPreToolUseHook(project.rootDir, isGlobal);
```
sophio-japharidze-sonarsource: Let's move this logic out to copilot/hooks.ts. Like that, hooks module becomes fully self-contained and responsible for the hooks installation

src/cli/commands/integrate/copilot/index.ts
```diff
@@ -57,12 +40,72 @@ export async function integrateCopilot(_auth: ResolvedAuth, options: IntegrateCo
...
+
+  reportInstallationOutcome(isGlobal, existingGlobalHookPath, existingGlobalInstructionsPath);
+}
+
+/**
+ * Print the scope-aware outcome after installation completes.
+ * Surfaces the existing global hook / instructions paths when project-level
+ * installs were skipped so the user knows where the active files live.
+ */
+function reportInstallationOutcome(
```
sophio-japharidze-sonarsource, line 69: I don't think this level of granularity is needed in the output. The function looks complex for very little value. We can simply say that integration was configured on global or project level, and log the path to the configuration. We don't need to be detailed and say "integration already existed and it's still good", instead we say "integration is good, and it's here"

src/cli/commands/integrate/copilot/instructions.ts
```diff
@@ -0,0 +1,88 @@
...
+const INSTRUCTIONS_FILENAME = 'sonarqube.instructions.md';
+const PROJECT_INSTRUCTIONS_REL_DIR = join('.github', 'instructions');
+const GLOBAL_INSTRUCTIONS_DIR = join(homedir(), '.copilot', 'instructions');
+
+/**
+ * Probe `~/.copilot/instructions/sonarqube.instructions.md` for an existing
+ * global prompt-secrets instructions file. Returns the path when present
+ * (caller should skip the project-level write), `undefined` otherwise.
+ */
+export function detectGlobalPromptSecretsInstructions(): string | undefined {
```
sophio-japharidze-sonarsource: I think we should ask PM & UX about this. There is an actual use-case where user who already has global  instruction would want to put it in the project folder - in case they want to commit it to the repository & share it with their team.

Let's ask, and track it in a followup ticket if needed. It might just be a matter of command interactivity, which is coming soon anyway.

src/cli/commands/integrate/copilot/instructions.ts
sophio-japharidze-sonarsource: Irrelevant, because we have our own instructions file. resolving.

tests/unit/cli/commands/integrate/claude/integrate.test.ts
```diff
@@ -490,20 +492,11 @@ describe('integrateCommand', () => {
...
-          c.method === 'info' &&
-          String(c.args[0]).includes(
-            'A global secrets scanning hook is already configured for SonarQube. To avoid duplicate execution, project-level secrets hooks were skipped.',
-          ),
-      );
-      expect(notice).toBeDefined();
+      // detectGlobalSecretsHook returns the hook directory path when a healthy
+      // global install is found. The user-facing info(...) message is emitted
+      // inside detectGlobalSecretsHook itself (see hooks.test.ts), so it does
+      // not fire here because the spy mocks the function away.
```
sophio-japharidze-sonarsource:
```suggestion
      // detectGlobalSecretsHook returns the hook directory path when a healthy
      // global install is found.
```
I think this comment is too detailed and makes sense _right now_ because it's fresh in our minds but will become confusing very soon. Let's keep it simple.

tests/unit/cli/commands/integrate/copilot/hooks.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    readFileSpy = spyOn(fsPromises, 'readFile').mockResolvedValue(JSON.stringify(json));
+
+    const result = await detectGlobalSecretsHook();
+
+    expect(result).toBeUndefined();
+    const noisy = getMockUiCalls().filter((c) => c.method === 'info' || c.method === 'warn');
+    expect(noisy).toHaveLength(0);
+  });
+
+  it('returns undefined and emits warn(...) when the JSON references a sonar-secrets script that does not exist (orphaned)', async () => {
```
sophio-japharidze-sonarsource: We have this exact scenario in the IT as well. please check all of the unit tests and only leave those that are not covered by ITs.

You might need to rebase on master to fix the coverage reporting issue.

tests/unit/cli/commands/integrate/copilot/hooks.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    // Forward slashes on Windows (the production code normalises backslashes).
+    const expected = `.github/hooks/sonar-secrets/build-scripts/pretool-secrets${SCRIPT_EXT}`;
+    expect(command).toBe(expected);
+    expect(command?.startsWith('/')).toBe(false);
+    if (IS_WINDOWS) {
+      expect(command?.includes('\\')).toBe(false);
+    }
+  });
+
+  it('is idempotent — re-running yields exactly one sonar-secrets preToolUse entry', async () => {
```
sophio-japharidze-sonarsource: same here

tests/unit/cli/commands/integrate/copilot/state.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function isHook(ext: AgentExtension): ext is HookExtension {
+  return ext.kind === 'hook';
+}
+
+function isInstructions(ext: AgentExtension): ext is InstructionsExtension {
+  return ext.kind === 'instructions';
+}
+
+describe('updateCopilotState', () => {
```
sophio-japharidze-sonarsource, line 30: some of the state-updating logic is also covered in ITs. We should only keep ITs.

src/cli/commands/integrate/copilot/hooks.ts
```diff
@@ -0,0 +1,175 @@
...
+): Promise<HookInstallResult> {
+  await installSecretsBinary();
+  if (!isGlobal) {
+    const existingGlobalHookPath = await detectGlobalSecretsHook();
+    if (existingGlobalHookPath) {
+      return { hookPath: existingGlobalHookPath, hookInstalled: false };
+    }
+  }
+  const hookPath = await installPreToolUseHook(projectRoot, isGlobal);
+  return { hookPath, hookInstalled: true };
```
sophio-japharidze-sonarsource: What if something goes wrong during installation like we are lacking permissions to write a file? AFAICT we are just letting the exception bubble up. We should instead wrap this in a try/catch, properly log that something went wrong and say that hook was not installed.

src/cli/commands/integrate/copilot/index.ts
```diff
@@ -57,12 +36,43 @@ export async function integrateCopilot(_auth: ResolvedAuth, options: IntegrateCo
...
   const project = await discoverProject(process.cwd());
   for (const configSource of project.configSources) {
     print(`Found ${configSource}`);
   }
+  const isGlobal = options.global ?? false;
+
+  // ============
+  // Installation
+  // ============
+  const { hookPath, hookInstalled } = await installHooks(project.rootDir, isGlobal);
```
sophio-japharidze-sonarsource, line 53: Right now we only have one hook but in future `installHooks` should probably return an array of hookPath, hookInstalled, and hookType. No need to change this now, just mentioning for visibility.

tests/integration/specs/integrate/copilot.test.ts
```diff
@@ -0,0 +1,597 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// Integration tests for `sonar integrate copilot`
+//
+// Table of contents:
+//
+//   describe('integrate copilot')
```
sophio-japharidze-sonarsource: Let's remove this. It should be easy to achieve the same overview by folding the test methods (example for VSCode-based IDEs - `Ctrl/Cmd + P -> Fold All`

sophio-japharidze-sonarsource (approved):
Looks good with small comments

PR #243, opened 2026-05-04
CLI-355 Add SQAS availability check for SCA analysis

georgii-borovinskikh-sonarsource:


src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -28,13 +30,20 @@ export interface AnalyzeDependencyRisksOptions {
...
-  _auth: ResolvedAuth,
+  auth: ResolvedAuth,
 ): Promise<void> {
   const stub = { project: options.project, risks: [] as unknown[] };
 
+  const client = new SonarQubeClient(auth.serverUrl, auth.token);
+  const enabled = await client.checkScaEnabled(auth.connectionType, auth.orgKey);
+  if (!enabled) {
+    throw new CommandFailedError(
+      'Software Composition Analysis is not available for the current server connection',
```
georgii-borovinskikh-sonarsource, line 43: I used Software Composition Analysis as that is the name of the setting on the server

tests/integration/specs/analyze/analyze-dependency-risks.test.ts
```diff
@@ -45,23 +47,65 @@ describe('analyze dependency-risks', () => {
...
+    harness.withAuth(server.baseUrl(), VALID_TOKEN, TEST_ORG);
 
     const result = await harness.run('analyze dependency-risks --project demo');
 
     expect(result.exitCode).toBe(0);
     expect(result.stdout).toContain('Project: demo');
     expect(result.stdout).toContain('(no risks)');
+    const scaCalls = server.getRecordedRequests().filter((r) => r.path === '/sca/enabled');
+    expect(scaCalls).toHaveLength(1);
+    expect(scaCalls[0].query.organization).toBe(TEST_ORG);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: it's only optional because it can be on-premise type. If auth is cloud, it's guaranteed to be non-optional

src/sonarqube/client.ts
```diff
@@ -269,6 +269,29 @@ export class SonarQubeClient {
...
+   * server. SonarCloud exposes this at `/sca/enabled?organization=<key>`
+   * (api.sonarcloud.io); SonarQube Server at `/api/v2/sca/enabled`. Any
+   * failure (404, network, unauthorized) is treated as "not available" so
+   * callers can gate analysis with a friendly error.
+   */
+  async checkScaEnabled(connectionType: 'cloud' | 'on-premise', orgKey?: string): Promise<boolean> {
+    try {
+      const isCloud = connectionType === 'cloud';
+      const endpoint = isCloud ? '/sca/enabled' : '/api/v2/sca/enabled';
+      const params = isCloud && orgKey ? { organization: orgKey } : undefined;
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: useless overly defensive check

src/cli/commands/analyze/dependency-risks.ts
damien-urruty-sonarsource: I think the official 'product' name is `SonarQube Advanced Security`, maybe we should use that? Also, your message says 'not available', I think this is imprecise. It could be that SQAS is available on the server but not enabled for a specific project. Suggestion : "SonarQube Advanced Security is not enabled for the current connection'.
Last thing, to be checked but I think for SQAA we fail softly if the feature is not enabled. If it's the case, I think we should do the same or adjust SQAA to stay consistent, maybe to be discussed with Clifford

damien-urruty-sonarsource (approved):
LGTM, I just left one suggestion for wording, but I would also recommend what should be the strategy for unmet prerequisites with Clifford

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: SQAS is the full package, SCA is a separate toggle under SQAS 🙃

georgii-borovinskikh-sonarsource: Update: had a discussion and decided to keep the error behavior as-is

PR #244, opened 2026-05-04
CLI-356 Fetch analysis properties for SCA analysis

georgii-borovinskikh-sonarsource:


src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -44,6 +44,16 @@ export async function analyzeDependencyRisks(
     );
   }
 
+  const componentExists = await client.checkComponent(options.project);
+  if (!componentExists) {
+    throw new CommandFailedError(`No project: ${options.project}`);
+  }
```
damien-urruty-sonarsource: Could we avoid this extra round-trip? Do we receive a 404 when we try to get settings for a non-existing project in the call below, and could we rely on that?

damien-urruty-sonarsource (approved):
LGTM, one suggestion to improve performance

PR #255, opened 2026-05-07
CLI-337 Rename 'analyze sqaa' to 'analyze agentic'

Krosovok:


src/cli/commands/hook/agent-post-tool-use.ts
nquinquenel, line 81: I still see some mention of SQAA, is it expected?

nquinquenel (approved):

src/cli/commands/hook/agent-post-tool-use.ts
Krosovok: Yes. Internal usage is ok, it's just user facing: command name, console output - should not use our internal acronym (since the users have no idea about it).

PR #261, opened 2026-05-11
CLI-357 Add output formatting

georgii-borovinskikh-sonarsource:
----
## Summary by Gitar

- **CLI output formatting:**
  - Added `--statuses` filter option to support `active`, `to_fix`, and `all` presets or raw status combinations
  - Implemented `formatDependencyRisksTable` to display risks with transitive chains, remediation paths, and summary statistics
  - Added `formatDependencyRisksJson` as an alternative output format
- **ViewModel refactoring:**
  - Created comprehensive `view-model` layer to normalize SCA scanner data and calculate resolution status
  - Added builder utilities for grouping risks, selecting fix versions, and aggregating package summary statistics
- **Command behavior updates:**
  - Enabled `dependency-risks` command visibility
  - Configured exit codes for unresolved risks (`51`) and analysis errors (`1`) to improve CI/CD integration
  - Added detailed validation and status definitions for risk filtering in `command-tree.ts`


src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -80,5 +94,38 @@ export async function analyzeDependencyRisks(
     new DefaultScaScannerSpawner(),
   ).run(invocation);
 
-  print(JSON.stringify({ project: options.project, ...result }, null, 2));
+  const filtered = applyStatusFilter(result, options.statusFilter);
+
+  if (options.format === 'json') {
+    print(formatDependencyRisksJson(options.project, filtered));
+  } else {
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: intentional, obvious from the usage and the wording

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -80,5 +94,38 @@ export async function analyzeDependencyRisks(
...
+    print(formatDependencyRisksTable(filtered, result.releases));
+  }
+
+  handleResult(countUnresolvedIssues(result), result.errors.length);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: there is already a comment and a todo about not having end to end tests. This part is also not possible to unit test due to current architecture

src/cli/commands/analyze/dependency-risk-helpers/table/format-dependency-risks-table.ts
```diff
@@ -0,0 +1,366 @@
...
+    appendNoRisksTail(lines);
+  }
+
+  lines.push('', '═'.repeat(MAX_LINE_WIDTH));
+  appendErrors(lines, errors);
+  appendSummaryBlock(lines, displayedReleases);
+
+  return lines.join('\n');
+}
+
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: This is intentional

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -25,22 +25,35 @@ import { type ResolvedAuth } from '../../../lib/auth-resolver';
...
 } from './dependency-risk-helpers/sca-scanner.ts';
 import { buildScaUrls } from './dependency-risk-helpers/sca-urls.ts';
+import { formatDependencyRisksTable } from './dependency-risk-helpers/table/format-dependency-risks-table.ts';
+import { buildDependencyRisksViewModel } from './dependency-risk-helpers/view-model/build/build-dependency-risks-view-model.ts';
+import type { DependencyRisksViewModel } from './dependency-risk-helpers/view-model/dependency-risks-view-model.ts';
+import type { DependencyRisksStatusFilter } from './dependency-risk-helpers/view-model/status-filter.ts';
 
 export const VALID_FORMATS = ['json', 'table'];
 
+export { DEPENDENCY_RISKS_STATUS_FILTERS as VALID_STATUS_FILTERS } from './dependency-risk-helpers/view-model/status-filter.ts';
```
damien-urruty-sonarsource: This is exported and renamed, and later imported and renamed back to the original name. We could maybe simplify

src/cli/commands/analyze/dependency-risk-helpers/view-model/status-filter.ts
```diff
@@ -0,0 +1,23 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+export const DEPENDENCY_RISKS_STATUS_FILTERS = ['including-safe', 'all', 'open', 'new'] as const;
```
damien-urruty-sonarsource: Why `including-safe` and not just `safe`? Also, shouldn't we have a 1-1 mapping with supported statuses? There are some that are not covered here

src/cli/command-tree.ts
```diff
@@ -305,11 +306,19 @@ const dependencyRisksFormatOption = new Option('--format <format>', 'Output form
   .choices(DEPENDENCY_RISKS_FORMATS)
   .default('table');
 
+const dependencyRisksStatusFilterOption = new Option(
+  '--status-filter <statusFilter>',
```
damien-urruty-sonarsource: We have a similar option to filter issue statuses in `list issues`, and the option is called `statuses`, where users can provide several ones separated by commas. I don't know what's best but we should be consistent

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -84,5 +97,50 @@ export async function analyzeDependencyRisks(
...
+    print(formatDependencyRisksTable(viewModel));
+  }
+
+  handleResult(countUnresolvedIssues(viewModel), result.errors.length);
+}
+
+function handleResult(unresolvedRisksCount: number, errorCount: number) {
+  function warnErrorsDuringAnalysis() {
+    if (errorCount > 0) {
+      warn(`Found ${errorCount} ${pluralize(errorCount, 'analysis error')}.`);
```
damien-urruty-sonarsource, line 115: How can users understand more about those analysis errors? Do we print them to logs? If so we could point them to the logs file

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -84,5 +97,50 @@ export async function analyzeDependencyRisks(
...
+    process.exitCode = EXIT_CODE_OK;
+  }
+}
+
+function pluralize(count: number, singular: string): string {
+  return `${singular}${count === 1 ? '' : 's'}`;
+}
+
+export function countUnresolvedIssues(vm: DependencyRisksViewModel): number {
+  const isUnresolved = buildRiskFilter('open');
```
damien-urruty-sonarsource: How should this play with the user-provided status filter? If users only ask for 'safe' dependency risks, should we still fail the command if there are open risks?

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -84,5 +97,50 @@ export async function analyzeDependencyRisks(
...
+    if (errorCount > 0) {
+      warn(`Found ${errorCount} ${pluralize(errorCount, 'analysis error')}.`);
+    }
+  }
+
+  if (unresolvedRisksCount > 0) {
+    warnErrorsDuringAnalysis();
+    error(
+      `Found ${unresolvedRisksCount} ${pluralize(unresolvedRisksCount, 'unresolved dependency risk')}.`,
+    );
```
damien-urruty-sonarsource, line 123: Same, should we display this message if e.g. the 'safe' filter is set

damien-urruty-sonarsource (commented):
First round of comments. The PR is big so maybe I missed some things. I think your idea with the view model and the builder is good, I'm not sure we need that many files, and you could keep the whole model in a single one. But I reckon this is more of a personal taste

src/cli/commands/analyze/dependency-risk-helpers/view-model/status-filter.ts
georgii-borovinskikh-sonarsource: RE: naming - me and Clifford couldn't really think for a better alternative
RE: statuses - it behaves like NEW = only new; OPEN = NEW + OPEN + CONFIRMED; ALL = ACCEPTED + OPEN; INCLUDING-SAFE = ALL + SAFE;
We though this is the best for usability - make safe issues appear only if added explicitly

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: They're printed to the output, as they're a part of view model. They usually have descriptions like

```
Errors:
  [MISSING_LOCKFILE] build/CopyDependencies/CopyDependencies.csproj: No lockfile was found for 'build/CopyDependencies/CopyDependencies.csproj' (nuget).
  [MISSING_LOCKFILE] src/ConnectedMode.UnitTests/Migration/FileCleanerTestCases/RulesetProp-project_key_aaa_Input.vbproj: No lockfile was found for 'src/ConnectedMode.UnitTests/Migration/FileCleanerTestCases/RulesetProp-project_key_aaa_Input.vbproj' (nuget).
```

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: Me and Clifford discussed this and agreed that it doesn't make sense to extend this gate to the resolved issues (safe and accepted), but it's limited to new issues if new issues filter is selected

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: https://github.com/SonarSource/sonarqube-cli/pull/261#discussion_r3286737861

src/cli/commands/analyze/dependency-risk-helpers/view-model/status-filter.ts
damien-urruty-sonarsource: I am a bit concerned for 2 reasons:
* I think the simplest and least confusing approach is to let users filter by statuses that exist in other products, and not come up with our own grouping, otherwise we need to explain it to users as you just did
* I find the case of OPEN particularly weird, because we use the same name as the existing status but to represent something else

As discussed in another comment and for consistency, I would advise to change the option to `statuses`, and let users provide a list of statuses they want to see. This comes with less surprise and is also more powerful because users can decide whatever combination they want

src/cli/commands/analyze/dependency-risks.ts
damien-urruty-sonarsource: Ah, right because currently you cannot filter e.g. only SAFE issues. If you could, then the command would fail for statuses different than the filter. Something to consider if you change the statuses filter

src/cli/command-tree.ts
cliffordgoh-sonar: Can we call it `status` ? Both here and in `list issues` < not sure who worked on this

src/cli/commands/analyze/dependency-risks.ts
cliffordgoh-sonar: This looks fine to me

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: I don't think it would make sense to make this gate fail for safe issues if only safe issues are selected. If I change how the filter behaves, I'm going to keep the gate behavior the same - it should only fail the command if there are issues that are not already acknowledged

src/cli/commands/analyze/dependency-risk-helpers/view-model/status-filter.ts
cliffordgoh-sonar: I initially like having grouped status, because I thought it'll be more convenient for user "just show me my active issues" & follows the IDE issues mental model.
e.g they can just type 'open' , instead of typing 'new,open,confirmed'

But I agree the point on OPEN is weird & hard to design optimally.

Throwing out an idea: how about we support both status **preset groups** & individual **raw status**?
Status group is for user INTENT, and individual status for giving user control
i.e.

> active    = new + open + confirmed   (I need to take some action/deal with this)
> resolved  = accepted + safe          (no action required, I already dealt with somehow)
> all       = everything               (no filter. entire universe of issues)
> 

so
> Status = 'new' | 'open' | 'confirmed' | 'accepted' | 'safe'
> StatusPreset = 'active' | 'resolved' | 'all'`

src/cli/command-tree.ts
damien-urruty-sonarsource: We could but then what if you want to list issues that have FP and Accept statuses? Users would have to do  requests. I think the current option works like this, I would adjust the SCA one

src/cli/commands/analyze/dependency-risk-helpers/view-model/build/fix-version-selector.ts
```diff
@@ -0,0 +1,75 @@
...
+export function selectPackageCompleteFixes(issues: AnalyzeProjectIssue[]): FixVersionVM[] {
+  const byVersion = new Map<string, VersionOption>();
+  for (const issue of issues) {
+    for (const option of issue.versionOptions ?? []) {
+      if (option.fixLevel !== 'COMPLETE') continue;
+      if (EXCLUDED_DESCRIPTION_CODES.has(option.descriptionCode)) continue;
+      if (!byVersion.has(option.version)) byVersion.set(option.version, option);
+    }
+  }
+  return sortByDescriptionCode([...byVersion.values()]).map(toFixVersionVM);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: we're not going to guard so defensively against this contract, first code is fine

src/cli/commands/analyze/dependency-risk-helpers/table/format-table-vulnerability-group.ts
```diff
@@ -0,0 +1,78 @@
...
+  hasRecommended: boolean,
+): string | null {
+  const partials = risk.partialFixes.slice(0, MAX_INLINE_PARTIAL_FIXES);
+  if (partials.length > 0) {
+    return partials
+      .map((p) => {
+        const fixed = totalVulnerabilities - p.vulnerabilityIds.length;
+        return `${p.version} (fixes ${fixed}/${totalVulnerabilities})`;
+      })
+      .join(' | ');
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: fixed this properly by using original count of vulnerabilities

src/cli/commands/analyze/dependency-risk-helpers/view-model/status-filter.ts
georgii-borovinskikh-sonarsource: I can try to implement a filter that allows for individual statuses to be provided, with a few presets like `active = new + open + confirmed` and `to-fix = active + accepted` and `all = to-fix + safe`. The default value will be `active`. WDYT? @damien-urruty-sonarsource @cliffordgoh-sonar

src/cli/command-tree.ts
georgii-borovinskikh-sonarsource: let's continue here: https://github.com/SonarSource/sonarqube-cli/pull/261#discussion_r3288006575

src/cli/commands/analyze/dependency-risk-helpers/view-model/status-filter.ts
cliffordgoh-sonar: - I am ok with default value `active`
- Can we show examples in the help menu, so that users know they can combine raw statuses and presets:

```
--status <filter>   Filter issues by status (default: active)

  Presets:   active | all
  Raw:       new | open | confirmed | accepted | safe

Examples:
    --status active
    --status new,confirmed
    --status active,safe

```

- When we run command, can we show a message to tell the user explicitly what filter was applied, i.e.:

> Filtering by: active (new, open, confirmed)

then output the result below

- we can allow preset + raw combinations (w deduplication handling), e.g. `active,safe` = "give me everything actionable, plus anything safe"
@georgii-borovinskikh-sonarsource

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: changed

src/cli/commands/analyze/dependency-risk-helpers/view-model/status-filter.ts
georgii-borovinskikh-sonarsource: added combinatorial filters with presets as discussed. added applied filters to summary.

src/cli/command-tree.ts
```diff
@@ -330,11 +330,25 @@ const dependencyRisksFormatOption = new Option('--format <format>', 'Output form
   .choices(DEPENDENCY_RISKS_FORMATS)
   .default('table');
 
+const dependencyRisksStatusFilterOption = new Option(
+  '--statuses <statuses>',
+  'Filter issues by status\n' +
+    '\n' +
+    '  Presets:   active | to_fix | all\n' +
```
damien-urruty-sonarsource: Should we document what 'active' and other presets map to?

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -84,5 +99,50 @@ export async function analyzeDependencyRisks(
...
+    process.exitCode = EXIT_CODE_OK;
+  }
+}
+
+function pluralize(count: number, singular: string): string {
+  return `${singular}${count === 1 ? '' : 's'}`;
+}
+
+export function countUnresolvedIssues(vm: DependencyRisksViewModel): number {
+  const isUnresolved = buildRiskFilter('active')!.predicate;
```
damien-urruty-sonarsource, line 138: I'm still concerned about this. If the user used `--statuses safe`, we would still count unresolved issues here and fail with an error just after. That would make the experience a bit strange, if users are only interested in safe issues, why would we fail the command if there are non-resolved risks?

Another thought related to that, it's the first analysis command where we allow users to filter by status (the `statuses` filter I was talking about earlier was for `list issues`). I don't see a big utility in allowing to filter (keep) only resolved ones. Usually when people analyze something it's because they want to find active issues. But it can stay no big deal

damien-urruty-sonarsource (approved):
Small remarks, LGTM. Should we have one or several integration tests that verifies the full output with different formats?

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: >we would still count unresolved issues here and fail with an error just after.
 
 Count happens on viewmodel, which is filtered - the counting will find no unresolved issues

damien-urruty-sonarsource: Ah I had missed that, all good then

src/cli/command-tree.ts
georgii-borovinskikh-sonarsource: Sure I will add that

PR #266, opened 2026-05-12
CLI-444 Use declarative framework for Git integration

damien-urruty-sonarsource:


src/cli/commands/integrate/git/tools/native/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    const message = error instanceof Error ? error.message : String(error);
+    throw new CommandFailedError(`Failed to run git [${message}]`, {
+      remediationHint: GLOBAL_GIT_CONFIG_REMEDIATION_HINT,
+    });
+  }
+
+  if (gitResult.exitCode !== 0) {
+    const detail = [gitResult.stderr, gitResult.stdout].filter(Boolean).join('\n');
+    throw new CommandFailedError(
+      `git config --global core.hooksPath failed (exit code ${gitResult.exitCode}): ${detail}`,
```
sophio-japharidze-sonarsource:
```suggestion
      `'git config --global core.hooksPath' failed (exit code ${gitResult.exitCode}): ${detail}`,
```

src/cli/commands/integrate/git/tools/pre-commit/config.ts
```diff
@@ -179,5 +172,17 @@ export async function installViaPreCommitFramework(root: string, hook: GitHookTy
       },
     );
   }
+}
+
+export async function installViaPreCommitFramework(root: string, hook: GitHookType): Promise<void> {
```
sophio-japharidze-sonarsource: If I understand correctly, this is now only called from tests, and should be removed completely.

sophio-japharidze-sonarsource (approved):
Great! 

One note about husky hooks - I tried running `sonar integrate git` in this repo and it wrote a new shell fragment in the `pre-commit` file. The start marker is the same - `# Sonar secrets scan - installed by sonar integrate git` so maybe we could do something with it? But I don't think it's worth the effort in case it's complicated.

PR #269, opened 2026-05-12
CLI-461 log MCP container command args in debug mode

sophio-japharidze-sonarsource:
Screenshot 2026-05-18 at 17 18 01

### DIFF
--- a/src/cli/commands/run/mcp.ts
+++ b/src/cli/commands/run/mcp.ts
@@ -25,6 +25,7 @@ import { homedir } from 'node:os';
 
 import type { ResolvedAuth } from '../../../lib/auth-resolver.js';
 import { canonicalizePath } from '../../../lib/fs-utils.js';
+import logger from '../../../lib/logger';
 import { getMcpContainerCommand, type McpServerContext } from '../../../lib/mcp/mcp-helper.js';
 import { discoverProject } from '../../../lib/project-workspace/project-info.js';
 import { detectContainerRuntime } from '../../../lib/tool-detector.js';
@@ -36,6 +37,11 @@ export interface McpRunOptions {
   readOnly?: boolean;
   toolsets?: string;
   project?: string;
+}
+
+function debugLog(message: string): void {
+  logger.debug(message);
+  process.stderr.write(`[sonarqube-cli] DEBUG ${message}\n`);
 }
 
 export async function runMcp(auth: ResolvedAuth, options: McpRunOptions = {}): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/cli/commands/run/mcp.ts
```diff
@@ -65,6 +69,13 @@ export async function runMcp(auth: ResolvedAuth, options: McpRunOptions = {}): P
 
   const config = getMcpContainerCommand(auth, runtime, context, options);
 
+  if (options.debug) {
+    debugLog(`runtime: ${runtime}`);
+    debugLog(`projectRoot: ${projectRoot ?? '(none)'}`);
+    debugLog(`projectKey: ${projectKey ?? '(none)'}`);
+    debugLog(`launching: ${config.command} ${config.args.join(' ')}`);
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: It does not. Sensitive values are part of config.env

src/cli/commands/run/mcp.ts
```diff
@@ -38,6 +39,11 @@ export interface McpRunOptions {
   project?: string;
 }
 
+function debugLog(message: string): void {
+  logger.debug(message);
+  process.stderr.write(`[sonarqube-cli] DEBUG ${message}\n`);
+}
```
nquinquenel, line 45: I think we define `warn()` and `error()`, could we define a `debug()` instead of this custom method?

src/cli/commands/run/mcp.ts
```diff
@@ -65,6 +71,13 @@ export async function runMcp(auth: ResolvedAuth, options: McpRunOptions = {}): P
 
   const config = getMcpContainerCommand(auth, runtime, context, options);
 
+  if (options.debug) {
+    debugLog(`runtime: ${runtime}`);
+    debugLog(`projectRoot: ${projectRoot ?? '(none)'}`);
+    debugLog(`projectKey: ${projectKey ?? '(none)'}`);
+    debugLog(`launching: ${config.command} ${config.args.join(' ')}`);
```
nquinquenel, line 78: Do we want to output via stderr to the MCP client? In this case, I don't think it will work as the `spawn()` happens later

src/cli/commands/run/mcp.ts
sophio-japharidze-sonarsource: This is custom because we need to output on stderr for mcp client.

src/cli/commands/run/mcp.ts
sophio-japharidze-sonarsource: Why not? This way all logs end up in MCP client output together.

nquinquenel: I think `stderr` will be carried to the MCP client via `stdio: 'inherit'`, here I'm not sure how it'd behave. Have you tried?

sophio-japharidze-sonarsource: Yes, I have tried and these logs appear along with normal MCP logs

nquinquenel (approved):

PR #270, opened 2026-05-12
CLI-456 Add SCA eligibility SQ Server version check

georgii-borovinskikh-sonarsource:
----
## Summary by Gitar

- **SCA eligibility:**
  - Added a requirement for SonarQube Server version `2026.4` or later when running SCA analysis.
  - Implemented `assertServerSupportsLocalSca` to enforce the version check before performing SCA probes.
- **Compatibility improvements:**
  - Added logic to skip the server version check for `cloud` connections.
  - Added integration tests to verify version enforcement, error handling for unknown versions, and cloud-specific bypassing.
- **Documentation:**
  - Updated the remediation hint for SCA compatibility to provide clearer guidance on edition, plan, and documentation links.

### DIFF
--- a/src/cli/commands/analyze/dependency-risks.ts
+++ b/src/cli/commands/analyze/dependency-risks.ts
@@ -166,7 +166,7 @@ async function assertServerSupportsLocalSca(
       'Software Composition Analysis is not available for the current server connection.',
       {
         remediationHint:
-          'Software Composition Analysis requires SonarQube Server Enterprise Edition or SonarQube Cloud Team plan or higher, and must be enabled by an administrator. Learn more: https://docs.sonarsource.com/sonarqube-server/latdontest/advanced-security/sca-overview/',
+          'Software Composition Analysis must be enabled by an administrator and requires an eligible SonarQube plan. Learn more: https://www.sonarsource.com/products/sonarqube/advanced-security/',
       },
     );
   }
--- a/src/cli/commands/analyze/dependency-risks.ts
+++ b/src/cli/commands/analyze/dependency-risks.ts
@@ -24,6 +24,7 @@ import { join } from 'node:path';
 import { type ResolvedAuth } from '../../../lib/auth-resolver';
 import { CLI_DIR } from '../../../lib/config-constants';
 import logger, { getLogLevelConfig } from '../../../lib/logger';
+import { fetchServerVersion, isAtLeast } from '../../../lib/server-info';
 import { SonarQubeClient } from '../../../sonarqube/client';
 import { error, print, warn } from '../../../ui';
 import { CommandFailedError, InvalidOptionError } from '../_common/error.js';
@@ -47,6 +48,8 @@ const EXIT_CODE_OK = 0;
 const EXIT_CODE_ERRORS_ONLY = 1;
 const EXIT_CODE_UNRESOLVED_RISKS = 51;
 
+const MIN_SCA_SQS_VERSION = '26.4';
+
 export interface AnalyzeDependencyRisksOptions {
   project: string;
   format: string;
@@ -63,6 +66,7 @@ export async function analyzeDependencyRisks(
   }
 
   const client = new SonarQubeClient(auth.serverUrl, auth.token);
+  await assertServerSupportsSca(auth);
   const enabled = await client.checkScaEnabled(auth.connectionType, auth.orgKey);
   if (!enabled) {
     throw new CommandFailedError(
@@ -145,4 +149,21 @@ export function countUnresolvedIssues(vm: DependencyRisksViewModel): number {
     }
   }
   return count;
+}
+
+async function assertServerSupportsSca(auth: ResolvedAuth): Promise<void> {
+  if (auth.connectionType === 'cloud') return;
+  let serverVersion: string;
+  try {
+    serverVersion = await fetchServerVersion(auth.serverUrl);
+  } catch {
+    throw new CommandFailedError(
+      `Could not determine SonarQube Server version. Software Composition Analysis requires SonarQube Server ${MIN_SCA_SQS_VERSION} or later.`,
+    );
+  }
+  if (!isAtLeast(serverVersion, MIN_SCA_SQS_VERSION)) {
+    throw new CommandFailedError(
+      `Software Composition Analysis requires SonarQube Server ${MIN_SCA_SQS_VERSION} or later (server is ${serverVersion}).`,
+    );
+  }
 }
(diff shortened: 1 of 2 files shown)


tests/integration/specs/analyze/analyze-dependency-risks.test.ts
```diff
@@ -171,4 +175,35 @@ describe('analyze dependency-risks', () => {
...
+      .withSystemStatusCode(503)
+      .start();
+    harness.withAuth(server.baseUrl(), VALID_TOKEN);
+
+    const result = await harness.run('analyze dependency-risks --project demo');
+
+    expect(result.exitCode).toBe(1);
+    expect(result.stdout + result.stderr).toContain('Could not determine SonarQube Server version');
+  });
 });
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The reviewer's main concern is incorrect. src/lib/server-info.ts:72-74 explicitly checks response.ok and throws ApiCallError on non-2xx:

  if (!response.ok) {
    throw new ApiCallError(`Server returned HTTP ${response.status} for ${url}`);
  }

  So .withSystemStatusCode(503) reliably throws and exercises the catch branch in assertServerSupportsSca — the test is passing for the right reason, just under a slightly loose label ("unreachable" vs. "returns 5xx").

  The fair part of the comment is the secondary suggestion: a 503 isn't the same path as a TCP-level failure or AbortError from the 2s timeout. Both also flow through the same catch (any thrown error → fallback), so behaviorally they're equivalent, but if you want strict coverage of the "actually
  unreachable" path, that's a separate test (e.g., point at http://127.0.0.1:1 like auth.test.ts does, or use an unbound port).

  Suggested reply to the reviewer:

  ▎ fetchServerVersion does check response.ok and throws ApiCallError on non-2xx (src/lib/server-info.ts:72), so the 503 reliably hits the catch in assertServerSupportsSca. Agreed the test name is loose — happy to rename to "5xx from /api/system/status" and/or add a second case pointing at an unbound port
  ▎ to cover the true network-failure path.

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: incorrect comment

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -47,6 +48,8 @@ const EXIT_CODE_OK = 0;
 const EXIT_CODE_ERRORS_ONLY = 1;
 const EXIT_CODE_UNRESOLVED_RISKS = 51;
 
+const MIN_SCA_SQS_VERSION = '26.4';
```
damien-urruty-sonarsource: That's a strange version name. For SQS, the version looks like 2026.4. The number you have here makes me thing about SQCB. 

OK I see now that the version is normalized in `isAtLeast`, so we should be good.

But I think the `assertServerSupportsSca` would mistakenly return true in two cases:
* user is running on SQCB 26.4+
* user is running on SQS DE 26.4+

Maybe there is a way to check editions? Since you already check if sca is enabled just below, I wonder if this check is needed at all. The SCA enabled is the only reliable indicator

georgii-borovinskikh-sonarsource: rearranged checks and renamed the method

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -146,3 +140,34 @@ export function countUnresolvedIssues(vm: DependencyRisksViewModel): number {
...
+      );
+    }
+  }
+  const enabled = await client.checkScaEnabled(auth.connectionType, auth.orgKey);
+  if (!enabled) {
+    throw new CommandFailedError(
+      'Software Composition Analysis is not available for the current server connection.',
+      {
+        remediationHint:
+          'Use a connection where SCA is enabled, or verify your server edition, plan, and organization entitlements.',
```
damien-urruty-sonarsource: Could you be a bit more specific here? 'Verify your edition or plan' doesn't say a lot about what is required. For the remediation agent we started introducing 'Learn more' links, we could use the same approach? Or at least explicitly mention requirements

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -146,3 +140,34 @@ export function countUnresolvedIssues(vm: DependencyRisksViewModel): number {
...
+    try {
+      serverVersion = await fetchServerVersion(auth.serverUrl);
+    } catch {
+      throw new CommandFailedError(
+        `Could not determine SonarQube Server version. Running Software Composition Analysis from this CLI requires SonarQube Server ${MIN_SCA_SQS_VERSION} or later.`,
+      );
+    }
+    if (!isAtLeast(serverVersion, MIN_SCA_SQS_VERSION)) {
+      throw new CommandFailedError(
+        `Running Software Composition Analysis from this CLI requires SonarQube Server ${MIN_SCA_SQS_VERSION} or later (server is ${serverVersion}).`,
```
damien-urruty-sonarsource, line 159: This message would give:
> Running Software Composition Analysis from this CLI requires SonarQube Server 26.4

Which will never be a thing. Can we just change the variable to 2026.4?

damien-urruty-sonarsource (approved):

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: I don't want to include specific product names as that can change; Should I just use https://www.sonarsource.com/products/sonarqube/advanced-security/ ?

src/cli/commands/analyze/dependency-risks.ts
damien-urruty-sonarsource: It would also help to gate SQCB from this step

georgii-borovinskikh-sonarsource: I will change this, however gating against sqcb would require me to remove normalization of the target version from `isAtLeast` - I'm not sure this is needed/can be done safely

damien-urruty-sonarsource: ah yes, correct. I don't know why this normalization was implemented

src/cli/commands/analyze/dependency-risks.ts
damien-urruty-sonarsource: It's a bit marketing, but it could do the job

georgii-borovinskikh-sonarsource: The other doc links are specific to SQS or SQC - I think I forgot to contact docs to create SCA-related pages. 
I will keep the marketing link for now, but I will create a ticket to replace it with SQ CLI docs link

georgii-borovinskikh-sonarsource: sending an email to the docs squad and https://sonarsource.atlassian.net/browse/CLI-521

PR #273, opened 2026-05-13
CLI-279 Fail soft on hook crash outside CI

kirill-knize-sonarsource:


src/cli/commands/_common/token.ts
```diff
@@ -100,7 +100,7 @@ export function buildAuthURL(serverURL: string, port: number, serverVersion?: st
 
 /**
  * Open browser, with fallback message if it fails.
- * Skipped when CI=true — token must be delivered directly to the loopback server.
+ * Skipped in CI — token must be delivered directly to the loopback server.
  */
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: https://sonarsource.atlassian.net/browse/CLI-463

src/cli/commands/hook/hook-dependencies.ts
```diff
@@ -22,14 +22,23 @@
...
   binaryPath: string;
 }
 
+export function shouldBlockOnScanError(context: 'Commit' | 'Push', err: Error): boolean {
+  if (isEnvBasedAuth()) return true;
+  warn(
+    `Secrets scan failed. ${context} is not blocked, but secrets were not checked. Reason: ${err.message}`,
+  );
+  return false;
+}
```
Krosovok: I don't like mixing a boolean-returning method with a side effect. And we use the result to throw seemingly identical `CommandFailedError` anyway. 

So maybe just move the error throw here and get rid of the return value?

src/cli/commands/hook/git-pre-push.ts
```diff
@@ -41,23 +41,27 @@ export async function gitPrePush(): Promise<void> {
...
   for (const ref of refs) {
-    if (ref.localSha === GIT_NULL_OID) continue; // branch deletion — nothing to scan
+    await scanRef(ref, emptyTree, deps);
+  }
+}
 
-    const files = await getFilesForRef(ref, emptyTree);
-    if (files.length === 0) continue;
+async function scanRef(ref: PushRef, emptyTree: string, deps: HookDependencies): Promise<void> {
+  if (ref.localSha === GIT_NULL_OID) return; // branch deletion — nothing to scan
```
Krosovok, line 49: Sophio mentioned at some point that we should try to follow the convention with always adding curly braces to blocks, even when they are one-liners. But that codestyle is not set anywhere.

kirill-knize-sonarsource: I think it's a matter of configuring a linter. Shall be done separately.

src/cli/commands/hook/hook-dependencies.ts
kirill-knize-sonarsource: Makes sense.

Krosovok (approved):

PR #281, opened 2026-05-18
CLI-338 Replace verify with analyze

kirill-knize-sonarsource:


voz:
Depending when it's merged, keep documentation updates in mind.

src/cli/commands/analyze/analyze-all.ts
```diff
@@ -0,0 +1,172 @@
...
+    return;
+  }
+
+  const agenticReport = await buildSqaaJsonReport(options, auth, command);
+  printCombinedReport(secretsReport([]), agenticReport);
+
+  if (agenticReport) {
+    applyExitCode(agenticReport.summary.totalIssues, agenticReport.summary.totalFailures);
+  }
+}
```
Krosovok, line 150: There are a lot of branches that end in a call of `printCombinedReport` and `applyExitCode`. And since the `printCombinedReport` function accepts nulls anyway, I think it's better to refactor this function to call other functions and have a clear exit point at the end:

1. run secrets (or maybe not because the library is absent, then just return null)
2. run agentic (again, it can return null for some cases already)
3. use the results (or their absence) in `printCombinedReport` and `applyExitCode`

src/ui/messages.ts
```diff
@@ -24,6 +24,27 @@ import { cyan, green, isTTY, red, yellow } from './colors.js';
 import { isMockActive, recordCall } from './mock.js';
 import type { ColorFn } from './types.js';
 
+let _formattedOutputMode = false;
+const _collectedMessages: string[] = [];
```
Krosovok, line 28: I think it's ok for now, but if want to scale the different output formats approach, we would need to come up with something else.

src/index.ts
```diff
@@ -24,7 +24,23 @@
...
+if (
+  process.argv.some(
+    (a, i) => (a === '--format' && process.argv[i + 1] === 'json') || a === '--format=json',
+  )
+) {
+  setFormattedOutputMode(true);
+}
+
+const state = loadState();
+initSentry(state);
```
Krosovok: It seems that sentry initialization was deliberately moved to preAction to command-tree.ts. I think you can remove it from here.

Krosovok (approved):

PR #282, opened 2026-05-18
CLI-467 Update Sentry version, move Sentry initialization to after argument parsing, reduce explicit flush timeout to 500ms

tomshafir-sonarsource:


nquinquenel (approved):
Looks good, nice!

src/cli/command-tree.ts
```diff
@@ -389,6 +391,13 @@ if (process.env[TELEMETRY_FLUSH_MODE_ENV]) {
   COMMAND_TREE.command('flush-telemetry', { hidden: true }).anonymousAction(flushTelemetry);
 }
 
+// Defer Sentry initialization until a command action is about to run, so that
+// non-execution paths like --help, --version, and unknown commands don't pay
+// for it. initSentry() is idempotent, so it's safe across nested commands.
+COMMAND_TREE.hook('preAction', () => {
```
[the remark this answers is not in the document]
tomshafir-sonarsource: Not worth it. Moving initSentry into preAction was specifically to skip the cost on --help, --version, and unknown commands — none of which fire preAction. Hoisting loadState() to module scope reintroduces that cost on every invocation.

src/lib/sentry.ts
```diff
@@ -27,11 +27,13 @@ import { getOrCreateUserId } from '../telemetry/user.js';
...
+
 /**
  * Initialize Sentry if telemetry is enabled.
- * Must be called before any other code that may throw.
  */
 export function initSentry(state: CliState): void {
+  if (initialized) return;
   if (!state.telemetry.enabled || process.env.SONARQUBE_CLI_DISABLE_SENTRY) return;
 
   const environment = process.env.SONARSOURCE_DOGFOODING === '1' ? 'dogfood' : 'production';
```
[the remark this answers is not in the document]
tomshafir-sonarsource: I don't think this is a real bug. Once initialized is true, the guard short-circuits before the telemetry check regardless of where the assignment lives — so a mid-session telemetry toggle wouldn't be caught either way.

src/cli/command-tree.ts
```diff
@@ -389,6 +391,13 @@ if (process.env[TELEMETRY_FLUSH_MODE_ENV]) {
...
+  initSentry(loadState());
+});
```
nquinquenel, line 402: Small comment here, even when initialized, it's still called on every command. We could have a small check around to check if it has been initialized or not

tomshafir-sonarsource: Idempotency check exists inside initSentry function

src/cli/command-tree.ts
```diff
@@ -389,6 +391,13 @@ if (process.env[TELEMETRY_FLUSH_MODE_ENV]) {
   COMMAND_TREE.command('flush-telemetry', { hidden: true }).anonymousAction(flushTelemetry);
 }
 
+// Defer Sentry initialization until a command action is about to run, so that
+// non-execution paths like --help, --version, and unknown commands don't pay
+// for it. initSentry() is idempotent, so it's safe across nested commands.
+COMMAND_TREE.hook('preAction', () => {
+  initSentry(loadState());
```
damien-urruty-sonarsource, line 401: Could we avoid loading the state several times?

damien-urruty-sonarsource, line 401: Just for my understanding, what is the rationale for not loading Sentry for help and version commands? Is it because the CLI cannot crash in this case?

src/index.ts
```diff
@@ -25,16 +25,10 @@
...
-const SENTRY_FLUSH_TIMEOUT = 2000;
-
-const state = loadState();
-initSentry(state);
 
 await postUpdate.runPostUpdateActions();
 
 await COMMAND_TREE.parseAsync(process.argv);
-await Sentry.flush(SENTRY_FLUSH_TIMEOUT);
+await Sentry.flush(SENTRY_FLUSH_TIMEOUT_MS);
```
damien-urruty-sonarsource: I suppose there is a guard to return early if Sentry is not initialized?

damien-urruty-sonarsource (approved):

src/cli/command-tree.ts
tomshafir-sonarsource: From the `Sentry SDK improvements for repositories` document:
> **move Sentry initialization to after argument parsing** so non-execution paths like --help and --version avoid unnecessary startup work

damien-urruty-sonarsource: I saw this comment in the document, yes, but I don't really understand. I get that the idea is to speed up execution for these commands, but conceptually, I don't like to remove part of the coverage. What if there is a rare bug with Commander? What if argument parsing fails where it shouldn't? We have some custom code for argument parsing. I feel we are restraining the scope of the collection for no good reason, or at least very marginal gains

src/index.ts
tomshafir-sonarsource: Outdated comment: reverted deferral of Sentry initialization

src/cli/command-tree.ts
nquinquenel: I think I agree with Sentry, for multiple reasons:
- Static commands should resolve as fast as possible - `--help` for instance is the basic command that will be used by every agent to figure out what to do, and will do so very frequently
- IMO, informational flags should be purely isolated and not trigger network communications
- GH CLI also does not trigger telemetry on `--help` or `--version`=

damien-urruty-sonarsource: I quickly did some measurements, and initializing Sentry takes between 6 and 10ms on my machine. As I said I don't think it's a huge gain to remove it. 

> informational flags should be purely isolated and not trigger network communications

They will do so only if an error occurs?

Anyway, I was just sharing my personal opinion, feel free to ignore it

nquinquenel: What is the behavior if the CLI is used in an environment without any internet connection, would it impact the commands (e.g. does Sentry timeout or retries multiple times)?

I think either way are ok but IMO it's more industry-standard not to have telemetry on such static commands.

nquinquenel:
@sonar-review-alpha review

src/cli/command-tree.ts
damien-urruty-sonarsource: From what I gathered there is no retry, it's a fire-and-forget approach

nquinquenel (approved):

PR #285, opened 2026-05-18
CLI-473 Automate GitHub release notes generation using Claude

nquinquenel:


nquinquenel:
@sonar-review-alpha review

voz:
I see it's a draft, yet do you think we can have it review readme (and other docs in this repo) as well based on the changes, given we need to keep them up to date as we evolve code?

nquinquenel:
> I see it's a draft, yet do you think we can have it review readme (and other docs in this repo) as well based on the changes, given we need to keep them up to date as we evolve code?

Everything is possible. This PR was specifically targeted towards release notes automation. I have another larger change coming that is handling our documentation website updates - this could fit in this one, where we could also check for updates regarding our own doc.

build-scripts/generate-release-notes.ts
```diff
@@ -0,0 +1,502 @@
...
+ *   bun build-scripts/generate-release-notes.ts --tag 0.12.0.1512 --dry-run
+ *
+ * In CI (RELEASED_VERSION + CLAUDE_CODE_API_KEY set):
+ *
+ *   bun build-scripts/generate-release-notes.ts --out release-notes.md
+ *
+ * The script:
+ *   1. Resolves the previous release tag with `git describe`.
+ *   2. Collects commit subjects between the previous tag and the released one.
+ *   3. Pulls the 2 most recent published GitHub releases as style examples (via `gh`).
```
tomshafir-sonarsource: If I'm not mistaken step 3 was replaced by the `RELEASE_NOTES_FORMAT_EXAMPLE` constant.

tomshafir-sonarsource (approved):
LGTM! Left one small comment.

build-scripts/generate-release-notes.ts
nquinquenel: Ah correct! This is a stale comment. 

For the explanation: my initial implementation was to fetch the last 2 releases as example for the LLM. But I think this was highly dependent on the last 2 releases, and required few more steps to fetch this. I ended up hardcoding 2 releases example that I find relevant.

PR #290, opened 2026-05-19
CLI-476 CLI output reports script

Krosovok:
A backbone file for generating a report with the CLI outputs and a Claude skill to adjust and run it for a UI report.

It's pretty basic but still can be a useful tool for us.


build-scripts/ux-report/generate-ux-report.ts
```diff
@@ -0,0 +1,898 @@
+/*
```
damien-urruty-sonarsource, line 1: What is in `build-scripts/docs` so far is for cli.sonarqube.com. We could put more things there, but I'm not sure this script is related to "docs". I suggest moving it to another folder, maybe `ux-report`?

.claude/commands/generate-ux-report.md
```diff
@@ -0,0 +1,124 @@
+Generate a UX output report for the SonarQube CLI.
+
+This command produces `docs/cli-output-report.md` — a human-readable document showing the actual terminal output for every meaningful CLI scenario, organized by section. No credentials, no running servers, no setup required beyond having the project checked out.
```
damien-urruty-sonarsource: Same here, we should move it to a different folder than `docs/` I guess. Also, could we gitignore the report?

damien-urruty-sonarsource (approved):
LGTM. The only concern is that the `generate-ux-reports.ts` is tied to the current shape of the CLI. We would have to update it whenever we change commands. Not sure how to ensure it stays up-to-date. Could we imagine something more generic, e.g. generating the cases from the command tree, by taking all commands and options and trying to combine them to exercise different behaviors? Not sure how to do that while keeping some structure in the report, and also how to manage interactive commands

build-scripts/ux-report/generate-ux-report.ts
Krosovok: Yes. Good idea. I was not sure where exactly it belongs.

Krosovok:
@damien-urruty-sonarsource updating the test file is part of the Claude skill. It will look through the existing ITs and will try to get the relevant cases and update the file. It's not too reliable, but should be ok for the most part. I tried it after removing a chunk of a test file and it's mostly acceptable. I expect it to be able to pick up whatever new is added and use the existing test as an example.

PR #301, opened 2026-05-20
CLI-488 Do not enable CAG toolset when starting MCP server from CLI

kirill-knize-sonarsource:


src/lib/mcp/mcp-helper.ts
```diff
@@ -125,8 +125,15 @@ export function getMcpContainerCommand(
   }
 
   if (options.toolsets) {
-    args.push('-e', 'SONARQUBE_TOOLSETS');
-    env.SONARQUBE_TOOLSETS = options.toolsets;
+    const filtered = options.toolsets
```
sophio-japharidze-sonarsource: Looking at [refinement doc](https://xtranet-sonarsource.atlassian.net/wiki/spaces/CodeRemedi/pages/5147164713/CLI-420+Refinement+-+Support+CAG+binary+integration#1.-Decouple-CAG-from-MCP-configuration) , "Explicit opt-in must remain supported:".

By default the `--toolsets` option is not provided. So when it is not provided, we should spawn the MCP command with default toolsets - CAG

sophio-japharidze-sonarsource (changes requested):
The logic should be reversed. Keep `--toolsets` that is provided explicitly, and when nothing is provided spawn MCP docker command with default toolsets minus CAG

src/lib/mcp/mcp-helper.ts
kirill-knize-sonarsource: Addressed!

sophio-japharidze-sonarsource (approved):

PR #303, opened 2026-05-20
CAG-416 Refresh CAG skills after CLI update

mpaladin:
## Summary
- Refresh the pinned CAG binary during post-update when CAG is already registered
- Reinstall registered non-global CAG skills without rerunning `cag init`
- Skip deleted project roots and unsupported agent entries gracefully


mpaladin:
@sonar-review-alpha review

kirill-knize-sonarsource (approved):
LGTM

If you feel fancy, you can add an integration test with real binary to validate that skill refresh works correctly when there's already old skill registered in the `state.json`.

mpaladin:
> LGTM
> 
> If you feel fancy, you can add an integration test with real binary to validate that skill refresh works correctly when there's already old skill registered in the `state.json`.

@kirill-knize-sonarsource  Thanks, I was considering that. If we want to add some e2e tests we would need to rely on a fixed project on SQC, would that be something ok on your side?

PR #304, opened 2026-05-21
CAG-418 Update sonarqube cli to use CAG 0.10.0.1024

mpaladin:
## Summary

Track the CAG 0.10.0.1024 release end-to-end: adapt to its renamed `tool integrate` subcommand, pass the context env vars and the new `--sca-enabled` flag, and refresh the binary distribution to the new unified linux build.

- **Context env wiring**: pass `SONAR_CONTEXT_ORGANIZATION`, `SONAR_CONTEXT_PROJECT`, `SONAR_CONTEXT_TOKEN`, and `SONAR_CONTEXT_URL` to every CAG invocation (passthrough + integrate flow). The shared builder lives in `src/cli/commands/_common/context-augmentation-env.ts` and has dedicated unit coverage.
- **SCA flag**: query SCA availability via `SonarQubeClient.getScaEnablement()` at integrate time, forward it as `--sca-enabled=true|false` to `tool integrate`, and persist it on the recorded `SkillExtension` so post-update replays the same flag without re-querying the server. Failures and legacy records resolve to `false`.
- **CAG CLI compat**: align with SonarSource/sonar-context-augmentation#150 and #159 — adopt the new `tool <subcommand>` interface, rename `tool init` → `tool integrate`, drop removed arguments, and collapse the integrate flow to a single `tool integrate` call (no more separate `skill --install`).
- **Stop-before-refresh on self-update**: post-update now runs `tool stop --all` against the previously-installed binary (read from `state.tools.installed`) before installing the new binary and re-running `tool install-skill` for every registered project skill. Best-effort: skipped for legacy skill-only state, debug-logged on failure.
- **Binary upgrade to 0.10.0.1024**: upstream now publishes a unified static linux build, so the temporary linux→alpine routing in `buildCagPlatformSuffix` is gone and the pinned platforms in `package.json` switch `alpine` → `linux`. Signatures regenerated via `bun run fetch:signatures`; embedded keys are now `linux-x64`/`linux-arm64`.
- **Tests & docs**: integration coverage for the post-update stop+refresh path, integrate tests rewritten for the single-step `tool integrate` shape, shared `CagInvocation` helper for the JSONL stub log, `CLAUDE.md` refreshed to describe the actual subcommands the CLI runs today.


mpaladin:
@sonar-review-alpha review

src/cli/commands/integrate/_common/context-augmentation.ts
```diff
@@ -216,22 +219,49 @@ export async function installContextAugmentationSkill({
...
+/**
+ * Best-effort `sonar-context-augmentation tool stop --all` invocation.
+ * Used during post-update to stop running CAG tools before refreshing skills
+ * so the refreshed skill templates take effect on next start. Failures are
+ * logged at debug level and never surfaced to the user.
+ */
+export async function stopAllContextAugmentationTools(binaryPath: string): Promise<boolean> {
+  const result = await runCagSubprocess(binaryPath, ['tool', 'stop', '--all'], {
+    projectRoot: process.cwd(),
+  });
```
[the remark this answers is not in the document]
mpaladin: it's ok to use process.cwd(), it is global

mpaladin:
@sonar-review-alpha review

mpaladin:
@sonar-review-alpha review

mpaladin:
@sonar-review-alpha review

mpaladin:
@sonar-review-alpha review

nquinquenel (approved):
Looks good, I also validated on my machine

src/cli/commands/context/index.ts
```diff
@@ -61,7 +162,15 @@ export async function runContextPassthrough(
...
     }
-    env = { ...process.env, SONAR_TOKEN: auth.token };
+    const recordedConfig = resolveRecordedContextAugmentationConfig(process.cwd());
+    const serverUrl = recordedConfig.serverUrl ?? auth.serverUrl;
+    const organization = recordedConfig.organization ?? auth.orgKey;
+    env = buildContextAugmentationEnv({
+      organization,
+      projectKey: recordedConfig.projectKey,
+      serverUrl,
+      token: await resolveContextToken(auth, serverUrl, organization),
```
michael-jabbour-sonarsource, line 172: A question for me to understand: Why does this depend on `serverUrl` and `organization` now?

michael-jabbour-sonarsource (approved):
LGTM. I did some tests on Linux and my Windows VM. Everything seems to behave as expected.

I couldn't find a way to test `self-update` locally, but the test looks good to me...

src/cli/commands/context/index.ts
mpaladin: @michael-jabbour-sonarsource it is to ensure that we get auth.token only when the url and org are the same as the one during integration.
Maybe we could skip https://github.com/SonarSource/sonarqube-cli/pull/304/changes#diff-941a2f131b1ec01b8a05ffa753ce60c44ad066f8d656aa33f635e3e0acd8f20aR133-R136 in favor of always bailing out.

mpaladin: @michael-jabbour-sonarsource I started creating some real e2e tests: https://github.com/SonarSource/sonarqube-cli/pull/317

michael-jabbour-sonarsource: I was mostly trying to remember if there was a recent change that requires this on sonar-context-augmentation. Thanks for the clarification...

I think it is an improvement, but I also couldn't find a way to test manually because I can't enable cag in my own org on SQC at the moment:

image

I am fine with the current version.

PR #322, opened 2026-05-27
CLI-532 Record subcommand telemetry for sonar context passthrough

mpaladin:
## Summary
- `sonar context <action> [args...]` is a Commander passthrough that owns a single `Command` node, so `storeEvent`'s parent-chain walk always recorded `subcommand=null` and the forwarded CAG action was lost from telemetry.
- Derive a subcommand string from the leading positional tokens (stopping at the first flag, so option values are never captured), map `--help`/`-h` to `"help"`, and publish it via a small `WeakMap<Command, string | null>`-backed API the telemetry module reads in `storeEvent`.

### Resulting telemetry shape

| Invocation | command | subcommand |
| --- | --- | --- |
| `sonar context` | `context` | `null` |
| `sonar context --help` / `-h` | `context` | `help` |
| `sonar context status` | `context` | `status` |
| `sonar context guidelines get` | `context` | `guidelines get` |
| `sonar context tool stop --all` | `context` | `tool stop` |
| `sonar context get-source --file foo.ts --line 4` | `context` | `get-source` |

Option flags and their values are intentionally never captured.


src/telemetry/index.ts
```diff
@@ -52,7 +61,9 @@ export function storeEvent(command: Command, success: boolean): Promise<void> {
     current = current.parent;
   }
   const topCommand = commandNames[0];
-  const subcommand = commandNames.length > 1 ? commandNames.slice(1).join(' ') : null;
+  const commandPathTail = commandNames.slice(1);
+  const fallbackSubcommand = commandPathTail.length > 0 ? commandPathTail.join(' ') : null;
+  const subcommand = passthroughSubcommands.get(command) ?? fallbackSubcommand;
```
kirill-knize-sonarsource: `makeCommand('context')` produces a single-level chain, so              
  `fallbackSubcommand` is already `null` — the assertion passes because
  `null ?? null === null`, not because the stash wins.                    
                                                                          
  Use `makeCommand('context child')` to get a non-null fallback; that version would expose the real bug: `??` treats the stashed `null` as nullish, so `null ?? 'child'` returns `'child'` instead.

  Fix:                                                                          
      const subcommand = passthroughSubcommands.has(command)
        ? passthroughSubcommands.get(command) ?? null
        : fallbackSubcommand;

kirill-knize-sonarsource (approved):
One comment on test, else LGTM

kirill-knize-sonarsource:
I've left a comment in the [CLI-532](https://sonarsource.atlassian.net/browse/CLI-532) to document the discussion.


[CLI-532]: https://sonarsource.atlassian.net/browse/CLI-532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/telemetry/index.ts
mpaladin: Good catch — fixed in b223571. Switched to a `has()` check so a stashed `null` always wins over the fallback, and updated the regression test to use `makeCommand('context child')` so it actually exercises the case (the previous `makeCommand('context')` had a null fallback, so the test passed for the wrong reason).

PR #323, opened 2026-05-27
CLI-519 Remove --with-token

Krosovok:
----
## Summary by Gitar

- **CLI Authentication:**
  - Removed `--with-token` flag from `auth login` command.
  - Updated authentication flow to support token-based login via environment variables (`SONARQUBE_CLI_TOKEN`, `SONARQUBE_CLI_SERVER`, `SONARQUBE_CLI_ORG`) for scripted/CI environments.
- **Testing & Documentation:**
  - Updated all integration and E2E tests to reflect the removal of the `--with-token` CLI argument.
  - Revised `README.md` and documentation examples to reflect the new environment-variable-based authentication flow.


README.md
```diff
@@ -401,18 +431,25 @@ sonar list issues --project my-org_my-project
...
+3. Either run interactive login:
+   ```bash
+   sonar auth login
+   ```
+   Or use environment variables directly (for scripted/CI environments):
    ```bash
-   sonar auth login --with-token YOUR_TOKEN
+   export SONARQUBE_CLI_TOKEN=YOUR_TOKEN
+   export SONARQUBE_CLI_SERVER=https://sonarcloud.io  # or your SonarQube Server URL
+   export SONARQUBE_CLI_ORG=your-org-key              # SonarQube Cloud only
```
damien-urruty-sonarsource: We had a related discussion recently with Andrii. It was on a bot's comment and got removed, and I think it applies here. I'm afraid that if we start documenting such commands, people might use them in the terminal literally just before running the command, and it will also end up the command in history. We should make it clear that these commands should be used in a .bashrc or .zshrc or similar, not directly in the terminal. Or just mention with a sentence that those env vars should be set but without refering to export

damien-urruty-sonarsource (approved):
Everything looks good, I just have one concern

README.md
Krosovok: Good point. I think I will rephrase it to just list the variables, not making it look like a command.

PR #326, opened 2026-05-27
CLI-443 Change model to distinguish features and dependencies

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Model refactoring:**
  - Introduced `dependencies` registry in `CliState` to decouple shared binaries from individual integration features.
  - Updated `InstalledIntegrationFeature` to reference shared dependencies by ID rather than embedding resources.
- **Migration:**
  - Added `migrateLegacyIntegrationDependencies` to automatically move existing binary resources into the new shared dependency registry.
- **Framework updates:**
  - Enhanced `IntegrationInstaller` to support `dependencyNeedsApply`, `onDependencyInstalled` callbacks, and cross-feature dependency tracking.
  - Updated `IntegrationRegistry` to validate unique dependency IDs during registration.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,7 +38,7 @@ To add a new command: add it to `src/cli/command-tree.ts` and implement the logi
 Please declare commands using the type defined in `src/cli/commands/_common/sonar-command.ts`.
 By default, new commands should register a `authenticatedAction()`, only technical commands will use `anonymousAction()`.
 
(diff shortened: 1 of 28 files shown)


tomshafir-sonarsource (approved):
Looks good! Found one issue worth noting:

`collectReferencedDependencyIds` (in `installer.ts`) calls `feature.dependencies.map(...)` on every installed feature. `migrateState` doesn't backfill that field on entries written by previous CLI versions, so it's undefined → throws.

Repro: run any integrate command against legacy state:

```
⚠️ Failed to update configuration state:
   undefined is not an object (evaluating 'feature.dependencies.map')
```

throw is caught before saveState, so files land on disk but state never updates. Every future integrate re-runs the install → no self-heal.

damien-urruty-sonarsource:
@tomshafir-sonarsource goog catch! The framework was released in the latest version but no integrations depended on it, only the preparatory work was released. This means the bug you shared would only impact internal users that ran integrate since last week. I don't want to add extra logic just for that. I will proceed with the merge

PR #334, opened 2026-05-28
CLI-526 Add a confirmation for the server selection

Krosovok:
----
## Summary by Gitar

- **Security enhancement:**
  - Added `confirmServerTrust` to prompt users to confirm server connection before authenticating.
  - Added security warning when connecting to non-SonarQube Cloud servers.
- **Logic updates:**
  - Integrated trust confirmation into both `--server` flag and interactive server selection flows.
  - Updated integration tests to simulate user input for the new trust confirmation prompt.


sophio-japharidze-sonarsource (approved):
We might need to test how this will look inside agent plugins that use `integrate` skill and are trying to set up connections themselves.

Krosovok:
> We might need to test how this will look inside agent plugins that use `integrate` skill and are trying to set up connections themselves.

Checked. The agent pre-filled the parameters and let the human run the command anyway.

PR #335, opened 2026-05-28
CLI-447 Use the declarative framework for migrations

damien-urruty-sonarsource:
----
## Summary by Gitar

- **New declarative framework:**
  - Introduced `IntegrationRegistry` core logic to centrally manage and validate integration declarations.
  - Added `ALL_INTEGRATIONS` entry point for automated registration of existing integrations.
- **Migration logic:**
  - Added `migrateDeclarativeIntegrations` to `post-update.ts` to automatically reapply features and prune stale states after CLI upgrades.
  - Integrated `migrateClaudeCodeHooks` with the new framework to skip legacy migration when declarative tracking is present.
- **Testing enhancements:**
  - Added comprehensive unit tests for registry validation and integration seeding.
  - Added integration tests to verify that declarative `Claude` hooks are refreshed correctly during CLI upgrades.

### DIFF
--- a/src/cli/commands/integrate/claude/declaration.ts
+++ b/src/cli/commands/integrate/claude/declaration.ts
@@ -22,19 +22,19 @@ import { join } from 'node:path';
 
 import { CLI_COMMAND } from '../../../../lib/config-constants';
 import { getMcpConfig, getMcpConfigFilePath } from '../../../../lib/mcp/mcp-helper';
+import { OBSOLETE_A3S_MARKER, removeObsoleteHookArtifacts } from '../../../../lib/migration';
 import { createSonarSecretsHooksFeature } from '../_common/features/sonar-secrets-hooks-feature';
 import {
   createAgentHookEntry,
   resolveAgentHookScriptPath,
   upsertAgentHooks,
 } from '../_common/hooks';
-import {
-  type IntegrationContext,
-  type IntegrationDeclaration,
-  jsonPatch,
-  supportedIntegrations,
-  wholeFile,
-} from '../_common/registry';
+import { jsonPatch, wholeFile } from '../_common/registry/resources';
+import type {
+  FeatureOperation,
+  IntegrationContext,
+  IntegrationDeclaration,
+} from '../_common/registry/types';
 import type { IntegrateAgentOptions } from '../_common/types';
 import {
   getSecretPreToolTemplateUnix,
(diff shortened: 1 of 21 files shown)


src/cli/commands/integrate/claude/declaration.ts
```diff
@@ -63,46 +63,49 @@ export const claudeIntegration: IntegrationDeclaration<ClaudeIntegrationOptions>
...
+          },
+          {
+            eventType: 'UserPromptSubmit',
+            matcher: '*',
+            marker: 'sonar-secrets',
+            scriptPath: PROMPT_SCRIPT_REL,
+          },
+        ],
+      }),
+      operations: [createRemoveObsoleteA3sArtifactsOperation()],
```
sophio-japharidze-sonarsource: Can we drop this logic completely? It has already been a few iterations since the A3S rename, and I think the less migration logic we have in v1.0 the better. I think the need for keeping this is minimal.

src/lib/post-update.ts
```diff
@@ -306,6 +370,11 @@ function isExistingDirectory(path: string): boolean {
 export async function migrateClaudeCodeHooks(homedirFn: () => string = homedir): Promise<void> {
   const state = loadState();
 
+  if (hasInstalledDeclarativeIntegration(state, CLAUDE_INTEGRATION_ID)) {
+    logger.debug('Declarative Claude integration detected — skipping legacy hook migration');
```
sophio-japharidze-sonarsource:
```suggestion
    logger.debug('Declarative Claude Code integration detected — skipping legacy hook migration');
```

tests/unit/cli/commands/self-update/post-update.test.ts
```diff
@@ -227,6 +236,130 @@ describe('runPostUpdateActions', () => {
...
+          dependencies: [],
+          resources: [],
+          operations: [],
+        },
+      ],
+    });
+    loadStateSpy.mockReturnValue(state);
+
+    const registry = new IntegrationRegistry();
+    registry.register({
```
sophio-japharidze-sonarsource, line 311: Can we also add `new-feature` here and assert that this new feature does not get applied by the migration without user consent?

sophio-japharidze-sonarsource (approved):
LGTM! Added a few questions/comments

src/cli/commands/integrate/claude/declaration.ts
damien-urruty-sonarsource: I agree, yes. I will remove it

tests/unit/cli/commands/self-update/post-update.test.ts
damien-urruty-sonarsource: Done

PR #340, opened 2026-05-29
CLI-520 Improve phrasing

Krosovok:
----
## Summary by Gitar

- **CLI improvements:**
  - Expanded `auth login` description to clarify authentication methods for CI and automation.
  - Updated `list` command description and added usage context to `analyze agentic` command.
- **Refactoring:**
  - Simplified `auth status` output logic by centralizing `connectionLines` usage and removing redundant formatting steps.
  - Standardized error messaging across `dependency-risks.ts` and updated corresponding test expectations in `analyze-dependency-risks.test.ts`.
  - Moved `integrate git` definition in `command-tree.ts` to improve code structure.

### DIFF
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -103,7 +103,11 @@ const auth = COMMAND_TREE.command('auth').description(
 
 auth
   .command('login')
-  .description('Save authentication token to keychain')
+  .description(
+    'Authenticate via browser and save credentials in the system keychain. ' +
+      'Must be run manually — agents cannot authenticate themselves. ' +
+      'For CI and automation, use environment variables instead: https://docs.sonarsource.com/sonarqube-cli/using-sonarqube-cli/environment-variables',
+  )
   .option(
     '-s, --server <server>',
     'SonarQube Server URL, SonarQube Cloud EU (https://sonarcloud.io), or SonarQube Cloud US (https://sonarqube.us). Defaults to SonarQube Cloud EU.',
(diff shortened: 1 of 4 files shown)


src/cli/commands/auth/status.ts
```diff
@@ -102,11 +95,7 @@ export async function authStatus(): Promise<void> {
   );
   blank();
 
-  if (status === 'valid') {
-    printConnected(conn.serverUrl, 'OS Keychain', conn.orgKey);
-  } else {
-    displayTokenStatus(conn.serverUrl, conn.orgKey, status);
-  }
+  displayTokenStatus(conn.serverUrl, conn.orgKey, status, 'OS Keychain');
```
damien-urruty-sonarsource: Is there another source than 'OS Keychain'? It seems it's the only argument we pass, so maybe we could inline it?

src/cli/command-tree.ts
```diff
@@ -156,6 +155,23 @@ const integrateCommand = COMMAND_TREE.command('integrate').description(
   'Setup SonarQube integration for AI coding agents, git and others.',
 );
 
+integrateCommand
```
damien-urruty-sonarsource, line 158: If I'm not mistaken, this will influence the order in which the commands appear on cli.sonarqube.com. But it's probably fine

src/cli/commands/auth/status.ts
```diff
@@ -32,28 +32,21 @@ function connectionLines(serverUrl: string, orgKey: string | undefined): string[
...
   if (status === 'valid') {
-    note(lines, '✓ Connected', NOTE_STYLES.success);
+    printConnected(serverUrl, source, orgKey);
   } else if (status === 'invalid') {
-    note(
-      [...lines, '', 'Run "sonar auth login" to reauthenticate'],
-      '✗ Token invalid',
-      NOTE_STYLES.error,
-    );
+    note(lines, '✗ Token invalid', NOTE_STYLES.error);
```
damien-urruty-sonarsource, line 48: Not directly related to your changes, but I find these function calls a bit verbose. We could maybe directly have `error(lines, '✗ Token invalid')`? Another thing that bothers me is the implementation of note that contains logic to mock the behavior. Maybe what we should do instead is introduce an `output` object that would expose `success` or `error` methods (or maybe via a first `note` field). And we could more easily provide a mock implementation for tests. I created https://sonarsource.atlassian.net/browse/CLI-546 to track that separately

damien-urruty-sonarsource (approved):
LGTM, some small comments

src/cli/command-tree.ts
Krosovok: The idea was to bring git up and keep the agents visually together. So it's for the better if it would be the same on docs site.

src/cli/commands/auth/status.ts
Krosovok: Thanks! It's a good idea.

PR #344, opened 2026-05-31
CLI-484 Fix release notes automation

nquinquenel:
gh release needs a repo context; without actions/checkout, it tries to use the local git remote and fails with fatal: not a git repository.

----
## Summary by Gitar

- **CI/CD Configuration:**
  - Added `GH_REPO` environment variable to `full-release.yml` for release verification and publication steps.

### DIFF
--- a/.github/workflows/full-release.yml
+++ b/.github/workflows/full-release.yml
@@ -32,6 +32,7 @@ jobs:
       - name: Verify draft release exists for expected version
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
           VERSION: ${{ steps.version.outputs.release-version }}
         run: |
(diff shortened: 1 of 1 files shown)


vnaskos-sonar (approved):
🚀

.github/workflows/full-release.yml
```diff
@@ -132,6 +133,7 @@ jobs:
       - name: Copy draft body to published release and delete draft
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
```
vnaskos-sonar, line 136: Note: Github also provides, `GITHUB_REPOSITORY` env variable available to the whole pipeline, which should contain the same value as github.repository

nquinquenel: We set `GH_REPO` because `gh` resolves the target repo from `GH_REPO`

PR #345, opened 2026-05-31
CLI-416, CLI-512 Ask features one by one

tomshafir-sonarsource:
----
## Summary by Gitar

- **Interactive Integration Flow:**
  - Refactored integration framework to enable interactive, per-feature selection, replacing rigid `--hook` or `--install` flags with a declarative `when` condition.
  - Added `--non-interactive` flag to all `integrate` commands to bypass prompts.
- **Feature Selection & Logic:**
  - Replaced hardcoded integration flags with a flexible `when` condition supporting `ask`, `skip`, or `install` logic.
  - Implemented `isFeatureInstalled` checks to prevent redundant installations and manage global-to-local transitions.
- **Instruction Templates:**
  - Updated `textSnippet` to use marker-delimited blocks, allowing multiple features (e.g., secrets scanning and SQAA) to coexist in instructions files.
- **CLI Command Cleanup:**
  - Standardized `integrate git` to prompt for hook selection interactively, removing the redundant `--hook` option.
  - Migrated Git, Copilot, and Codex integrations to the new declarative installation registry.

### DIFF
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -165,6 +165,10 @@ integrateCommand
   .description(
     'Install a Git pre-commit hook that scans staged files for secrets before each commit, or a Git pre-push hook that scans committed files for secrets before each push.',
   )
+  .option(
+    '--hook <type>',
+    'Hook to install: pre-commit (scan staged files) or pre-push (scan files in unpushed commits)',
+  )
   .option('--force', 'Overwrite existing hook if it is not from sonar integrate git')
   .option('--non-interactive', 'Non-interactive mode (no prompts)')
   .option(
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -207,6 +207,7 @@ integrateCommand
   )
   .option('-p, --project <project>', 'Project key. Mutually exclusive with --global.')
   .option('--skip-context', 'Skip the sonar-context-augmentation install/init/skill step')
+  .option('--non-interactive', 'Non-interactive mode (no prompts)')
   .addHelpText('after', projectKeyExtraHelp)
   .authenticatedAction((auth, options: IntegrateAgentOptions) => integrateCopilot(auth, options));
 
@@ -236,6 +237,7 @@ integrateCommand
   )
   .option('-p, --project <project>', 'Project key. Mutually exclusive with --global.')
   .option('--skip-context', 'Skip the sonar-context-augmentation install/init/skill step')
+  .option('--non-interactive', 'Non-interactive mode (no prompts)')
   .addHelpText('after', projectKeyExtraHelp)
   .authenticatedAction((auth, options: IntegrateAgentOptions) => integrateCodex(options, auth));
 
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -165,10 +165,6 @@ integrateCommand
   .description(
     'Install a Git pre-commit hook that scans staged files for secrets before each commit, or a Git pre-push hook that scans committed files for secrets before each push.',
   )
-  .option(
-    '--hook <type>',
-    'Hook to install: pre-commit (scan staged files) or pre-push (scan files in unpushed commits)',
-  )
   .option('--force', 'Overwrite existing hook if it is not from sonar integrate git')
   .option('--non-interactive', 'Non-interactive mode (no prompts)')
   .option(
--- a/src/cli/commands/integrate/_common/features/sonar-secrets-hooks-feature.ts
+++ b/src/cli/commands/integrate/_common/features/sonar-secrets-hooks-feature.ts
@@ -22,12 +22,13 @@ import { join } from 'node:path';
 
 import { createAgentHookEntry, resolveAgentHookScriptPath, upsertAgentHooks } from '../hooks';
 import { sonarSecretsBinaryDependency } from '../registry/dependencies';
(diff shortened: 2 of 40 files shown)


nquinquenel:
@tomshafir-sonarsource The PR is quite big, do you think we could have splitted it in smaller pieces? I feel like it's implementing the framework, and then applying it 4 times for codex/claude/git/copilot. Also Copilot is not part of any specific commit.

Lastly, we try to keep the ticket reference in the commits. If you plan on squashing everything it's ok, otherwise I'd suggest double checking commits name.

tomshafir-sonarsource:
> @tomshafir-sonarsource The PR is quite big, do you think we could have splitted it in smaller pieces? I feel like it's implementing the framework, and then applying it 4 times for codex/claude/git/copilot. Also Copilot is not part of any specific commit.
> 
> Lastly, we try to keep the ticket reference in the commits. If you plan on squashing everything it's ok, otherwise I'd suggest double checking commits name.

You're right, I didn't assume it would amount to such a big PR when I started working on this. I restructured the tasks of the epic after seeing the scale of this PR.

However, even after restructuring the epic, all the commits of this PR belong to a single JIRA task, and I'm not sure if I could have broken it down more. This is because the initial change of this PR was the framework implementation, and immediately after it, the existing integrations broke, so I had to update them to use the new implementation. After I was done, I needed to also update the tests of everything I've touched. All of these together made the PR be the size that it is.

src/cli/commands/integrate/_common/features/sonar-secrets-hooks-feature.ts
```diff
@@ -66,16 +67,22 @@ export function resolveAgentHooksConfigPath(
...
   const resolveHooksConfigPath = (context: IntegrationContext) =>
     resolveAgentHooksConfigPath(context, config.configDir, config.hooksConfigFileName);
 
   return {
-    id: 'sonar-secrets-hooks',
-    displayName: 'secrets hooks',
-    when: ({ options }) => options.installSecretsHooks === true,
+    id: SONAR_SECRETS_HOOKS_FEATURE_ID,
+    displayName: 'secret scanning hooks',
+    hint: SONAR_SECRETS_HOOKS_HINT,
```
damien-urruty-sonarsource, line 74: `hint` is maybe a tag vague ? Is it the `description` of the feature?

src/cli/commands/integrate/_common/features/sonar-secrets-hooks-feature.ts
```diff
@@ -66,16 +67,22 @@ export function resolveAgentHooksConfigPath(
...
+    // Skip project-level install when a global secrets hook already covers it; otherwise ask.
+    when: ({ scope, state }) =>
+      scope === 'project' &&
+      isFeatureInstalled(state, config.integrationId, SONAR_SECRETS_HOOKS_FEATURE_ID, 'global')
+        ? { kind: 'skip', reason: 'global hook already covers this' }
+        : { kind: 'ask' },
```
damien-urruty-sonarsource, line 80: 1/ Here you are re-purposing the `when` field, and I think this name does not reflect the meaning of the callback anymore. What about `shouldInstall`? I'm not good at naming things...

2/ Another question I have, is how will you manage post-update migrations with this approach? Are you just ignoring the result?

3/ I think the type is a bit verbose, maybe we could come up with a more friendly syntax, just an idea:

> isFeatureInstalled(state, config.integrationId, SONAR_SECRETS_HOOKS_FEATURE_ID, 'global')
        ? skip('global hook already covers this')
        : askUser()

4/ I think we are very rarely relying on objects across the CLI, in some cases it makes sense, in some other I think the code would be clearer, e.g. `state.hasFeature(config.integrationId, SONAR_SECRETS_HOOKS_FEATURE_ID, 'global')

src/cli/command-tree.ts
```diff
@@ -180,10 +180,6 @@ integrateCommand
   .description(
     'Install a git hook that scans staged files for secrets before each commit (pre-commit) or scans committed files for secrets before each push (pre-push).',
   )
-  .option(
-    '--hook <type>',
-    'Hook to install: pre-commit (scan staged files) or pre-push (scan files in unpushed commits)',
-  )
```
damien-urruty-sonarsource: I am surprised that we are removing this option. I think my assumption when I refined this epic is that we want to provide an interactive mode for humans, but we would still allow power users or agents to specify what they want via command-line. This does not exist for other integrate commands, so we would have to introduce it, but for git I thought we would keep the option. Is it not possible now with the new interactive way?

src/cli/commands/integrate/git/index.ts
```diff
@@ -200,18 +158,14 @@ async function integrateGitGlobal(options: IntegrateGitOptions): Promise<void> {
...
-  text(`Hook: ${hook}`);
-  blank();
-
-  await installGitFeatures({ ...options, hook }, GLOBAL_HOOKS_DIR, 'global');
-  showPostInstallInfo(hook);
-  showVerificationGuide(hook);
+  const hooks = await installGitFeatures(options, GLOBAL_HOOKS_DIR, 'global');
+  for (const hook of hooks) {
+    showPostInstallInfo(hook);
+    showVerificationGuide(hook);
```
damien-urruty-sonarsource, line 174: Could we move this inside the installer? I expect printing a summary or a verification guide for features could be useful for several integrations

PR #343, opened 2026-05-29
CLI-543 Update sonarqube cli to use CAG 0.11.0.1346

mpaladin:
## Summary
- Bump `sonar-context-augmentation` from `0.10.0.1024` to `0.11.0.1346` in `package.json` and `src/lib/signatures.ts`.
- Refresh PGP signatures via `bun run fetch:signatures`.

Note: it requires `sonar context tool stop --all` on a local machine with dogfood.
`sonar context tool stop --all` is ran as part of self-update.

PR #346, opened 2026-06-01
CLI-543 Speak Cloud in fake SonarQube server to unhang the CAG integration test

mpaladin:
## Summary

- Make the fake SonarQube server impersonate SonarCloud (only when `withCagEntitlement` is configured) so the real `sonar-context-augmentation` binary completes `tool integrate` in ~2 s instead of hanging until the 30 s CLI-runner timeout.
- `baseUrl()` now returns `http://localhost:...` (server still binds to `127.0.0.1`) so CAG's `host → api.<host>` Cloud-API transformation lands on `api.localhost`, which resolves to 127.0.0.1 per RFC 6761 and reaches the same listener. `api.127.0.0.1` was a literal hostname, never resolvable.
- Add the two endpoints CAG calls during daemon startup: `/api/project_branches/list` (project existence) and the no-param branch of `/organizations/organizations` (open-beta org list).

## Why

`master` is red after `CLI-543 Update sonarqube cli to use CAG 0.11.0.1346`. CAG 0.11 introduced a daemon that validates against SonarCloud APIs during `tool integrate`; with a fake server speaking only on-prem endpoints, the daemon died before binding its IPC socket and the parent CLI sat in its 120 s readiness poll until the test's 30 s outer timeout killed it. PR #343 happened to pass on a runner where the timing landed differently — turned out to be the only green run before merge.

Test-only fix; no production CLI changes.

## Test plan

- [x] `bun test tests/integration/specs/integrate/context-augmentation.test.ts` — 19/19 pass (was 18/19 on master), ~18 s total
- [x] `bun test ./tests/integration/` — same 8 pre-existing unrelated failures as `master` HEAD, no new regressions
- [x] `bun run lint` — 0 errors (16 pre-existing warnings)

### DIFF
--- a/tests/integration/harness/fake-sonarqube-server.ts
+++ b/tests/integration/harness/fake-sonarqube-server.ts
@@ -97,7 +97,11 @@ export class FakeSonarQubeServer {
   }
 
   baseUrl(): string {
-    return `http://127.0.0.1:${this.server.port}`;
+    // Use `localhost` (not `127.0.0.1`) so CAG's Cloud-API URL transformation
+    // (host → `api.<host>`) lands on `api.localhost`, which resolves to
+    // 127.0.0.1 per RFC 6761 and reaches the same server. `api.127.0.0.1`
+    // would not resolve, breaking any test that exercises Cloud-mode CAG.
+    return `http://localhost:${this.server.port}`;
   }
 
   getRecordedRequests(): RecordedRequest[] {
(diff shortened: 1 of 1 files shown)


tests/integration/harness/fake-sonarqube-server.ts
kirill-knize-sonarsource, line 1: there's a latent UUID mismatch in the no-param branch of `/organizations/organizations` added by this PR.
  
`withCagEntitlement` stores no UUID — the cag-org-config handler derives it as `${orgKey}-uuid-v4`. But when an org also has withSqaaEntitlement with an explicit UUID, the no-param response uses the SQAA UUID
instead. CAG then calls `/a3s-analysis/cag-org-config/<sqaa-uuid>`, which the handler can't resolve, and returns 404.

No current test combines both, so nothing breaks today. Reproducer:

```
const server = await new FakeSonarQubeServerBuilder()
    .withSqaaEntitlement('my-org', 'test-uuid-1234')
    .withCagEntitlement('my-org')
    .start();

const orgs = await
fetch(`${server.baseUrl()}/organizations/organizations`).then(r =>  r.json());
const uuid = orgs[0].uuidV4; // 'test-uuid-1234' (from SQAA)

const resp = await
fetch(`${server.baseUrl()}/a3s-analysis/cag-org-config/${uuid}`);
resp.status; // 404 — expected 200
```
  
Fix: store the CAG UUID consistently (either let withCagEntitlement accept an optional UUID, or always derive it as `${orgKey}-uuid-v4` in the no-param response regardless of what SQAA has).

kirill-knize-sonarsource (approved):
Small problem in test harness that may backfire in future.
Else LGTM

tests/integration/harness/fake-sonarqube-server.ts
mpaladin: @kirill-knize-sonarsource  good catch, fixed in https://github.com/SonarSource/sonarqube-cli/pull/346/changes/d78555040574a662fed90cb1b00848a61c3ea132 .

PR #353, opened 2026-06-02
CLI-556 Update CAG to 0.12.0.1451

mpaladin:
Replace `tool install-skill` with `tool print-skill` and let the
wrapper place the rendered skill at the agent-specific workspace path.
The `tool integrate` invocation no longer needs `--agent` or
`--sca-enabled` flags; those are now only passed to `print-skill`.


damien-urruty-sonarsource (approved):
Thanks!

PR #356, opened 2026-06-02
CLI-561 SCA pre-push hook logic implementation

georgii-borovinskikh-sonarsource:
Part of CLI-541

----
## Summary by Gitar

- **New features:**
  - Implemented `git-pre-push` secondary stage for SCA dependency risk analysis via `--dependency-risks` flag.
  - Added `git-pre-push-dependency-risks.ts` to coordinate scanner execution and filter results.
  - Implemented `runSecretsStage` in `git-pre-push-secrets.ts` to modularize existing secret scanning logic.
- **CLI updates:**
  - Added `-p, --project` and `--dependency-risks` options to `hook git-pre-push` command.
  - Refactored `git-pre-push.ts` to support multi-stage hook execution (Secrets and SCA) and added uncommitted change detection.
- **Integration testing:**
  - Added comprehensive test suite in `hook-git-pre-push.test.ts` covering error handling, skip logic, and failure scenarios.
- **Improvements:**
  - Added `hasUncommittedChanges` check to notify users about local changes during the push process.

### DIFF
--- a/src/cli/commands/hook/git-pre-push-dependency-risks.ts
+++ b/src/cli/commands/hook/git-pre-push-dependency-risks.ts
@@ -43,7 +43,6 @@ import {
 } from '../analyze/dependency-risk-helpers/sca-watch-patterns';
 import { formatDependencyRisksTable } from '../analyze/dependency-risk-helpers/table';
 import { buildDependencyRisksViewModel } from '../analyze/dependency-risk-helpers/view-model/build';
-import { hasUncommittedChanges } from './git-pre-push';
 
 const HOOK_STATUS_FILTER = 'new';
 
@@ -62,12 +61,6 @@ export async function runDepRisksStage(options: DepRisksStageOptions): Promise<v
 
   if (!(await shouldRunDependencyRiskAnalysis(binaryPath, options.changedFiles))) {
     return;
-  }
-
-  if (await hasUncommittedChanges()) {
-    warn(
-      'Uncommitted changes detected — they are not part of this push, but will be included in the dependency-risks scan.',
-    );
   }
 
   const filter = buildRiskFilter(HOOK_STATUS_FILTER);
@@ -97,7 +90,7 @@ export async function runDepRisksStage(options: DepRisksStageOptions): Promise<v
 
   print(formatDependencyRisksTable(viewModel));
   throw new CommandFailedError(
-    `Dependency risks detected in this push (${matchedCount} matching the configured filter).`,
+    `Dependency risks detected (${matchedCount} matching the configured filter).`,
     {
       remediationHint: `Run 'sonar analyze dependency-risks -p ${options.project}' to inspect, fix or accept the risks, then retry the push.`,
     },
--- a/src/cli/commands/hook/git-pre-push-dependency-risks.ts
+++ b/src/cli/commands/hook/git-pre-push-dependency-risks.ts
@@ -92,7 +92,7 @@ export async function runDepRisksStage(options: DepRisksStageOptions): Promise<v
   throw new CommandFailedError(
     `Dependency risks detected (${matchedCount} matching the configured filter).`,
     {
-      remediationHint: `Run 'sonar analyze dependency-risks -p ${options.project}' to inspect, fix or accept the risks, then retry the push.`,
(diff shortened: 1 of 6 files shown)


src/cli/commands/hook/git-pre-push.ts
```diff
@@ -18,50 +18,51 @@
...
+    );
   }
-}
-
-async function scanRef(ref: PushRef, emptyTree: string, deps: HookDependencies): Promise<void> {
-  if (ref.localSha === GIT_NULL_OID) return; // branch deletion — nothing to scan
 
-  const files = await getFilesForRef(ref, emptyTree);
-  if (files.length === 0) return;
+  await runSecretsStage(filesByRef, auth);
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: the same file that's in the push diff can be modified further in the working tree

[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: I have another PR that will refactor this method and resolve the list of files before running the analysis, I will make sure to move the warning to after we have the file list. I will fix the untracked file issue in this PR, working on it

src/cli/commands/hook/git-pre-push.ts
```diff
@@ -131,3 +146,12 @@ async function getFilesForNewBranch(localSha: string, emptyTree: string): Promis
     return [];
   }
 }
+
+export async function hasUncommittedChanges(): Promise<boolean> {
+  try {
+    const result = await spawnProcess('git', ['status', '--porcelain', '-uno']);
```
damien-urruty-sonarsource, line 152: I don't know if this is the best way to determine if there are uncommitted files. I also expect we already need something similar for SQAA? Might be good to use the same implementation

src/cli/commands/hook/git-pre-push.ts
```diff
@@ -18,50 +18,51 @@
...
-  if (!deps) return;
-
   const emptyTree = await getEmptyTree();
+  const filesByRef = await collectFilesForRefs(refs, emptyTree);
 
-  for (const ref of refs) {
-    await scanRef(ref, emptyTree, deps);
+  if (await hasUncommittedChanges()) {
+    warn(
+      'Uncommitted changes detected — they are not part of this push but will be included in the scan.',
```
damien-urruty-sonarsource, line 57: Would we have a way to avoid that? By taking exactly the non-pushed commits and the file that were changed in them?

src/cli/commands/hook/git-pre-push-dependency-risks.ts
```diff
@@ -0,0 +1,116 @@
...
+  ).run();
+  if (patterns.length === 0) {
+    logger.debug('Dependency-risks hook: no watch patterns returned, skipping.');
+    return false;
+  }
+
+  if (!anyFileMatches(changedFiles, patterns)) {
+    success('No dependency manifests changed in this push — skipping dependency-risks scan.');
+    return false;
+  }
```
damien-urruty-sonarsource, line 113: Just thinking now, instead of calling the SCA CLI 2 times (check if files could raise issues + analyze), couldn't we merge the 2 commands ?

damien-urruty-sonarsource (approved):
LGTM, some questions

src/cli/commands/hook/git-pre-push.ts
georgii-borovinskikh-sonarsource: This is about the content of the files. We do collect the correct file paths that will be pushed, but the analyzer read the file content directly from disk (working tree)

src/cli/commands/hook/git-pre-push-dependency-risks.ts
georgii-borovinskikh-sonarsource: We could in theory, but it depends on the sca team to add this functionality to the CLI

src/cli/commands/hook/git-pre-push.ts
georgii-borovinskikh-sonarsource: Not exactly the same.

Changed the command here to `git diff --quiet HEAD` because we're only interested in true/false - let's the user run `git status` or check the changes however else they want

PR #358, opened 2026-06-02
CLI-562 Support remove/undo in integrations framework

Krosovok:
----
## Summary by Gitar

- **Integration removal and undo:**
  - Added `removeFeature` to `IntegrationInstaller` which executes resource `remove` hooks and reverses `operations` in reverse order.
  - Added `remove` capability to `whole-file`, `text-snippet`, and `json/yaml/toml-patch` resources, enabling automated cleanup of integration artifacts.
- **New installation utilities:**
  - Introduced `installIntegration.ts` to orchestrate feature selection, context resolution, and state persistence with integrated logging and error handling.
  - Created `installation-recorder.ts` to manage lifecycle tracking and upsert logic for `features`, `resources`, and `operations` in the integration state.
- **Framework extensibility:**
  - Updated `PatchResource` and `PatchResourceOptions` to support generic document types and explicit `removePatch` callbacks for cleaner reversal logic.
  - Added an `undo` hook to `FeatureOperation` to allow reversible operations during removal workflows.
- **Testing:**
  - Added `registry-remove.test.ts` and `registry-resources.test.ts` providing extensive coverage for removal, undo, and state lifecycle management.


src/cli/commands/integrate/_common/registry/resources/whole-file.ts
```diff
@@ -77,6 +80,13 @@ export class WholeFileResource implements ResourceDeclaration {
     return existing === (await this.resolveContent(context));
   }
 
+  async remove(context: IntegrationContext): Promise<void> {
+    const path = await resolvePath(context, this.options.targetPath);
+    if (existsSync(path)) {
+      await rm(path);
+    }
+  }
```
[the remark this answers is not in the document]
Krosovok: This ticket is a precondition for full clean-up, we don't want any leftovers of full files created by us, even if they were modified by a user. For preserving, there are patch resources that imply that we only touch part of the file.

src/cli/commands/integrate/_common/registry/resources/common.ts
```diff
@@ -76,11 +77,16 @@ export async function readTextFile(path: string): Promise<string | undefined> {
   return readFile(path, 'utf-8');
 }
 
-export interface PatchResourceOptions extends BaseResourceOptions {
+export interface PatchResourceOptions<TDoc = unknown> extends BaseResourceOptions {
   targetPath: PathResolver;
+  patch: (document: TDoc, context: IntegrationContext) => MaybePromise<unknown>;
+  removePatch?: (document: TDoc, context: IntegrationContext) => MaybePromise<unknown>;
```
damien-urruty-sonarsource, line 83: Does this need to be nullable? Shouldn't we force callers to provide it?

src/cli/commands/integrate/_common/registry/resources/common.ts
```diff
@@ -103,5 +109,25 @@ export abstract class PatchResource<O extends PatchResourceOptions> implements R
     return (await readTextFile(path)) === (await this.renderContent(path, context));
   }
 
-  protected abstract renderContent(path: string, context: IntegrationContext): Promise<string>;
+  async remove(context: IntegrationContext): Promise<void> {
+    if (!this.options.removePatch) {
+      return;
+    }
```
damien-urruty-sonarsource, line 115: It would simplify this line

src/cli/commands/integrate/_common/registry/resources/text-snippet.ts
```diff
@@ -67,6 +67,20 @@ export class TextSnippet implements ResourceDeclaration {
...
 
+  async remove(context: IntegrationContext): Promise<void> {
+    const path = await resolvePath(context, this.options.targetPath);
+    const existing = await readTextFile(path);
+    if (!existing?.includes(this.startMarker)) {
+      return;
+    }
+
+    const pattern = new RegExp(
+      String.raw`(?:\r?\n)?${escapeRegExp(this.startMarker)}[\s\S]*?${escapeRegExp(this.endMarker)}(?:\r?\n)?`,
```
damien-urruty-sonarsource, line 78: I think `endMarker` is nullable, what would it do in this case?

src/cli/commands/integrate/_common/registry/install-integration.ts
```diff
@@ -0,0 +1,226 @@
...
+  registry?: IntegrationRegistry;
+  integrationId: string;
+  options: TOptions;
+  targetRoot: string;
+  scope: IntegrationScope;
+  force?: boolean;
+  attrs?: Record<string, IntegrationStateAttribute>;
+}
+
+export async function installIntegration<TOptions>({
```
damien-urruty-sonarsource, line 58: You will probably have conflicts with this class. Out of curiosity, why did you split it out of the installer?

damien-urruty-sonarsource (approved):
LGTM! Some minor comments

src/cli/commands/integrate/_common/registry/resources/text-snippet.ts
Krosovok: In this case we don't have a clear section to remove. So for now it does not find it and does not remove. I thought most of the time it would mean the file has changed and we don't want to remove everything until the end just because we didn't find the end marker.

src/cli/commands/integrate/_common/registry/resources/common.ts
Krosovok: For now the idea was to create a mechanism. The plan is to go over the implementations in https://sonarsource.atlassian.net/browse/CLI-565 and we can make it non-optional there.

src/cli/commands/integrate/_common/registry/install-integration.ts
Krosovok: Just the files getting bigger and I thought work on them is mostly finished.

src/cli/commands/integrate/_common/registry/resources/common.ts
Krosovok: We will drop it if we make it non-optional anyway though.

src/cli/commands/integrate/_common/registry/resources/text-snippet.ts
damien-urruty-sonarsource: I was a bit concerned about nullability, which is not reflected in `escapeRegExp`. If we know we don't want to support undoing if no endMarker is specified, I would make it more explicit with an early return at the top

Krosovok: Good idea. I will add an explicit exit on not having endMarker.

Krosovok: Actually, it calls the getter that just falls back to a default `# sonar:end ${this.id}` so there is no scenario where it's just not set. But still, it is not checked if it is present in the file, so I will add an additional chek.

src/cli/commands/integrate/_common/registry/installer.ts
```diff
@@ -280,6 +262,28 @@ export class IntegrationInstaller {
...
+    for (const resource of feature.resources ?? []) {
+      if (resource.remove) {
+        await resource.remove(context);
+        callbacks.onResourceRemoved?.(resource);
+      } else {
+        callbacks.onResourceSkipped?.(resource);
+      }
+    }
+
+    for (const operation of [...(feature.operations ?? [])].reverse()) {
```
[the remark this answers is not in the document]
Krosovok: Currently only case for operation is to refresh git hooks: it would read patched file and un-register removed hooks. So for now the implementation is correct. If we would need more precise pre-removal, we will add it separately.

src/cli/commands/integrate/_common/registry/resources/text-snippet.ts
```diff
@@ -67,6 +67,20 @@ export class TextSnippet implements ResourceDeclaration {
...
+    if (!existing?.includes(this.startMarker) || !existing.includes(this.endMarker)) {
+      return;
+    }
+
+    const pattern = new RegExp(
+      String.raw`(?:\r?\n)?${escapeRegExp(this.startMarker)}[\s\S]*?${escapeRegExp(this.endMarker)}(?:\r?\n)?`,
+      'g',
+    );
+    await writeFileIfChanged(path, existing.replaceAll(pattern, ''));
+  }
```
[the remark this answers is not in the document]
Krosovok: Yes, we explicitly don't want to remove unspecified part of the file based only on the start marker. That's why we introduced the endMarker in the first place. If there is no end marker, we assume that the removePatch can delete more than it should and choose not not intervene. If the user modified the file to not include the endMarker, then we will probably do more harm than good by automatically removing arbitrary part of the file.

PR #370, opened 2026-06-03
CLI-438 Reset command

Krosovok:
Happy path:
image

Unknown example:
image

Schema:

image

----
## Summary by Gitar

- **New CLI command:**
  - Added `sonar system reset` to factory-reset the environment by purging credentials, binaries, integrations, and cached logs/data.
- **Cleanup implementation:**
  - Implemented `purgeAuth` for token revocation, `removeBinaries` with `safe-path` validation, `removeAllIntegrations` for declarative revert, and `clearFilesystem` for cache.
- **Safety guard:**
  - Introduced `resolveSafePath` to ensure binary removal is constrained to authorized directories (`BIN_DIR`, `CLI_DIR`).
- **Integration teardown:**
  - Added `removePatch` hooks to existing integration declarations (Copilot, Claude, Codex, Git) to allow clean uninstallation of patches.
- **Testing and documentation:**
  - Added `reset.test.ts` for end-to-end coverage and updated `CLAUDE.md` explaining the system reset design.


src/cli/commands/integrate/_common/registry/install-integration.ts
```diff
@@ -57,8 +56,18 @@ export interface InstallIntegrationOptions<TOptions> {
...
 }
 
+async function resolveIntegrationRegistry(
+  registry: IntegrationRegistry | undefined,
+): Promise<IntegrationRegistry> {
+  if (registry) {
+    return registry;
+  }
+  const { supportedIntegrations } = await import('../../index.js');
+  return supportedIntegrations;
```
Krosovok: Can we make it required and pass it from callers? I think it will break the circular dependency issue we are trying to address here.

src/cli/commands/system/reset-filesystem.ts
```diff
@@ -0,0 +1,79 @@
...
+      continue;
+    }
+
+    try {
+      if (existsSync(safeDir)) {
+        const dirBytes = directorySizeBytes(safeDir);
+        rmSync(safeDir, { recursive: true, force: true });
+        bytesFreed += dirBytes;
+        removedCount += 1;
+      }
```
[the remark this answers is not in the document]
nquinquenel: Not a problem

src/cli/commands/integrate/_common/registry/install-integration.ts
nquinquenel: I agree, good point

src/cli/commands/system/reset.ts
```diff
@@ -110,7 +125,7 @@ async function confirmDestructiveAction(): Promise<boolean> {
     return false;
   }
   warn(
-    'This will remove all local credentials, uninstall Sonar binaries, and break active tools integrations.',
+    'This will remove all local credentials, uninstall Sonar binaries, and break active AI integrations.',
```
Krosovok: I wanted to generalize because Git is also in there.

src/cli/commands/auth/revoke-server-token.ts
```diff
@@ -0,0 +1,64 @@
...
+    const detail = (error as Error).message;
+    if (context === 'logout') {
+      warn(
+        `Failed to revoke the server-side token "${connection.tokenName}": ${detail}. Continuing with local logout.`,
+      );
+      return;
+    }
+    warn(
+      `Failed to revoke the server-side token "${connection.tokenName}" for ${connection.serverUrl}: ${detail}`,
+    );
```
Krosovok: Instead of giving a `context` parameter I think it's better to return success/failure+error in a returned object. The method returns void anyway and if depends on the caller what we want to do with the result, it should be on the caller side. We can have distinction success/failure/skipped here as well as messages and add a line about Continuing on top of it in the `reset` caller.

nquinquenel: Makes sense, I'm returning a structured output now, and it's handled once in the parent via a new `reportRevokeOutcome` method

src/cli/commands/system/reset-integrations.ts
```diff
@@ -0,0 +1,159 @@
...
+): Promise<FeatureRemoveOutcome> {
+  const featureDeclaration = declaration.features.find(
+    (feature) => feature.id === installedFeature.featureId,
+  );
+  if (!featureDeclaration) {
+    return {
+      status: 'failed',
+      message: `${installed.integrationId}.${installedFeature.featureId}: unknown feature`,
+    };
+  }
```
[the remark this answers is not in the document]
nquinquenel: No need

nquinquenel (approved):

PR #377, opened 2026-06-04
CLI-437 Provide 'system status' command

kirill-knize-sonarsource:
----
## Summary by Gitar

- **New CLI command:**
  - Added `system status` command to display diagnostic information, including authentication, installed binaries, and integrations.
  - Added `--json` flag to `system status` for machine-readable output.
- **Diagnostics:**
  - Implemented logic to verify authentication status, check for CLI and binary updates, and validate MCP server configurations.
  - Added functionality to detect the status of MCP containers using the local container runtime.
- **Integration tests:**
  - Introduced a full suite of integration tests for `system status` covering various authentication states, binary presence, and update scenarios.
  - Added `FakeUpdateScriptServer` and `docker` mock capabilities to the test harness to support robust environment verification.
- **Code refinements:**
  - Updated `checkForUpdate` to support custom base URLs via an optional parameter for testability.
  - Enhanced `EnvironmentBuilder` to include a `docker` binary mock for testing MCP integration status.


kirill-knize-sonarsource:
PR is a bit large. 1200 lines, but 800 of them are tests.

src/cli/commands/system/status.ts
```diff
@@ -0,0 +1,495 @@
...
+      }
+    }
+  }
+  return features;
+}
+
+function checkMcpConfigFile(configPath: string): McpConfigStatus {
+  if (!existsSync(configPath)) return 'not_configured';
+  // TOML configs (e.g. Codex ~/.codex/config.toml) are not JSON — trust file existence
+  if (configPath.endsWith('.toml')) return 'configured';
```
Krosovok: Is this intentional? And if we want to address it later, maybe we can create a ticket to track it in CLI 1.0 maintenance?

Krosovok (approved):
LGTM, one clarifying question.

src/cli/commands/system/status.ts
kirill-knize-sonarsource: Good catch! Added toml config check.

PR #379, opened 2026-06-04
CLI-598 Git hook pre-commit framework delegates to sonar hook subcommands

georgii-borovinskikh-sonarsource:
Part of CLI-505

----
## Summary by Gitar

- **Pre-commit framework integration:**
  - Updated `upsertSonarHook` to delegate `pre-commit` and `pre-push` stages to specific `sonar hook` subcommands.
  - Configured `pass_filenames: true` for sonar hooks to allow direct file scanning via the framework.
- **CLI Command updates:**
  - Refactored `git-pre-commit` and `git-pre-push` to accept optional `[files...]` arguments for framework integration.
  - Added `getFilesToScan` logic to `git-pre-push` to support both explicit file lists and fallback git ref collection.
- **Logic refactoring:**
  - Optimized `hasUncommittedChanges` to use `git diff --quiet HEAD` instead of parsing `git status --porcelain`.
  - Consolidated file scanning for `git-pre-push` into a single binary invocation to improve performance.
- **Testing:**
  - Added comprehensive integration tests for `sonar hook git-pre-push` in file-scanning mode and unit tests for framework activation logic.

### DIFF
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -495,7 +495,8 @@ hookCommand
 hookCommand
   .command('git-pre-commit')
   .description('git pre-commit handler: scan staged files for secrets')
-  .anonymousAction(() => gitPreCommit());
+  .argument('[files...]', 'Changed files passed by pre-commit (pass_filenames: true)')
+  .anonymousAction((files: string[]) => gitPreCommit(files));
 
 hookCommand
   .command('git-pre-push')
@@ -507,7 +508,8 @@ hookCommand
     '--dependency-risks',
     'Also run a dependency-risks scan after the secrets scan (requires -p)',
   )
-  .anonymousAction((options: GitPrePushOptions) => gitPrePush(options));
+  .argument('[files...]', 'Changed files passed by pre-commit (pass_filenames: true)')
+  .anonymousAction((files: string[], options: GitPrePushOptions) => gitPrePush(options, files));
 
 // Hidden flush command — only registered when running as a telemetry worker.
 if (process.env[TELEMETRY_FLUSH_MODE_ENV]) {
--- a/src/cli/commands/hook/git-pre-commit.ts
+++ b/src/cli/commands/hook/git-pre-commit.ts
@@ -27,8 +27,8 @@ import { CommandFailedError } from '../_common/error';
 import { EXIT_CODE_SECRETS_FOUND, runSecretsBinary } from '../analyze/secrets';
 import { handleScanError, resolveAuthAndSecrets } from './hook-dependencies';
 
-export async function gitPreCommit(): Promise<void> {
-  const stagedFiles = await getStagedFiles();
+export async function gitPreCommit(files: string[] = []): Promise<void> {
+  const stagedFiles = files.length > 0 ? files : await getStagedFiles();
   if (stagedFiles.length === 0) return;
 
   const deps = await resolveAuthAndSecrets();
--- a/src/cli/commands/hook/git-pre-push.ts
+++ b/src/cli/commands/hook/git-pre-push.ts
@@ -49,8 +49,8 @@ export async function gitPrePush(
   const auth = await resolveAuth().catch(() => null);
   if (!auth) return;
 
-  const changedFiles = await getFilesToScan(files);
-  if (changedFiles === null) return;
+  const fileGroups = await getFileGroupsToScan(files);
+  if (fileGroups === null) return;
 
   if (await hasUncommittedChanges()) {
     warn(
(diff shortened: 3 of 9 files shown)


src/cli/command-tree.ts
```diff
@@ -481,7 +482,8 @@ hookCommand
     '--dependency-risks',
     'Also run a dependency-risks scan after the secrets scan (requires -p)',
   )
-  .anonymousAction((options: GitPrePushOptions) => gitPrePush(options));
+  .argument('[files...]', 'Changed files passed by pre-commit (pass_filenames: true)')
```
damien-urruty-sonarsource, line 511: We moved away from passing files from the hook when introducing the `hook` command, for many reasons. The idea is that the hook script is dumb and as much as possible of the logic is in `sonar hook`. Why would you need to pass files as argument and not compute them directly in the CLI?

src/cli/commands/hook/git-pre-commit.ts
```diff
@@ -27,8 +27,8 @@ import { CommandFailedError } from '../_common/error';
 import { EXIT_CODE_SECRETS_FOUND, runSecretsBinary } from '../analyze/secrets';
 import { handleScanError, resolveAuthAndSecrets } from './hook-dependencies';
 
-export async function gitPreCommit(): Promise<void> {
-  const stagedFiles = await getStagedFiles();
+export async function gitPreCommit(files: string[] = []): Promise<void> {
```
damien-urruty-sonarsource, line 30: Not directly related to your changes but I'm often confused between the native git pre commit hook and the pre-commit framework. Maybe we could reflect that this file is about the framework in the name?

src/cli/commands/integrate/git/tools/pre-commit/config.ts
```diff
@@ -55,11 +55,11 @@ export interface PreCommitConfig {
   [key: string]: unknown;
 }
 
-function buildSonarPreCommitHook(stage: GitHookType): PreCommitHookEntry {
+function buildSonarHook(stage: GitHookType): PreCommitHookEntry {
```
damien-urruty-sonarsource, line 58: Also not related to your changes, but in a couple of places we have a parameter called stage, and the type is GitHookType. I don't mind which name to use be we should be consistent to not bring confusion

src/cli/command-tree.ts
damien-urruty-sonarsource: After some discussion I understood that the pre-commit framework will provide the list of files. In this case it's an acceptable API

src/cli/commands/hook/git-pre-commit.ts
damien-urruty-sonarsource, line 30: Do we need a default empty array? It looks like we always pass something

src/cli/commands/hook/git-pre-push.ts
```diff
@@ -38,34 +38,47 @@ export interface GitPrePushOptions {
   dependencyRisks?: boolean;
 }
 
-export async function gitPrePush(options: GitPrePushOptions = {}): Promise<void> {
+export async function gitPrePush(
+  options: GitPrePushOptions = {},
+  files: string[] = [],
```
damien-urruty-sonarsource, line 43: Same

damien-urruty-sonarsource (approved):
Some suggestions. Also not directly related to your changes, but I don't think we have E2E tests for Git integration, e.g. starting from `integrate git` and checking that a commit/push is blocked, could you create a follow-up ticket?

src/cli/commands/hook/git-pre-push.ts
```diff
@@ -49,34 +49,44 @@ export async function gitPrePush(
...
-  } else {
-    const refs = await readGitPushRefs();
-    if (refs.length === 0) return null;
-
-    const emptyTree = await getEmptyTree();
-    const nonDeletionRefs = refs.filter((ref) => ref.localSha !== GIT_NULL_OID);
-    const filesByRef = await collectFilesForRefs(nonDeletionRefs, emptyTree);
-    return [...new Set(Array.from(filesByRef.values()).flat())];
+    /*
+     * pre-commit framework pre-chunks files before calling our tool in parallel.
```
georgii-borovinskikh-sonarsource, line 74: cc @damien-urruty-sonarsource

src/cli/commands/hook/git-pre-commit.ts
georgii-borovinskikh-sonarsource: it's useful for unit tests, so we don't have to repeat `[]` everywhere

src/cli/commands/hook/git-pre-push.ts
georgii-borovinskikh-sonarsource: https://github.com/SonarSource/sonarqube-cli/pull/379#discussion_r3371834601

src/cli/commands/integrate/git/tools/pre-commit/config.ts
georgii-borovinskikh-sonarsource: I agree, but this will cause merge conflicts for my next PR

src/cli/commands/hook/git-pre-commit.ts
georgii-borovinskikh-sonarsource: I agree, but this will cause merge conflicts for my next PR

src/cli/commands/hook/git-pre-push.ts
georgii-borovinskikh-sonarsource: This is a fundamental problem with pre-push hooks, we will consider changing SCA analysis to pre-commit hook - but the overall intent of this PR stays the same - proceeding without addressing this

PR #382, opened 2026-06-04
CLI-499 skip SQAA during global integration

sophio-japharidze-sonarsource:
----
## Summary by Gitar

- **Refactored SQAA integration:**
  - Added `resolveSqaaSetup` to centralize entitlement checks and `--global` installation logic.
  - Enforced skipping SQAA during global integration across `claude`, `codex`, and `copilot` commands.
- **Improved user feedback:**
  - Unified the skip notice for SQAA and Context Augmentation when using `--global`.
  - Restricted skip warnings to only show for organizations actually entitled to the features to reduce noise.
- **Test coverage:**
  - Updated integration and unit tests to verify that SQAA and Context Augmentation are correctly skipped and warned about during global installations.


nquinquenel (approved):
LGTM. I think there might be an optimization to make: we currently check for organization lookup twice, both for SQAA and CAG when we integrate. We could get this info only once.

Also, we end up always checking entitlement in global mode, even if it will always return false (because we want to show a log)  - but this one is on purpose so it's ok.

sophio-japharidze-sonarsource:
@nquinquenel I created https://sonarsource.atlassian.net/browse/CLI-603 for optimization followup

PR #404, opened 2026-06-05
CLI-614 Add hook markers migration

georgii-borovinskikh-sonarsource:
Part of CLI-505

----
## Summary by Gitar

- **Git Hook Migration:**
  - Transitioned to per-stage IDs (`sonar-pre-commit`, `sonar-pre-push`) to allow concurrent hook management, replacing the legacy `sonar-secrets` ID.
  - Added declarative `legacyCleanups` to handle idempotency and migration of old hook artifacts.
- **Resource Management:**
  - Replaced the `nativeGitHookResource` with a flexible `wholeFile` resource and introduced a `RemovableResource` interface to standardize cleanup.
  - Refactored `TextSnippet`, `WholeFile`, and patch resources to use dedicated remover classes for safer, idempotent cleanup.
- **Hook Identification:**
  - Implemented `markers.ts` to manage distinct, framework-aware markers for Git hooks, replacing a single global marker for improved detection.
- **UX Improvements:**
  - Refined error messages for file overwrites to remove prescriptive `--force` advice, providing context-aware remediation hints instead.


src/cli/commands/integrate/_common/registry/resources/common.ts
```diff
@@ -35,14 +35,20 @@ export interface BaseResourceOptions {
...
 
-export interface ResourceDeclaration {
+export interface ResourceIdentity {
   id: string;
-  displayName?: string;
   resourceType: string;
   version?: string;
+}
+
+export interface RemovableResource {
```
damien-urruty-sonarsource, line 44: What is the benefit of having a separate interface for that? Why not keeping `remove` as part of the declaration? It looks like it introduces 2 type hierarchies that mirror each other, I'm not sure it makes the design cleaner

EDIT: I see it's used for `legacyCleanup`, why not accepting the resource there directly?

src/cli/commands/integrate/git/tools/markers.ts
```diff
@@ -0,0 +1,61 @@
...
+// only so existing installs are recognized and migrated; it is never emitted for new installs.
+
+import type { GitHookType } from '../options';
+import { LEGACY_HOOK_MARKER } from './shared';
+
+// Native (whole-file) ownership comment — a descriptive header, one per hook type.
+const NATIVE_HOOK_MARKERS: Record<GitHookType, string> = {
+  'pre-commit': 'sonar pre-commit hook - installed by sonar integrate git',
+  'pre-push': 'sonar pre-push hook - installed by sonar integrate git',
+};
```
damien-urruty-sonarsource: Those constant should be moved to each specific tool we support, each tool is in a separate folder, and each constant should belong there

damien-urruty-sonarsource (commented):
I find this new type hierarchy a bit too over-engineered. It makes things a bit more complex to follow, and I think we could keep it simpler. Is this something you would consider simplifying?

Some other remarks:
* there is no description in the ticket, neither in the child or in the parent. Could you please update at least one of them, so we know why we make this change? It's quite useful for reviewers, out future selves, but also when we build the release notes
* the PR is still quite big, and contributes a lot of things: modifying the framework, fixing husky, fixing pre-commit. I'm sure we could have split it in a better way. Here the changeset is pretty big so it's easy to miss things
* the commit/ticket message/PR title is a bit vague, I'm sure we could find something better. Maybe a consequence of the previous point

src/cli/commands/integrate/_common/registry/resources/common.ts
georgii-borovinskikh-sonarsource: It is needed as a separate interface as the Declaration requires `apply`, which may contain (at that point) legacy generation logic which should be removed - I don't think it's right to force the caller to keep it or manually replacing with a stub

I think it's a simple case of interface segregation. I could go further and split it into `ApplicableResource` with `apply` and `isApply`, while keeping the main `ResourceDeclaration` almost empty, but there was no need at the moment to introduce a separate 'applicable' entity, so I decided not to do it

georgii-borovinskikh-sonarsource:
> I find this new type hierarchy a bit too over-engineered. It makes things a bit more complex to follow, and I think we could keep it simpler. Is this something you would consider simplifying?
> 
> Some other remarks:
> 
> * there is no description in the ticket, neither in the child or in the parent. Could you please update at least one of them, so we know why we make this change? It's quite useful for reviewers, out future selves, but also when we build the release notes
> * the PR is still quite big, and contributes a lot of things: modifying the framework, fixing husky, fixing pre-commit. I'm sure we could have split it in a better way. Here the changeset is pretty big so it's easy to miss things
> * the commit/ticket message/PR title is a bit vague, I'm sure we could find something better. Maybe a consequence of the previous point

I think in fact point 2 answers the question of point 1 - the need is specifically to be able to have hook migrations without having to implement the logic as part of the `apply` method. The benefit of this is that the removal logic is only kept specific to the given format, while the overall logic of which removal to apply is moved to the generic place inside the declarative framework.

This change also allows to in the future do more complicated migrations in the declarative way: you could use Whole File legacy remover with a new File Patch resource declaration if you moved a feature from a standalone file into a shared one, for example.

I will put this same information in the ticket, I did forget to update it unfortunately

I understand your point regarding having the framework changes and hook fixes being merged together. I think it would have been harder to review if the intent was only stated in the ticket (as opposed to be included in the code changes like it is now), however I can split this into 2 PRs as it would be cleaner from the git history pov.

src/cli/commands/integrate/_common/registry/resources/common.ts
damien-urruty-sonarsource: OK got it

damien-urruty-sonarsource:
@georgii-borovinskikh-sonarsource I already reviewed so now we can continue with this PR, no need to split. But for next time I think it makes sense to try and reduce the size. The split I suggested above is one possibility, but there are maybe others

damien-urruty-sonarsource (approved):

PR #407, opened 2026-06-08
CLI-549 Add sonar integrate antigravity command and install orchestration

nquinquenel:
Register integrate antigravity in the command tree, add integrateAntigravity() orchestration, register antigravity-cli in ALL_INTEGRATIONS, and CAG.

More to come

----
## Summary by Gitar

- **New Antigravity integration:**
  - Registered `antigravity` command under `sonar integrate` with support for project and global scopes.
  - Implemented orchestration to configure secrets scanning hooks, Agentic Analysis, and MCP server setup.
- **Configuration & Infrastructure:**
  - Added Antigravity configuration constants for local (`.agents/`) and global (`~/.gemini/config`) workspaces.
  - Introduced `antigravity-cli` registry declaration to manage integration lifecycle and context augmentation.
- **Testing:**
  - Added comprehensive integration tests covering successful setup, idempotency, global configuration, and flag validation.
  - Included unit tests for resolving installation targets based on scope.


src/cli/command-tree.ts
```diff
@@ -247,6 +248,23 @@ integrateCommand
   .addHelpText('after', projectKeyExtraHelp)
   .authenticatedAction((auth, options: IntegrateAgentOptions) => integrateCodex(options, auth));
 
+integrateCommand
+  .command('antigravity')
+  .description(
+    'Setup SonarQube integration for Antigravity. This will install secrets scanning hooks, configure SonarQube Agentic Analysis and MCP Server.',
+  )
```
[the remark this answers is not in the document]
nquinquenel: It's expected for now, it's only the first part

src/cli/commands/integrate/antigravity/declaration.ts
```diff
@@ -0,0 +1,74 @@
...
+import type { IntegrationContext, IntegrationDeclaration } from '../_common/registry';
+import { askUser } from '../_common/registry';
+import type { IntegrateAgentOptions } from '../_common/types';
+
+export const ANTIGRAVITY_INTEGRATION_ID = 'antigravity-cli';
+
+export interface AntigravityIntegrationOptions extends IntegrateAgentOptions {
+  projectRoot?: string;
+  globalSecretsHookExists?: boolean;
+  /** Install PostInvocation SQAA hook (project scope). Wired in CLI-551. */
```
sophio-japharidze-sonarsource: I think it's fine to drop this comment

src/cli/commands/integrate/antigravity/declaration.ts
```diff
@@ -0,0 +1,74 @@
...
+  installSqaaHook?: boolean;
+  sqaaEntitled?: boolean;
+  installContextAugmentation?: boolean;
+}
+
+/**
+ * Minimal declarative integration for CLI-549 orchestration. Secrets hooks,
```
sophio-japharidze-sonarsource: TMI ? :D

src/cli/commands/integrate/antigravity/declaration.ts
```diff
@@ -0,0 +1,74 @@
...
+  id: ANTIGRAVITY_INTEGRATION_ID,
+  displayName: 'Antigravity',
+  features: [
+    {
+      id: 'integration-setup',
+      displayName: 'Antigravity integration setup',
+      shouldInstall: () => askUser(),
+    },
+    createContextAugmentationFeature<AntigravityIntegrationOptions>({
+      agentDisplayName: 'Antigravity',
```
sophio-japharidze-sonarsource, line 52: Will the same integration also work with Antigravity IDE? Or only CLI? Depending on the answer we might want to update this display name or the integration ID on line 32

src/cli/commands/integrate/index.ts
```diff
@@ -28,6 +29,7 @@ export const ALL_INTEGRATIONS = [
   claudeIntegration,
   copilotIntegration,
   codexIntegration,
+  antigravityIntegration,
```
sophio-japharidze-sonarsource: I think at some point we said we should not have feature branches. So instead I would suggest removing this change / making `antigravity` command hidden, and merging directly to master. We can also discuss with Damien.

src/cli/command-tree.ts
```diff
@@ -247,6 +248,23 @@ integrateCommand
   .addHelpText('after', projectKeyExtraHelp)
   .authenticatedAction((auth, options: IntegrateAgentOptions) => integrateCodex(options, auth));
 
+integrateCommand
```
sophio-japharidze-sonarsource, line 305: See comment above. Maybe it's best to keep the groundwork but remove wiring from customer-facing surfaces.

src/lib/config-constants.ts
```diff
@@ -69,6 +69,42 @@ export const BIN_DIR = join(CLI_DIR, 'bin');
 /** Directory used for global git hooks when core.hooksPath is set via sonar integrate git --global. */
 export const GLOBAL_HOOKS_DIR = join(CLI_DIR, 'hooks');
 
+// ---------------------------------------------------------------------------
+// Antigravity
+// ---------------------------------------------------------------------------
+
+/** Antigravity workspace agents directory (relative to the project root). */
+export const ANTIGRAVITY_PROJECT_AGENTS_DIR = '.agents';
```
sophio-japharidze-sonarsource, line 91: I might be wrong but I vaguely remember Codex also having a similar path? Just something to double-check, not to have duplicate or overwritten integrations

tests/unit/cli/commands/integrate/antigravity/integrate.test.ts
sophio-japharidze-sonarsource, line 1: Do we really need this file? I think similar stuff is covered in the integration test

sophio-japharidze-sonarsource (commented):
Looks good! I think it's great to go in small increments. I would suggest to target master branch directly but not declare the new feature in command-tree.

src/cli/commands/integrate/antigravity/declaration.ts
nquinquenel: I like to write comments on feature branch to make sure I don't forget about implementation of each tickets, but thanks for suggesting to target master directly! I think it makes sense, I'll remove all the temporary bits I added

src/cli/commands/integrate/antigravity/declaration.ts
nquinquenel: Both IDE and CLI, we patch `.agents/` and `~/.gemini/config`, which Google documents as the shared harness. So I kept Antigravity as display name and antigravity-cli as integration id (same pattern as copilot-cli). Open to renaming the id if we prefer something less CLI-specific

src/lib/config-constants.ts
nquinquenel: Yes, Codex only uses `.agents` for the CAG skill, hooks/MCP live under `.codex/`. Antigravity uses `.agents/`  for the full layout, so the only overlap is that skill path. Same file if both are integrated, but same rendered content; integrations don’t overwrite each other's hooks/MCP/state

src/cli/command-tree.ts
```diff
@@ -301,6 +302,23 @@ integrateCommand
...
+  .command('antigravity', { hidden: true })
+  .description(
+    'Setup SonarQube integration for Antigravity. This will install secrets scanning hooks, configure SonarQube Agentic Analysis and MCP Server.',
```
sophio-japharidze-sonarsource, line 308: I think there were some discussions about defining exactly what we mean by each agent name. For example here specifying that we mean `Antigravity (CLI & IDE)` ?? To be checked with UX

sophio-japharidze-sonarsource (approved):

src/cli/command-tree.ts
nquinquenel: Antigravity refers to both CLI & IDE, there's no differentiation and that's the way they mention it in the documentation. There are products: Antigravity CLI and Antigravity IDE, but skills/hooks/etc are just Antigravity skills/hooks/etc

PR #411, opened 2026-06-09
CLI-635 system reset must not tear down pre-commit framework hooks

nquinquenel:
----
## Summary by Gitar

- **Refactored pre-commit integration:**
  - Removed `deactivatePreCommitFramework` to prevent accidental teardown of the entire framework during system resets.
- **Testing:**
  - Added unit tests to verify that removing Sonar hooks preserves other local and third-party pre-commit hooks.

### DIFF
--- a/src/cli/commands/integrate/git/tools/pre-commit/config.ts
+++ b/src/cli/commands/integrate/git/tools/pre-commit/config.ts
@@ -190,3 +190,12 @@ export async function activatePreCommitFramework(root: string, hook: GitHookType
     );
   }
 }
+
+/** Best-effort pre-commit gc on Sonar hook removal; never throws. */
+export async function garbageCollectPreCommitFramework(root: string): Promise<void> {
+  try {
+    await runPreCommitCommand(['gc'], root);
+  } catch {
+    // Missing framework or gc failure — non-fatal for reset.
+  }
+}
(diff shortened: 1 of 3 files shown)


src/cli/commands/integrate/git/tools/pre-commit/config.ts
```diff
@@ -190,13 +190,3 @@ export async function activatePreCommitFramework(root: string, hook: GitHookType
     );
   }
 }
-
-/** Best-effort pre-commit uninstall during factory reset; never throws. */
-export async function deactivatePreCommitFramework(root: string): Promise<void> {
-  try {
-    await runPreCommitCommand(['uninstall'], root);
-    await runPreCommitCommand(['clean'], root);
```
damien-urruty-sonarsource, line 198: From what I read, we could at least call `pre-commit gc` that will do some clean-up. But it's a nice-to-have:

> The goal of pre-commit gc (garbage collection) is to clean up and delete unused hook environments to free up disk space. 
When you use the pre-commit framework, it doesn't just run tools globally from your machine. Instead, it creates isolated, virtual environments (like individual Python virtualenvs or Node environments) for every single hook repository you use, specifically locked to the exact version you defined in your .pre-commit-config.yaml.
Over time, as you update hook versions or remove hooks from your configuration, the old environments just sit there, taking up space. pre-commit gc scans your global cache and purges those ghost environments.

damien-urruty-sonarsource (approved):
LGTM, one suggestion

src/cli/commands/integrate/git/tools/pre-commit/config.ts
nquinquenel: Makes sense! I'm not super familiar with this framework

PR #419, opened 2026-06-09
CLI-639 Mark analyze dependency-risks as beta

georgii-borovinskikh-sonarsource:
Part of CLI-202

----
## Summary by Gitar

- **Beta Status:**
  - Marked `analyze dependency-risks` as beta in the CLI description and help output.
  - Added a runtime warning to notify users that dependency manifest files are uploaded to SonarQube.

### DIFF
--- a/src/cli/commands/analyze/dependency-risks.ts
+++ b/src/cli/commands/analyze/dependency-risks.ts
@@ -49,7 +49,7 @@ export async function analyzeDependencyRisks(
   options: AnalyzeDependencyRisksOptions,
   auth: ResolvedAuth,
 ): Promise<void> {
-  warn('analyze dependency-risks is in Beta');
+  warn(`'analyze dependency-risks' is in Beta`);
   warn(
     'Dependency manifest files (e.g. package-lock.json, pom.xml) will be uploaded to SonarQube for analysis.\n' +
       '  → Learn more: https://docs.sonarsource.com/sonarqube-server/advanced-security/analyzing-projects-for-dependencies#supported-languages-and-package-managers',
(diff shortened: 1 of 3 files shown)


src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -49,6 +49,12 @@ export async function analyzeDependencyRisks(
   options: AnalyzeDependencyRisksOptions,
   auth: ResolvedAuth,
 ): Promise<void> {
+  warn('analyze dependency-risks is in Beta');
+  warn(
```
sophio-japharidze-sonarsource, line 53: Is it a bit too late to warn about this? Should we ask for a confirmation before launching the analysis instead?

src/cli/commands/analyze/dependency-risks.ts
```diff
@@ -49,6 +49,12 @@ export async function analyzeDependencyRisks(
   options: AnalyzeDependencyRisksOptions,
   auth: ResolvedAuth,
 ): Promise<void> {
+  warn('analyze dependency-risks is in Beta');
```
sophio-japharidze-sonarsource:
```suggestion
  warn(`'analyze dependency-risks' is in Beta`);
```

sophio-japharidze-sonarsource (approved):

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: Me & Clifford discussed this and decided to go with a simple warning, given we're marking this as beta and the warning will be removed in the next update.

sophio-japharidze-sonarsource (approved):

src/cli/commands/analyze/dependency-risks.ts
georgii-borovinskikh-sonarsource: @sophio-japharidze-sonarsource do you consider this critical? the main reason why this wasn't done is that we would need to add interactive confirmation, `--no-warn` option and/or 'don't ask again' mechanic for a temporary warning

sophio-japharidze-sonarsource: All good, approved.

PR #425, opened 2026-06-09
CLI-651 Preserve existing line endings when writing managed resource blocks

georgii-borovinskikh-sonarsource:
Part of CLI-505

----
## Summary by Gitar

- **EOL handling utilities:**
  - Added `detectEol` to identify existing file line endings (`\r\n` vs `\n`).
  - Added `toEol`, `equalsIgnoringEol`, and `includesIgnoringEol` to standardize or compare content regardless of line endings.
- **Resource management updates:**
  - Updated `TextSnippet` and `WholeFileResource` to detect and apply the existing file's EOL convention when writing managed blocks.
  - Modified `isApplied` logic to treat files with matching content but different line endings as already applied.
- **Test coverage:**
  - Added `resources-eol.test.ts` to verify EOL preservation for both `wholeFile` and `textSnippet` resource types.

### DIFF
--- a/src/cli/commands/integrate/_common/registry/resources/common.ts
+++ b/src/cli/commands/integrate/_common/registry/resources/common.ts
@@ -20,6 +20,7 @@
 
 import { existsSync } from 'node:fs';
 import { chmod, mkdir, readFile, writeFile } from 'node:fs/promises';
+import { EOL } from 'node:os';
 import { dirname } from 'node:path';
 
 import type { AppliedResource, IntegrationContext, MaybePromise } from '../types';
@@ -75,6 +76,32 @@ export async function readTextFile(path: string): Promise<string | undefined> {
     return undefined;
   }
   return readFile(path, 'utf-8');
+}
+
+/** Returns true when `a` and `b` have the same content, ignoring differences in line endings. */
+export function equalsIgnoringEol(a: string, b: string): boolean {
+  return toEol(a, '\n') === toEol(b, '\n');
+}
+
+/** Rewrites every line ending in `value` to `eol`. */
+export function toEol(value: string, eol: string): string {
+  return value.replaceAll('\r\n', '\n').replaceAll('\n', eol);
+}
+
+/**
+ * Returns the newline sequence the content already uses (`\r\n` or `\n`), so generated text
+ * can match the file's existing convention. Falls back to the OS default when both are present.
+ */
+export function detectEol(content: string): string {
+  const containsCrlf = content.includes('\r\n');
+  const containsLf = /(?<!\r)\n/.test(content);
+  if (containsCrlf && containsLf) {
+    return EOL;
+  }
+  if (containsCrlf) {
+    return '\r\n';
+  }
+  return '\n';
 }
 
 export interface PatchResourceOptions<TDoc = unknown> extends BaseResourceOptions {
--- a/src/cli/commands/integrate/_common/registry/resources/common.ts
+++ b/src/cli/commands/integrate/_common/registry/resources/common.ts
@@ -80,7 +80,11 @@ export async function readTextFile(path: string): Promise<string | undefined> {
 
 /** Returns true when `a` and `b` have the same content, ignoring differences in line endings. */
 export function equalsIgnoringEol(a: string, b: string): boolean {
-  return toEol(a, '\n') === toEol(b, '\n');
+  return toEol(a, EOL) === toEol(b, EOL);
+}
+
+export function includesIgnoringEol(content: string, substring: string): boolean {
+  return toEol(content, EOL).includes(toEol(substring, EOL));
 }
 
 /** Rewrites every line ending in `value` to `eol`. */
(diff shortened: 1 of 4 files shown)


src/cli/commands/integrate/_common/registry/resources/common.ts
```diff
@@ -77,6 +78,32 @@ export async function readTextFile(path: string): Promise<string | undefined> {
...
+ * Returns the newline sequence the content already uses (`\r\n` or `\n`), so generated text
+ * can match the file's existing convention. Falls back to the OS default when both are present.
+ */
+export function detectEol(content: string): string {
+  const containsCrlf = content.includes('\r\n');
+  const containsLf = /(?<!\r)\n/.test(content);
+  if (containsCrlf && containsLf) {
+    return EOL;
+  }
+  if (containsCrlf) {
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: unsupported

src/cli/commands/integrate/_common/registry/resources/text-snippet.ts
```diff
@@ -103,14 +107,15 @@ export class TextSnippet implements ResourceDeclaration {
 
     const startMarkerIndex = existing.indexOf(this.startMarker);
     if (startMarkerIndex >= 0) {
-      return `${existing.slice(0, startMarkerIndex)}${managedBlock}\n`;
+      return `${existing.slice(0, startMarkerIndex)}${managedBlock}${eol}`;
     }
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: The test case is already present — line 135 of resources-eol.test.ts covers exactly the replace path: a CRLF file with a pre-existing managed block, re-applied, keeps CRLF. The quality finding is already addressed.

src/cli/commands/integrate/_common/registry/resources/common.ts
```diff
@@ -77,6 +78,32 @@ export async function readTextFile(path: string): Promise<string | undefined> {
...
+export function toEol(value: string, eol: string): string {
+  return value.replaceAll('\r\n', '\n').replaceAll('\n', eol);
+}
+
+/**
+ * Returns the newline sequence the content already uses (`\r\n` or `\n`), so generated text
+ * can match the file's existing convention. Falls back to the OS default when both are present.
+ */
+export function detectEol(content: string): string {
+  const containsCrlf = content.includes('\r\n');
```
damien-urruty-sonarsource, line 100: Maybe way too edge case, but it could happen that the file has a mix of different line endings. We could try to find the closest EOL where the text should inserted. Not worth

src/cli/commands/integrate/_common/registry/resources/common.ts
```diff
@@ -77,6 +78,32 @@ export async function readTextFile(path: string): Promise<string | undefined> {
...
+  const containsLf = /(?<!\r)\n/.test(content);
+  if (containsCrlf && containsLf) {
+    return EOL;
+  }
```
damien-urruty-sonarsource, line 104: Why do we have a separate branch in this case? I thought the next 2 branches would cover all cases?

src/cli/commands/integrate/_common/registry/resources/text-snippet.ts
```diff
@@ -69,7 +71,8 @@ export class TextSnippet implements ResourceDeclaration {
     if (existing === undefined) {
       return false;
     }
-    return existing.includes(this.renderManagedBlock(await this.resolveContent(context)));
+    const content = await this.resolveContent(context);
+    return existing.includes(this.renderManagedBlock(content, detectEol(existing)));
```
damien-urruty-sonarsource: As per the edge case I mentioned earlier, here `isApplied` would return false. Another option would be to completely ignore line ends for this check. 

EDIT: I see we do have a `equalsIgnoringEol` for whole-file, so it would be the equivalent for text-snippet

damien-urruty-sonarsource (approved):
Some nitpicks, LGTM

src/cli/commands/integrate/_common/registry/resources/common.ts
georgii-borovinskikh-sonarsource: I decided to simplify it. I think we could do different things: localized endings for where we want to insert, or calculate the majority endings, etc., but I didn't think it was worth it.

src/cli/commands/integrate/_common/registry/resources/common.ts
georgii-borovinskikh-sonarsource: If the file has mixed endings, instead of some complex calculation or a forced default, this returns the default ending of the system.

src/cli/commands/integrate/_common/registry/resources/text-snippet.ts
georgii-borovinskikh-sonarsource: In mixed line endings case this can return false, yes. Let me try to fix it...

georgii-borovinskikh-sonarsource: Fixed

PR #447, opened 2026-06-12
CLI-674 Support impact severity values in --severities filter

kirill-knize-sonarsource:
----
## Summary by Gitar

- **CLI improvements:**
  - Updated `--severities` filter to support both impact severity values (`HIGH`, `MEDIUM`, `LOW`) and legacy severity values.
  - Added validation logic to prevent mixing impact and legacy severities in the same command.
- **API and internal updates:**
  - Added `impactSeverities` parameter to `IssuesSearchParams` and request handlers to correctly route filtered requests to the API.
  - Updated command-line help text to clarify supported severity types for users.
- **Testing:**
  - Added comprehensive unit and integration tests to verify parameter routing and invalid input handling.

### DIFF
--- a/src/cli/commands/list/issues.ts
+++ b/src/cli/commands/list/issues.ts
@@ -33,7 +33,8 @@ import { print } from '../../../ui';
 import { InvalidOptionError } from '../_common/error';
 
 export const VALID_FORMATS = ['json', 'toon', 'table', 'csv'];
-export const VALID_SEVERITIES = ['INFO', 'MINOR', 'MAJOR', 'CRITICAL', 'BLOCKER'];
+export const VALID_STANDARD_SEVERITIES = ['INFO', 'MINOR', 'MAJOR', 'CRITICAL', 'BLOCKER'];
+export const VALID_MQR_SEVERITIES = ['INFO', 'LOW', 'MEDIUM', 'HIGH', 'BLOCKER'];
 export const VALID_STATUSES = ['OPEN', 'CONFIRMED', 'FALSE_POSITIVE', 'ACCEPTED', 'FIXED'];
 
 export interface ListIssuesOptions {
@@ -49,6 +50,24 @@ export interface ListIssuesOptions {
   format?: string;
   pageSize: number;
   page: number;
+}
+
+function normalizeSeverityValues(raw: string): string[] {
+  return raw.split(',').map((s) => s.trim().toUpperCase());
+}
+
+function parseSeverities(
+  raw: string,
+  mode: 'mqr' | 'standard',
+): { severities?: string; impactSeverities?: string } {
+  const values = normalizeSeverityValues(raw);
+  const validSet = mode === 'mqr' ? VALID_MQR_SEVERITIES : VALID_STANDARD_SEVERITIES;
+  if (values.some((s) => !validSet.includes(s))) {
+    throw new InvalidOptionError(
+      `Invalid severity(es): '${raw}'. Valid values for ${mode === 'mqr' ? 'Multi-Quality Rule (MQR)' : 'Standard Experience'} mode: ${validSet.join(', ')}.`,
+    );
+  }
+  return mode === 'mqr' ? { impactSeverities: values.join(',') } : { severities: values.join(',') };
 }
 
 /**
@@ -89,24 +108,32 @@ export async function listIssues(options: ListIssuesOptions, auth: ResolvedAuth)
     normalizedStatuses = statuses.join(',');
   }
 
-  let normalizedSeverities = options.severities;
-  if (normalizedSeverities) {
-    const severities = normalizedSeverities.split(',').map((s) => s.toUpperCase());
-    if (!severities.every((s) => VALID_SEVERITIES.includes(s))) {
+  if (options.severities) {
+    const preflightValues = normalizeSeverityValues(options.severities);
+    if (
+      preflightValues.some(
+        (s) => !VALID_STANDARD_SEVERITIES.includes(s) && !VALID_MQR_SEVERITIES.includes(s),
+      )
+    ) {
       throw new InvalidOptionError(
-        `Invalid severity(es): '${options.severities}'. Valid severities are: ${VALID_SEVERITIES.join(', ')}`,
+        `Invalid severity(es): '${options.severities}'. ` +
+          `Multi-Quality Rule (MQR) mode values: ${VALID_MQR_SEVERITIES.join(', ')}. ` +
+          `Standard Experience mode values: ${VALID_STANDARD_SEVERITIES.join(', ')}.`,
       );
     }
-    normalizedSeverities = severities.join(',');
   }
 
   const client = new SonarQubeClient(auth.serverUrl, auth.token);
   const issuesClient = new IssuesClient(client);
 
+  const { severities: normalizedSeverities, impactSeverities: normalizedImpactSeverities } =
+    options.severities ? parseSeverities(options.severities, await client.getServerMode()) : {};
+
   const params: IssuesSearchParams = {
     projects: options.project,
     organization: auth.orgKey,
     severities: normalizedSeverities,
+    impactSeverities: normalizedImpactSeverities,
     types: options.type,
     issueStatuses: normalizedStatuses,
     rules: options.rule,
--- a/src/cli/command-tree.ts
+++ b/src/cli/command-tree.ts
@@ -76,7 +76,8 @@ import {
   listIssues,
   type ListIssuesOptions,
   VALID_FORMATS,
-  VALID_SEVERITIES,
+  VALID_MQR_SEVERITIES,
+  VALID_STANDARD_SEVERITIES,
   VALID_STATUSES,
 } from './commands/list/issues';
 import { listProjects, type ListProjectsOptions } from './commands/list/projects';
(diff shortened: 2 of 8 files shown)


src/cli/commands/list/issues.ts
```diff
@@ -33,7 +33,9 @@ import { print } from '../../../ui';
 import { InvalidOptionError } from '../_common/error';
 
 export const VALID_FORMATS = ['json', 'toon', 'table', 'csv'];
-export const VALID_SEVERITIES = ['INFO', 'MINOR', 'MAJOR', 'CRITICAL', 'BLOCKER'];
+export const VALID_LEGACY_SEVERITIES = ['INFO', 'MINOR', 'MAJOR', 'CRITICAL', 'BLOCKER'];
+export const VALID_IMPACT_SEVERITIES = ['HIGH', 'MEDIUM', 'LOW'];
```
georgii-borovinskikh-sonarsource: I think impact severities were extended at some point to include BLOCKER and INFO

src/cli/commands/list/issues.ts
```diff
@@ -51,6 +53,27 @@ export interface ListIssuesOptions {
...
+  const hasImpact = values.some((s) => VALID_IMPACT_SEVERITIES.includes(s));
+  const hasInvalid = values.some(
+    (s) => !VALID_LEGACY_SEVERITIES.includes(s) && !VALID_IMPACT_SEVERITIES.includes(s),
+  );
+  if (hasInvalid) {
+    throw new InvalidOptionError(
+      `Invalid severity(es): '${raw}'. Use legacy values (${VALID_LEGACY_SEVERITIES.join(', ')}) or impact values (${VALID_IMPACT_SEVERITIES.join(', ')}).`,
+    );
+  }
+  if (hasLegacy && hasImpact) {
```
georgii-borovinskikh-sonarsource: this logic will need to be adjusted due to blocker and info MQR severities

src/cli/commands/list/issues.ts
```diff
@@ -51,6 +52,20 @@ export interface ListIssuesOptions {
...
+): { severities?: string; impactSeverities?: string } {
+  const values = raw.split(',').map((s) => s.trim().toUpperCase());
+  const validSet = mode === 'mqr' ? VALID_MQR_SEVERITIES : VALID_STANDARD_SEVERITIES;
+  if (values.some((s) => !validSet.includes(s))) {
+    throw new InvalidOptionError(
+      `Invalid severity(es): '${raw}'. Valid values for ${mode === 'mqr' ? 'Multi-Quality Rule (MQR)' : 'Standard Experience'} mode: ${validSet.join(', ')}.`,
+    );
+  }
+  return mode === 'mqr' ? { impactSeverities: values.join(',') } : { severities: values.join(',') };
+}
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: It is intended.

src/cli/commands/list/issues.ts
kirill-knize-sonarsource: Great catch! Reworked to check an internal API to get the severities mode.

src/cli/commands/list/issues.ts
kirill-knize-sonarsource: Done

georgii-borovinskikh-sonarsource (approved):

kirill-knize-sonarsource:
> Code Review 👍 Approved with suggestions 6 resolved / 7 findings
> Adds impact severity support to the --severities filter, but the implementation lacks robust error handling for mode endpoints, introduces misleading NotFoundError messages, and contains duplicated severity validation logic.
> 
> 💡 Bug: 404 GETs now throw NotFoundError instead of 'Access denied' msg
> 📄 [src/sonarqube/client.ts:88-98](https://github.com/SonarSource/sonarqube-cli/pull/447/files#diff-d453cd78d77948fb69a12ad5cada15382b899808aca0baa3df8cf29e99307292R88-R98)
> 
> The change makes every GET that receives HTTP 404 throw the new `NotFoundError` ("Resource not found (404).") instead of the previous generic `Error` with message "Access denied (HTTP 404). Check that the supplied token and organization are valid." This affects all callers of `get()`/`getSafe()`, not just `getServerMode()`.
> 
> While most existing catch blocks swallow errors generically and remain functionally correct, any caller (or test) that matched on the old "Access denied" message string, or that relied on a 404 producing that user-facing remediation hint, will now surface a different, less actionable message. Verify that no command-level error handling or unit test depends on the old 404 message, and confirm the new generic message is acceptable for end users encountering a 404 (e.g. unknown project/component).
> 
> ✅ 6 resolved
> ✅ Edge Case: parseSeverities does not trim whitespace around values
> > 📄 src/cli/commands/list/issues.ts:55-67
> > `parseSeverities` splits on `,` and uppercases but does not `.trim()` each token. A common invocation like `--severities "HIGH, MEDIUM"` produces a token `" MEDIUM"` that is not in `validSet`, causing a spurious `Invalid severity(es)` error. This matches the pre-existing behavior but is a new helper worth hardening. Suggested fix: trim each value during the map.
> 
> ✅ Quality: getServerMode silently falls back to 'standard' on error
> > 📄 src/sonarqube/client.ts:259-267
> > When `/api/v2/clean-code-policy/mode` errors (network issue, older server version, transient 5xx), `getServerMode` swallows the error and returns `'standard'`. On an actual MQR server this then makes valid MQR severity values (HIGH/MEDIUM/LOW) be rejected with a misleading `Invalid severity` message, and silently routes filters to the wrong API parameter. Consider logging the failure (debug) and/or documenting the fallback so the confusing downstream validation error is traceable.
> 
> ✅ Quality: getServerMode relies on brittle error-message substring match
> > 📄 src/sonarqube/client.ts:259-270 📄 src/sonarqube/client.ts:88-100
> > `getServerMode` decides whether an old server lacks MQR support by catching the error and testing `err.message.includes('HTTP 404')`. This works today only because `raiseForStatus` throws a plain `Error` whose message happens to contain `(HTTP 404)`. This couples mode detection to a free-text message rather than a status code or typed error, so any refactor of error formatting (e.g. introducing a typed `NotFoundError`, which is already a tracked suggestion for this file) will silently break detection and cause `list issues --severities` to start throwing on legacy servers. Consider having the client expose the HTTP status (e.g. a typed error carrying `status`, or using `getSafe` which already returns the raw `response`) and branch on `response.status === 404` instead of string matching.
> 
> ✅ Edge Case: Cloud users can no longer filter by standard severities
> > 📄 src/cli/commands/list/issues.ts:55-67 📄 src/cli/commands/list/issues.ts:110-111 📄 src/sonarqube/client.ts:259-268
> > `getServerMode()` returns `'mqr'` unconditionally for SonarQube Cloud (`if (this.isCloud) return 'mqr'`). As a result, `parseSeverities` validates Cloud requests against `VALID_MQR_SEVERITIES` (INFO, LOW, MEDIUM, HIGH, BLOCKER) only. Before this PR, the single `VALID_SEVERITIES` list accepted MINOR/MAJOR/CRITICAL for all servers, so existing Cloud scripts that pass e.g. `--severities MAJOR` will now fail with `InvalidOptionError`. This is likely the intended behavior for CLI-674, but it is a breaking change for existing Cloud users and there is no migration/deprecation hint in the error.
> > Consider either accepting both severity vocabularies on Cloud (mapping standard→impact), or at minimum adding a `💡` remediation hint to the `InvalidOptionError` that explains the MQR severity values to use (the `parseSeverities` error currently passes no second hint argument). This eases the transition for scripted callers.
> 
> ✅ Quality: Severity validation duplicated between preflight and parseSeverities
> > 📄 src/cli/commands/list/issues.ts:107-120 📄 src/cli/commands/list/issues.ts:55-67
> > The handler validates `options.severities` twice: an inline preflight block (lines 107-120) checks values against the union of both severity sets, and `parseSeverities` (lines 55-67) re-splits/trims/uppercases and validates again against the mode-specific set. This is two sources of truth for the same parsing logic; future changes (e.g. adding a new severity value or changing trimming behavior) must be kept in sync in both places or validation will drift. Consider consolidating: keep the cheap union preflight only to fail fast before the network call, but factor the split/trim/uppercase into a single shared helper that both the preflight and `parseSeverities` reuse, so the normalization logic exists in exactly one place.
> 
> _...and 1 more resolved from earlier reviews_
> 
> 🤖 Prompt for agents
> ```
> Code Review: Adds impact severity support to the --severities filter, but the implementation lacks robust error handling for mode endpoints, introduces misleading NotFoundError messages, and contains duplicated severity validation logic.
> 
> 1. 💡 Bug: 404 GETs now throw NotFoundError instead of 'Access denied' msg
>    Files: src/sonarqube/client.ts:88-98
> 
>    The change makes every GET that receives HTTP 404 throw the new `NotFoundError` ("Resource not found (404).") instead of the previous generic `Error` with message "Access denied (HTTP 404). Check that the supplied token and organization are valid." This affects all callers of `get()`/`getSafe()`, not just `getServerMode()`.
>    
>    While most existing catch blocks swallow errors generically and remain functionally correct, any caller (or test) that matched on the old "Access denied" message string, or that relied on a 404 producing that user-facing remediation hint, will now surface a different, less actionable message. Verify that no command-level error handling or unit test depends on the old 404 message, and confirm the new generic message is acceptable for end users encountering a 404 (e.g. unknown project/component).
> ```
> 
> Options
> Auto-apply is off → Gitar will not commit updates to this branch.Display: compact → Showing less information.
> 
> Comment with these commands to change:
> 
> Auto-apply	Compact
> ```
> gitar auto-apply:on         
> ```
> 
> ```
> gitar display:verbose         
> ```
> 
> Was this helpful? React with 👍 / 👎 | [Gitar](https://gitar.ai)

 This finding is stale. NotFoundError was introduced and reverted within earlier iterations — it is not present in the final diff. raiseForStatus is unchanged from master and still throws "Access denied (HTTP 404)" for regular GETs. getServerMode uses the pre-existing HTTP_STATUS_NOT_FOUND constant only to branch on the status code from getSafe, which never calls raiseForStatus on its own. Tests confirm: getSafe does not throw on 404 (client.test.ts:401), getServerMode returns 'standard' on 404 (client.test.ts:1529), and raiseForStatus still throws 'Access denied' on 403 (client.test.ts:1118).

PR #457, opened 2026-06-16
CLI-684 Provide feedback on SCA analysis progress

tomshafir-sonarsource:
----
## Summary by Gitar

- **UI & UX improvements:**
  - Added `OutputChannel` type to route terminal output to either `stdout` or `stderr`.
  - Updated `withSpinner`, `text`, `discreetSuccess`, and `print` helpers to support explicit stream selection.
- **Refactor:**
  - Redirected CLI progress indicators (spinners and status text) to `stderr` in `ScaScanOrchestrator` and `manifest-secrets-guard` to keep `stdout` clean for programmatic (JSON) output.
  - Updated binary installers and debug logging to default or route appropriately to `stderr`.
- **Tests:**
  - Added integration test `routes progress to stderr and keeps stdout free of progress noise` to verify stream isolation.
  - Added unit tests for UI stream routing in `messages.test.ts` and `ui-components.test.ts`.

### DIFF
--- a/src/cli/commands/_common/install/binary.ts
+++ b/src/cli/commands/_common/install/binary.ts
@@ -33,7 +33,7 @@ import {
   verifyBinarySignature,
 } from '../../../../lib/sonarsource-releases';
 import { recordInstallationInState } from '../../../../lib/state-manager';
-import { print, text, withSpinner } from '../../../../ui';
+import { type OutputChannel, print, text, withSpinner } from '../../../../ui';
 import {
   cleanupOldVersionBinaries,
   ensureBinDirectory,
@@ -52,7 +52,7 @@ export interface BinarySpec {
 export interface InstallOptions {
   force?: boolean;
   binDir?: string;
-  stream?: NodeJS.WriteStream;
+  channel?: OutputChannel;
 }
 
 export interface InstallResult {
@@ -99,22 +99,22 @@ async function downloadAndInstall(
     return { skipped: true, binaryPath };
   }
 
-  const stream = options.stream ?? process.stdout;
+  const channel = options.channel ?? 'stdout';
 
-  text(`     Installing ${spec.name} ${spec.version}`, undefined, stream);
+  text(`     Installing ${spec.name} ${spec.version}`, undefined, channel);
 
   const downloadUrl = buildDownloadUrl(spec.name, spec.version, spec.distPrefix, platform);
   await withSpinner(
     `Downloading ${spec.name} ${spec.version}`,
     () => downloadBinary(downloadUrl, binaryPath),
-    stream,
+    channel,
   );
 
   try {
     await withSpinner(
       'Verifying signature',
       () => verifyBinarySignature(binaryPath, platform, spec.signatures, spec.publicKey),
-      stream,
+      channel,
     );
   } catch (err) {
     rmSync(binaryPath, { force: true });
@@ -128,9 +128,9 @@ async function downloadAndInstall(
   const installedVersion = await withSpinner(
     'Verifying installation',
     () => verifyInstallation(binaryPath),
-    stream,
+    channel,
   );
-  print(`     ${spec.name} ${installedVersion}`, stream);
+  print(`     ${spec.name} ${installedVersion}`, channel);
 
   recordInstallationInState(spec.name, installedVersion, binaryPath);
   cleanupOldVersionBinaries(resolvedBinDir, spec.name, binaryName);
--- a/src/cli/commands/_common/install/binary.ts
+++ b/src/cli/commands/_common/install/binary.ts
@@ -33,7 +33,7 @@ import {
   verifyBinarySignature,
 } from '../../../../lib/sonarsource-releases';
 import { recordInstallationInState } from '../../../../lib/state-manager';
-import { print, text, withSpinner } from '../../../../ui';
+import { print, withSpinner } from '../../../../ui';
 import {
   cleanupOldVersionBinaries,
   ensureBinDirectory,
@@ -52,6 +52,7 @@ export interface BinarySpec {
 export interface InstallOptions {
   force?: boolean;
   binDir?: string;
+  stream?: NodeJS.WriteStream;
 }
 
 export interface InstallResult {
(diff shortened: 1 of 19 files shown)


src/cli/commands/_common/install/binary.ts
```diff
@@ -98,16 +99,22 @@ async function downloadAndInstall(
     return { skipped: true, binaryPath };
   }
 
-  text(`     Installing ${spec.name} ${spec.version}`);
+  const stream = options.stream ?? process.stdout;
```
georgii-borovinskikh-sonarsource: It would be nice to change all installers to this, but not in this PR I guess

src/cli/commands/_common/install/sca-scanner.ts
```diff
@@ -43,10 +43,12 @@ export const SCA_SCANNER_SPEC: BinarySpec = {
   publicKey: SONARSOURCE_PUBLIC_KEY,
 };
 
-export async function installScaScannerBinary(): Promise<string> {
-  const { binaryPath, freshlyInstalled } = await installBinary(SCA_SCANNER_SPEC);
+export async function installScaScannerBinary(
+  stream: NodeJS.WriteStream = process.stdout,
```
georgii-borovinskikh-sonarsource: Why is there a default value here and why is it stdout?

src/cli/commands/analyze/dependency-risk-helpers/sca-scan-orchestrator.ts
```diff
@@ -44,8 +45,14 @@ export class ScaScanOrchestrator {
   ) {}
 
   async run(auth: ResolvedAuth, projectKey: string): Promise<AnalyzeProjectResponse> {
-    await assertScaAvailable(this.client, auth);
-    const settings = await this.client.getProjectSettings(projectKey);
+    const settings = await withSpinner(
+      'Checking dependency analysis availability',
```
georgii-borovinskikh-sonarsource: I would change this to something like 'synchronizing settings'. Any error after that would be obviously attributable to connection/configuration errors

src/cli/commands/analyze/dependency-risk-helpers/sca-scan-orchestrator.ts
```diff
@@ -44,8 +45,14 @@ export class ScaScanOrchestrator {
   ) {}
 
   async run(auth: ResolvedAuth, projectKey: string): Promise<AnalyzeProjectResponse> {
-    await assertScaAvailable(this.client, auth);
-    const settings = await this.client.getProjectSettings(projectKey);
+    const settings = await withSpinner(
```
georgii-borovinskikh-sonarsource: should be a separate method

src/cli/commands/analyze/dependency-risk-helpers/sca-scan-orchestrator.ts
```diff
@@ -73,6 +80,10 @@ export class ScaScanOrchestrator {
       secretsInstaller: this.secretsInstaller,
     });
 
-    return new ScaScannerRunner(this.installer, this.spawner).run(invocation);
+    return withSpinner(
```
georgii-borovinskikh-sonarsource: introduce a private method for this

src/ui/messages.ts
```diff
@@ -73,16 +73,21 @@ export function success(message: string): void {
...
+  stream: NodeJS.WriteStream = process.stdout,
+): void {
   if (isMockActive()) {
     recordCall('discreetSuccess', message);
     return;
   }
-  if (_formattedOutputMode) {
+  // Only stdout participates in formatted-output buffering; an explicit stderr
+  // stream writes through immediately since stderr does not carry the payload.
+  if (stream === process.stdout && _formattedOutputMode) {
```
georgii-borovinskikh-sonarsource: what is the formatted output mode? can you explain?

src/ui/messages.ts
```diff
@@ -102,17 +107,23 @@ export function error(message: string): void {
 }
 
 // Plain terminal output — human-readable, no semantic icon, optional color
-export function text(message: string, color?: ColorFn): void {
+export function text(
+  message: string,
+  color?: ColorFn,
+  stream: NodeJS.WriteStream = process.stdout,
```
georgii-borovinskikh-sonarsource: overall, I think it would be cleaner if we passed around an enum 'stderr' | 'stdout'  and converted it to streams internally in the classes that directly handle output

georgii-borovinskikh-sonarsource (changes requested):

src/ui/messages.ts
tomshafir-sonarsource: It's a toggle we turn on when a command prints JSON or TOON to `stdout`. While it's on, the UI helpers don't print right away, but instead, buffer their lines. We then embed those lines into the JSON under a `messages` key, so `stdout` stays valid JSON. `stderr` is not buffered because it doesn't carry the payload.

georgii-borovinskikh-sonarsource: I didn't know about this. Should we in general use it for the sca analysis output?

georgii-borovinskikh-sonarsource: Also this feels a bit like what we're doing here with stderr. Can you double check with Clifford?

tomshafir-sonarsource: Progress/spinners go to stderr, stdout stays payload-only. Buffering spinners into the JSON doesn't make sense. The broader "messages" buffering mechanism is out of scope here. Leaving it as-is.

src/cli/commands/analyze/dependency-risk-helpers/manifest-secrets-guard.ts
```diff
@@ -49,8 +50,11 @@ export async function preScanManifestsForSecrets(deps: {
     ...invocation,
     workDir: join(tmpdir(), `sonar-sca-discover-${Date.now()}`),
   };
-  const manifestFiles = await new ScaDiscoverManifestsRunner(scaInstaller, scaSpawner).run(
-    discoverInvocation,
+  await scaInstaller.install(); // up front so its download spinners don't nest inside the discovery spinner
```
georgii-borovinskikh-sonarsource: isn't the nesting good in this case?

georgii-borovinskikh-sonarsource (approved):

PR #458, opened 2026-06-16
CLI-687 Push to history when navigating commands

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Navigation updates:**
  - Introduced `navigateToCommand` to handle hash-based navigation and history state updates consistently.
  - Replaced direct `selectCommand` calls with `navigateToCommand` in event listeners across `commands.html`.


kirill-knize-sonarsource (approved):
LGTM

PR #463, opened 2026-06-16
CLI-623 Add Context Augmentation skill for Cursor

vnaskos-sonar:
Adds SonarQube Context Augmentation (CAG) to `sonar integrate cursor`, delivered as a **native, on-demand skill** shared with the other agent tools.

## What this does
- Writes the rendered CAG skill verbatim to `.agents/skills/sonar-context-augmentation/SKILL.md` — the shared cross-tool skills directory Cursor reads alongside Codex and Antigravity. The three tools therefore share **one** skill (no `.cursor/skills` duplicate), and Cursor loads it on demand, identical to how the other agents receive it. Cursor uses the shared `createContextAugmentationFeature` factory unchanged.
- Wires `resolveContextAugmentationSetup()` + the `--skip-context` gate into `integrate/cursor` and persists the CAG connection attrs, mirroring the other agents. Entitlement-gated, project-scoped; binary download, `tool integrate`, and post-update refresh are all inherited from the factory.

> Two earlier iterations were dropped during review: (1) an `alwaysApply: true` `.cursor/rules/*.mdc` rule — Cursor supports skills natively, the wrapping produced a malformed double-frontmatter file, and CAG should load on demand rather than every turn; (2) a Cursor-private `.cursor/skills` copy — correct format but it duplicated the skill Codex/Antigravity already install in `.agents/skills`.

## Shared install-tail refactor
Extracts the install tail shared by cursor/codex/copilot (CAG resolution → integration-options assembly → recorded attrs → `installIntegration`) into a generic `finalizeAgentInstall<TOptions>()` in a new `_common/agent-integrate-install.ts`, removing the copy-paste SonarCloud flagged. `claude` keeps its own tail (DB-style migrations, try/catch cleanup, custom auth token) and is intentionally not migrated.

## Tests
- Cursor integration specs (MCP / secrets hook / SQAA / CAG / scope / idempotency).
- CAG e2e: `cag-cursor.test.ts` (real-binary skill refresh under `.agents/skills`) and a cursor skip-path case in `cag-integrate.test.ts`.
- All four agents' integration suites pass (the shared prelude + new install helper are touched).

## Notes
- **SQAA** delivery (CLI-621) intentionally stays an `alwaysApply: true` `.cursor/rules/*.mdc` rule — that protocol must run on every file-editing turn, whereas CAG is a consultable skill. Different mechanisms for different needs.


src/cli/commands/integrate/_common/agent-integrate-prelude.ts
```diff
@@ -157,3 +163,63 @@ export function resolveIntegrateInstallTarget(
...
+  featureOptions?: Partial<TOptions>;
+}
+
+/**
+ * Shared install tail for agent integrations: resolves Context Augmentation
+ * (honouring `--skip-context`), assembles the integration options and recorded
+ * state attrs, and runs `installIntegration`. Keeps each agent handler focused
+ * on its agent-specific setup (prompts, scope warnings, SQAA option name).
+ */
+export async function finalizeAgentInstall<TOptions extends IntegrateAgentOptions>(
```
sophio-japharidze-sonarsource: Nice!

I think this file is either not the correct place for this or it needs to be renamed. Because until now it was meant to be a shared PRElude, not really for final steps.

src/cli/commands/integrate/cursor/declaration.ts
```diff
@@ -90,6 +94,13 @@ function buildCursorSqaaRule(context: IntegrationContext): string {
   return `---\nalwaysApply: true\n---\n\n${buildSqaaSectionBody(projectKey)}`;
 }
 
+// Cursor auto-discovers skills under `.cursor/skills/<name>/SKILL.md` and loads
+// them on demand, so Context Augmentation is delivered as a native skill (same
+// rendered SKILL.md the other agents receive), not an always-applied rule.
+function resolveCursorCagSkillPath(context: IntegrationContext): string {
```
sophio-japharidze-sonarsource, line 105: [Cursor docs](https://cursor.com/docs/skills) say that it also reads skills from `.agents/skills`, like Codex & Antigravity do. Do you think we should check that cag skill does not exist in either of the locations before creating it again?

sophio-japharidze-sonarsource (commented):
Looks good. Let's make sure that we don't duplicate CAG skills between `.cursor` & `.agents` folders

src/cli/commands/integrate/_common/agent-integrate-prelude.ts
vnaskos-sonar: Good catch!

src/cli/commands/integrate/cursor/declaration.ts
```diff
@@ -90,6 +97,15 @@ function buildCursorSqaaRule(context: IntegrationContext): string {
...
 }
 
+// Context Augmentation is delivered as a native, on-demand skill (the same
+// rendered SKILL.md the other agents receive), written to the shared
+// `.agents/skills` directory that Cursor auto-discovers — not an always-applied
+// rule, and not a Cursor-private `.cursor/skills` copy that would duplicate the
+// skill Codex/Antigravity already install there.
+function resolveCursorCagSkillPath(context: IntegrationContext): string {
+  return join(context.targetRoot, AGENTS_SKILLS_DIR, CAG_SKILL_NAME, 'SKILL.md');
+}
```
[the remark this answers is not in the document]
vnaskos-sonar: No change needed for this PR. The "no reference-counted teardown for shared .agents/skills skills" is a pre-existing, factory-level limitation affecting Codex/Antigravity equally - if the team wants to fix it (e.g. only rm when no other recorded integration references the shared skill), that belongs in the shared factory as a separate follow-up, not scoped to Cursor.

src/cli/commands/integrate/_common/agent-integrate-postlude.ts
sophio-japharidze-sonarsource, line 1: Suggestion - `agent-integrate-postlude.ts` to easily distinguish and understand its meaning

sophio-japharidze-sonarsource (approved):

PR #479, opened 2026-06-18
CLI-622 Fix Cursor secrets hooks on Windows

vnaskos-sonar:
Three bugs prevented the hooks from blocking secrets on Windows:

1. Missing `exit \$LASTEXITCODE` in the PowerShell hook template — sonar's exit 2 (deny) was silently swallowed, so Cursor always saw exit 0 (allow).

2. Missing `-ExecutionPolicy Bypass` in the powershell invocation — PS1 scripts were blocked on systems with Restricted execution policy.

3. Cursor >=3.8.11 prepends `n++` framing bytes before the JSON hook payload. `JSON.parse("n++{...")` throws, the catch block silently returns, and the hook exits 0. Fixed in `readStdinJson` by stripping any non-JSON prefix before the first `{` or `[`.


tomshafir-sonarsource (approved):
LGTM!

PR #485, opened 2026-06-19
CLI-714 Update git hook messaging for multi scan type

georgii-borovinskikh-sonarsource:
Part of CLI-505

----
## Summary by Gitar

- **CLI messaging improvements:**
  - Renamed all Git hooks to `code scanning hook` in logs, prompts, and documentation to reflect multi-scan capabilities.
  - Updated `integrate git` description to explicitly mention both secrets and dependency risks scanning.
- **CLI output enhancements:**
  - Added introductory messages to clarify that hooks cover both scan types and that availability may vary.

### DIFF
--- a/src/cli/commands/integrate/git/index.ts
+++ b/src/cli/commands/integrate/git/index.ts
@@ -27,7 +27,7 @@ import { GLOBAL_HOOKS_DIR } from '../../../../lib/config-constants';
 import { normalizePath } from '../../../../lib/fs-utils';
 import { findGitRoot } from '../../../../lib/project-workspace';
 import { blank, confirmPrompt, info, intro, text, warn } from '../../../../ui';
-import { yellow } from '../../../../ui/colors.ts';
+import { yellow } from '../../../../ui/colors.js';
 import { CommandFailedError, InvalidOptionError } from '../../_common/error';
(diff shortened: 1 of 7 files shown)


src/cli/commands/integrate/git/index.ts
```diff
@@ -140,7 +141,9 @@ export async function integrateGit(options: IntegrateGitOptions): Promise<void>
     throw new InvalidOptionError('--dependency-risks and -p are not supported with --global.');
   }
 
-  intro('SonarQube Git Integration (secrets scanning)');
+  intro('SonarQube Git Integration (source code scanning)');
+  info('This integration includes secrets and dependency risks detection in your git repository.');
+  info(yellow('Some scan types may be unavailable for certain hook types.'));
```
tomshafir-sonarsource, line 146: Should we just use `warn` here for the yellow text?

tomshafir-sonarsource (approved):
Approved with 1 small comment. Also check Gitar's comment before merging.

src/cli/commands/integrate/git/index.ts
georgii-borovinskikh-sonarsource: warn doesn't work - it's printed to stderr and doesn't follow the same indentation

PR #497, opened 2026-06-22
CLI-727 Telemetry foundation: timed() utility and CliAnalysisFindingDetected types

vnaskos-sonar:
## What

Purely additive foundation that the subsequent CLI-728, CLI-730, CLI-731, and CLI-733 PRs depend on. No behaviour changes.

- **`src/lib/timed.ts`** — generic `timed<T>(fn: () => Promise<T>)` utility that wraps any async call and returns its result alongside the wall-clock duration in milliseconds. Used by all three analyzer call sites (sonar-secrets, SQAA, SCA) to measure scan duration without changing existing function signatures.
- **`src/lib/state.ts`** — two new types: `AnalysisFindingEventPayload` (the four new fields `caller_command`, `analyzer`, `rule_key`, `scan_duration_ms` plus all shared identity/connection fields from `CliCommandExecuted`) and `StoredFindingEvent` (metadata envelope + payload, mirrors `StoredTelemetryEvent`).
- **`src/lib/config-constants.ts`** — `TELEMETRY_DIR` constant pointing to `~/.sonar/sonarqube-cli/telemetry/`, the directory for telemetry artefacts not stored in state.json.

## Tests

- `timed()` returns the correct value and a non-negative `durationMs`
- `timed()` propagates errors from the wrapped function unchanged
- `durationMs` reflects actual elapsed time (wraps a real 50 ms sleep and asserts the result lands within tolerance — catches implementations that always return 0)

## Dependencies

None. Must land before CLI-728 (sink + flush), CLI-730 (secrets call sites), CLI-731 (SQAA call sites), and CLI-733 (SCA call sites).

----
## Summary by Gitar

- **Build and test configuration:**
  - Updated `bunfig.toml` to include `tests/coverage/preload-instrumenter.ts` in the test preload list.
  - Simplified `package.json` test scripts by removing explicit preload arguments, now handled globally by `bunfig.toml`.

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -3,4 +3,7 @@ exact = true
 registry = { url = "https://repox.jfrog.io/artifactory/api/npm/npm/", token = "$ARTIFACTORY_PRIVATE_READER_PASSWORD" }
 
 [test]
-preload = ["./tests/_common/preload-isolated-env.ts"]
+preload = [
+  "./tests/coverage/preload-instrumenter.ts",
+  "./tests/_common/preload-isolated-env.ts",
+]
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "test:integration": "bun test ./tests/integration/",
     "test:all": "bun run test:unit && bun run test:integration",
     "test:e2e": "bun test ./tests/e2e/",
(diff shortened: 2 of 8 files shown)


nquinquenel (approved):
LGTM. I wonder if moving the preload to any test will add an overhead for unit testing, I guess it's negligible 🤔

PR #503, opened 2026-06-24
CLI-729 Switch all sonar-secrets call sites to --json output and bump binary to 2.45.0.11557

vnaskos-sonar:
sonar-secrets 2.45.0 (SONARTEXT-1007) introduced structured JSON output. Switching to it unlocks `ruleKey` per finding (consumed by CLI-557 / CLI-730 telemetry) and removes the fragile plain-text parser (`parseSecretsOutput`).

The binary's `errors[]` field is treated as advisory: non-empty entries surface as scan warnings — in text output and in the combined `analyze --format json` report as `secrets.warnings[]` — and never abort the scan. A scan error in that channel is not a secrets finding, so failing hard would produce false negatives in CI.

`sonar analyze secrets` output is redesigned to match the agentic display style: a green `✓` row per scanned file with `No issues found · Nms` on clean runs, and grouped findings with `N secret(s) found · Nms` on detections.

Note for release: `sonar analyze --format json` consumers will see renamed fields in `secrets.issues[]` (`message`→`description`, `secret`→`maskedSecret`, flat offsets→nested `location` object). This is a breaking change in that output shape.

The `2.45.0.11557` binary is published on binaries.sonarsource.com, so this is mergeable.


tests/e2e/secrets-output-parser.test.ts
```diff
@@ -117,11 +120,11 @@ describe.skipIf(binaryPath === null)('parseSecretsOutput — real binary', () =>
...
 
-    const issues = parseSecretsOutput(result.stdout);
+    const { issues } = parseSecretsJson(result.stdout);
 
     expect(issues.length).toBeGreaterThanOrEqual(2);
     const files = issues.map((i) => i.file);
-    expect(files.some((f) => f.includes('file1.ts'))).toBe(true);
-    expect(files.some((f) => f.includes('file2.ts'))).toBe(true);
+    expect(files.some((f) => f?.includes('file1.ts'))).toBe(true);
+    expect(files.some((f) => f?.includes('file2.ts'))).toBe(true);
```
nquinquenel, line 128: Could we have an E2E test where we actually test the full JSON output? I think this is crucial as we now rely on a very specific output provided by a third-party binary. We need to ensure compatibility, to avoid unexpected breakage when we update the version.

tests/e2e/secrets-output-parser.test.ts
```diff
@@ -117,11 +120,40 @@ describe.skipIf(binaryPath === null)('parseSecretsOutput — real binary', () =>
...
+    expect(typeof issue['description']).toBe('string');
+    expect(typeof issue['file']).toBe('string');
+    expect(typeof issue['maskedSecret']).toBe('string');
+
+    expect(typeof issue['location']).toBe('object');
+    const loc = issue['location'] as Record<string, unknown>;
+    expect(typeof loc['startLine']).toBe('number');
+    expect(typeof loc['startColumn']).toBe('number');
+    expect(typeof loc['endLine']).toBe('number');
+    expect(typeof loc['endColumn']).toBe('number');
```
nquinquenel: I think it'd be great if we could validate the full JSON schema, I think we use this pattern elsewhere, Cursor gave me this example:
```
const raw = JSON.parse(result.stdout);

expect(raw).toEqual({
  issues: [
    {
      ruleKey: expect.stringMatching(/^secrets:/),
      description: expect.any(String),
      file: expect.stringContaining('schema-check.ts'),
      maskedSecret: expect.any(String),
      location: {
        startLine: expect.any(Number),
        startColumn: expect.any(Number),
        endLine: expect.any(Number),
        endColumn: expect.any(Number),
      },
    },
  ],
  // omit errors when absent, or:
  // errors: expect.arrayContaining([expect.any(String)]),
});
```
IMO that's really high-value test, to ensure we're in control of what we output. If something changes we should be aware, and avoid any failure later

nquinquenel (approved):
LGTM, just one comment that is quite important IMO

PR #504, opened 2026-06-24
CLI-664 Agent integrations for multi-file DEEP analysis

nquinquenel:
----
## Summary by Gitar

- **CLI command enhancements:**
  - Added repeatable `--file` argument support and a `--depth` option (`STANDARD` or `DEEP`) to `analyze agentic` and `analyze all`.
  - Updated command parsing logic to support multiple files and granular analysis depth configuration.
- **Analysis engine & Reporting:**
  - Implemented multi-file analysis support with `resolveSqaaFileArgs` and depth-aware API calls.
  - Refactored JSON report generation into `sqaa-display-json.ts` and updated output to include `analysisDepth`.
- **Agentic integration:**
  - Added `installSqaaInstructions` to inject mandatory end-of-turn `DEEP` analysis protocols into `CLAUDE.md`.
  - Updated agent prompts to prioritize single-invocation, multi-file `DEEP` analysis commands.
- **Testing:**
  - Added unit tests for path validation, depth resolution, and JSON report generation, plus integration tests for multi-file `DEEP` analysis.


src/cli/commands/analyze/analyze-all.ts
```diff
@@ -34,14 +34,16 @@ import type { OutputFormat } from './sqaa';
...
 
 export interface AnalyzeAllOptions {
-  file?: string;
+  file?: string[];
   staged?: boolean;
   base?: string;
   project?: string;
   force?: boolean;
   format?: OutputFormat;
+  depth?: string;
```
damien-urruty-sonarsource, line 46: Do we really need to expose such low-level and technical options to the user?

src/cli/command-tree.ts
```diff
@@ -380,18 +382,28 @@ const sqaaFormatOption = new Option('--format <format>', 'Output format')
   .choices(SQAA_FORMATS)
   .default('text');
 
+const sqaaDepthOption = new Option(
+  '--depth <depth>',
+  `Analysis depth: ${SQAA_DEPTH_CHOICES.join(' (fast) or ')} (cross-file). Defaults depend on scope.`,
```
damien-urruty-sonarsource: This dynamic printing of possible options looks strange, because you expect values to be in a specific order. Also what does 'depend on scope' mean in this context? Is is change set vs file list? We could maybe be a little bit more precise

damien-urruty-sonarsource (approved):
LGTM, but the PR is big (+1.2K lines), with many different changes. This impacts the depth (no pun intended) of my review, because there are just too many details to focus on. It could have been split into 3-4 PRs IMO (managing depth, multi-file option support, ...). 

I think we should discuss in the team what is an acceptable size for a PR. A quick search suggests around 400 lines. I think it's even more important these days, with agents able to generate huge amounts of code

src/cli/commands/analyze/analyze-all.ts
nquinquenel: Yes, this has been discussed with PM and we want users to be able to use `depth`, for example if they want to script something

nquinquenel: Today `depth` has 2 options, but later we may have more. The idea is that this option is decided on the type of workflow, not necessarily the command, so it makes sense on that regard to let user decide

src/cli/command-tree.ts
nquinquenel: In a recent refactoring I had to move the `SQAA_DEPTH_CHOICES` out of this class, and I wasn't sure I wanted to keep hardcoded values that could drift from actual implementation. But I also agree it's not ideal, I actual spent time thinking about this line 🤦🏻 

I can change it back to something static:
```
const sqaaDepthOption = new Option(
  '--depth <depth>',
  'Analysis depth: STANDARD (fast) or DEEP (cross-file). Default: STANDARD for one --file; DEEP otherwise.',
).choices(SQAA_DEPTH_CHOICES);
```

nquinquenel:
> the PR is big (+1.2K lines), with many different changes. This impacts the depth (no pun intended) of my review, because there are just too many details to focus on. It could have been split into 3-4 PRs IMO (managing depth, multi-file option support, ...).
> 
> I think we should discuss in the team what is an acceptable size for a PR. A quick search suggests around 400 lines. I think it's even more important these days, with agents able to generate huge amounts of code

I agree, it's probably not ideal here and I should be more mindful about it.
It was kinda chaotic as this PR started by another one (https://github.com/SonarSource/sonarqube-cli/pull/471) that included more changes, and which in the end got refactored multiple times, reduced in scope, and then we added a few new last-minute features 😅

PR #515, opened 2026-06-25
CLI-773 Pull latest version from binaries.sonarsource.com from install scripts

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Refactored update mechanism:**
  - Replaced version extraction from install scripts with direct `stable.version` metadata fetching from `binaries.sonarsource.com`.
  - Introduced a `Version` helper class for cleaner version comparison and build-number stripping.
- **New CLI commands and structure:**
  - Added `src/cli/commands/self-update/index.ts` to manage update logic and execution.
  - Centralized install script fetching and update logic in `src/cli/commands/self-update/update-check.ts`.
- **Updated install scripts:**
  - Modified `install.sh` and `install.ps1` to pull the latest version from `stable.version` instead of a static file or hardcoded string.
  - Retained legacy compatibility markers for older `self-update` clients, updated by release automation.
- **Testing improvements:**
  - Enhanced `FakeBinariesServer` to support serving the new `stable.version` metadata.
  - Updated integration and unit tests to validate the new version-checking flow and error handling.


nquinquenel (commented):
I think there's a useless step in `build.yml` now:

`- name: Generate latest-version.txt`

Also I think the self-update may breaks for all old CLI after the merge, because we were reading the install script from master 🤔

damien-urruty-sonarsource:
@nquinquenel that's a great catch! I think this `latest-version.txt` was just dead code.

> the self-update may breaks for all old CLI after the merge

It shouldn't, why do you think so?

damien-urruty-sonarsource:
@nquinquenel you were right, that was a pretty bad bug, so great catch! Thanks.

I pushed a change to reinstate the version in the install scripts, and the automation that goes with it. The version is not used from the install script itself, it only serves for previous CLI versions and the self-update mechanism. We will be able to remove it in a while. Please re-review

nquinquenel (approved):
LGTM

PR #527, opened 2026-06-30
CLI-781 Enforce network config for self-update.

Krosovok:
During the implementation the self-update command was changed in parallel with Proxy/SSL addition, so it’s missing the config enforcement.


damien-urruty-sonarsource (approved):
Good catch, thanks! I am bit worried that this will happen again. Shouldn't we introduce our own centralized 'fetch'? And try to enforce its usage everywhere?

Krosovok:
> Good catch, thanks! I am bit worried that this will happen again. Shouldn't we introduce our own centralized 'fetch'? And try to enforce its usage everywhere?

Yes, it is a concern. I didn't go out of my way to add it and we have fetch-guarded.ts with some common functionality (even though it doesn't seem applicable to every case).

Maybe we can have some brainstorm on this topic and include it with part 2 of the Proxy+SSL epic.

damien-urruty-sonarsource:
One idea to enforce it:

image

PR #528, opened 2026-06-30
CLI-782 Two-event analysis telemetry schema and ndjson sink

nquinquenel:


src/telemetry/findings.ts
```diff
@@ -96,6 +79,70 @@ export function emitSecretsFindings(
     sqs_installation_id: conn?.sqsInstallationId ?? null,
     caller_agent: detectCallerAgent(),
   };
+}
+
+/**
+ * Appends one CliAnalysisCompleted event to findings.ndjson. Fire-and-forget.
+ */
+export function appendAnalysisCompleted(payload: AnalysisCompletedEventPayload): void {
```
vnaskos-sonar: These two functions (`appendAnalysisCompleted` and `appendAnalysisFindingsDetected`) write unconditionally — if the user has opted out of telemetry, events still get written if a caller skips `buildAnalysisIdentityBase`. Worth either moving the `isTelemetryEnabled` check inside here, or going with a self-contained `emitAnalysisCompleted` function (like `emitSecretsFindings` was) that handles gating, identity resolution, and UUID minting in one place so it is impossible to bypass.

src/lib/state.ts
```diff
@@ -469,29 +466,96 @@ export interface AnalysisFindingEventPayload {
...
+  details_schema_version: number;
+  /** JSON-encoded allowlist blob (rule keys and per-rule counts only) */
+  details: string;
+}
+
+/**
+ * Payload for a CliAnalysisFindingDetected event — one event per finding
+ * produced by sonar-secrets, SQAA, or SCA (legacy per-issue schema).
+ */
+export interface AnalysisFindingEventPayload extends AnalysisEventIdentityPayload {
```
vnaskos-sonar: Since we have not emitted any `CliAnalysisFindingDetected` events to production yet, I think we can fully remove this along with `StoredFindingEvent`, `appendFinding`, and `emitSecretsFindings` — no migration needed. The new two-event schema (`CliAnalysisCompleted` + `CliAnalysisFindingsDetected`) is the replacement for all three analyzers.

vnaskos-sonar (commented):
Good work on this — the schema types are well-structured and the ndjson sink wiring is clean. One broader note: since CLI-731 specifically describes emitting the events at the SQAA call sites (`analyze agentic`, `agent-post-tool-use`, `codex-post-tool-use`), this PR covers the foundation but not the instrumentation. It might be worth tracking the actual emit work under a new ticket so the scope stays aligned with what's merged. Either way, this is a solid foundation and the right step before the instrumentation tickets.

src/lib/state.ts
```diff
@@ -469,29 +466,96 @@ export interface AnalysisFindingEventPayload {
...
+export type AnalysisRunStatus = 'clean' | 'findings' | 'error';
+
+export type AnalysisTelemetryAnalyzer = 'sonar-secrets' | 'sqaa' | 'sca-scanner-cli';
+
+/**
+ * Payload for a CliAnalysisCompleted event — one event per analyzer run
+ * (clean, findings, or error).
+ */
+export interface AnalysisCompletedEventPayload extends AnalysisEventIdentityPayload {
+  /** Literal CLI subcommand name (e.g. "analyze-agentic", "git-pre-commit") */
```
vnaskos-sonar: Minor: the `caller_command` example uses a hyphen (`analyze-agentic`) but the agreed format is a space (`analyze agentic`). Same fix needed in the test fixture `makeCompletedPayload`.

src/telemetry/findings.ts
nquinquenel: Modified to that both resolve identity via `buildAnalysisIdentityBase()` that gates for telemetry enabled

vnaskos-sonar (commented):
LGTM!

vnaskos-sonar (approved):
🚀

PR #534, opened 2026-07-01
GROW-207 Add hidden sonar import command with org selection

kevinmlsilva:
## Summary
- Adds the hidden `sonar import` command (Step 1 of MMF-5732), currently scoped to resolving a SonarQube Cloud organization.
- Filters orgs to those where the user is admin and has a DevOps platform bound (`alm`), always prompting the user to choose even with one eligible org, with local pagination (10/page + "Load more...").
- Supports `--org <key>` to bypass the prompt and `--non-interactive` (requires `--org`, exits 2 otherwise); adds `fetchAllUserOrganizations()` and a paginated `OrganizationCollection` helper on `SonarQubeClient`.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run test:unit`
- [x] `bun test tests/integration/specs/import/import.test.ts` (11/11 passing)

----
## Summary by Gitar

- **Fixed telemetry:**
  - Awaited async resolution and emission in `scanAndEmitSecrets` to ensure reliable telemetry capture.


src/sonarqube/client.ts
```diff
@@ -55,6 +55,63 @@ export type CagEntitlementStatus = 'allowed' | 'not_allowed' | 'check_failed';
...
+  alm?: { key: string };
+  actions?: { admin: boolean };
+}
+
+const ORGANIZATION_LOCAL_PAGE_SIZE = 10;
+
+export class OrganizationCollection {
+  private readonly orgs: Organization[];
+  private readonly pageSize: number;
+  private page: number;
```
nquinquenel: I'd move this to the `_common` folder for example, it's about prompt/UX logic and should not be in the client class

src/cli/commands/import/_common/resolve-options.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export async function resolveOrg(
+  client: SonarQubeClient,
+  opts: { org?: string; nonInteractive?: boolean },
+): Promise<string> {
+  if (!client.isCloud) {
+    throw new CommandFailedError('sonar import is only supported on SonarQube Cloud.');
+  }
+
+  if (opts.org) return opts.org;
```
nquinquenel, line 36: Is it expected? There's no check at all (org without ALM, admin, etc)

src/cli/commands/import/_common/resolve-options.ts
```diff
@@ -0,0 +1,81 @@
...
+import { selectPrompt, withSpinner } from '../../../../ui';
+import { CommandFailedError, InvalidOptionError } from '../../_common/error';
+
+export async function resolveOrg(
+  client: SonarQubeClient,
+  opts: { org?: string; nonInteractive?: boolean },
+): Promise<string> {
+  if (!client.isCloud) {
+    throw new CommandFailedError('sonar import is only supported on SonarQube Cloud.');
+  }
```
nquinquenel, line 34: Do you think we can add a `remediationHint`? On a similar command we have:

`remediationHint: "Run 'sonar auth login' and connect to SonarQube Cloud, then retry.",`

src/sonarqube/client.ts
```diff
@@ -464,21 +521,44 @@ export class SonarQubeClient {
...
     try {
-      const result = await this.get<{
-        organizations: Array<{ key: string; name: string }>;
-        paging: { total: number };
-      }>('/api/organizations/search', { member: true, ps: 10 });
-      return { organizations: result.organizations, total: result.paging.total };
+      return await this.fetchUserOrganizationsPage(page, ps);
     } catch {
       return { organizations: [], total: 0 };
     }
```
nquinquenel, line 482: Should we have a try/catch here? I wonder what happens if it fails during the command execution, what's the UX/display message? `fetchAllUserOrganizations` below would return an actual error (exit 1) with a message for example

tests/integration/specs/import/import.test.ts
```diff
@@ -0,0 +1,291 @@
...
+  describe('organization selection', () => {
+    it(
+      'prompts to select org even when only one eligible org exists',
+      async () => {
+        const server = await harness
+          .newFakeServer()
+          .withAuthToken('test-token')
+          .withOrganizations([
+            { key: 'my-org', name: 'My Organization', alm: GITHUB_ALM, actions: ADMIN_ACTIONS },
+          ] as unknown as Array<{ key: string; name: string }>)
```
nquinquenel: That's a kinda ugly cast 😄 

Do you think we could extend the fake server type to support `alm` and `actions`?

nquinquenel (commented):
Look good! Some minor comments

src/sonarqube/client.ts
kevinmlsilva: Good call — moved `OrganizationCollection` out of `client.ts` into `src/cli/commands/import/_common/organization-collection.ts`. The API client now just returns `Organization[]` from `fetchAllUserOrganizations()`; pagination/filtering for the select prompt lives in the command layer.

src/cli/commands/import/_common/resolve-options.ts
kevinmlsilva: It's intentional for now: this is Step 1 of MMF-5732 and `--org` is an explicit override that just sets the value used later in `importHandler` (currently only printed). Since no import/API call is wired up yet in this step, there's nothing downstream to validate against. Once the actual import call lands, we'll want the org resolved this way to hit the same server-side validation as the discovered path, so I'd rather not bake a client-side re-check in here that we'd have to keep in sync.

src/cli/commands/import/_common/resolve-options.ts
kevinmlsilva: Done — added a `remediationHint: "Run 'sonar auth login' and connect to SonarQube Cloud, then retry."` to that error.

src/sonarqube/client.ts
kevinmlsilva: Fixed — wrapped the fetch in `resolveOrg` in a try/catch that rethrows as `CommandFailedError('Failed to load organizations: <cause>', { remediationHint: 'Check your network connection and authentication, then retry.' })`. Before this, a mid-fetch failure would've bubbled as a bare Error (still exit 1, but a generic/unstyled message with no remediation hint).

tests/integration/specs/import/import.test.ts
kevinmlsilva: Done — `fake-sonarqube-server.ts` now imports the `Organization` type from `src/sonarqube/client.ts` and `withOrganizations`/`memberOrganizations` are typed as `Organization[]`. Removed the `as unknown as` casts from all 5 spots in the test file.

src/cli/commands/import/_common/resolve-options.ts
```diff
@@ -0,0 +1,92 @@
...
+      value: org.key,
+      label: `${org.name} (${org.key})`,
+    }));
+
+    if (eligible.hasMore) {
+      options.push({ value: LOAD_MORE, label: 'Load more...' });
+    }
+
+    const choice = await selectPrompt('Select an organization', options);
+
```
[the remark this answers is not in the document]
kevinmlsilva: Fixed — swapped the string sentinel for `Symbol('load-more')`, so it can never collide with a real org key value-wise (unlike a string, which would still need this exact-key edge case handled). Also added a regression test with an org literally keyed `__load_more__` to prove it's selectable.

nquinquenel (approved):
LGTM!

PR #541, opened 2026-07-02
CLI-187 notify users when new version is available

sophio-japharidze-sonarsource:
----
## Summary by Gitar

- **New feature:**
  - Added background check and user notification when a newer CLI version is available.
  - Implemented daily throttling for update checks and notifications via persistent state.
- **Infrastructure:**
  - Added `maybeNotifyUpdateAvailable` hook to post-command execution flow.
  - Configured command-specific opt-ins for update notifications with optional conditional logic.
  - Suppressed notifications automatically in CI, non-interactive, or machine-readable (JSON) output modes.
- **Testing:**
  - Added comprehensive unit tests in `update-notification.test.ts` covering notification logic, suppression, and state persistence.

### DIFF
--- a/src/lib/update-notification.ts
+++ b/src/lib/update-notification.ts
@@ -1,0 +1,199 @@
+/*
+ * SonarQube CLI
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


src/lib/update-notification.ts
```diff
@@ -0,0 +1,199 @@
...
+    // when the previous check failed) instead of hitting the network again.
+    return {
+      latestVersion: updateCheck?.latestVersion,
+      updateCheck: toPersistedUpdateCheck(updateCheck),
+    };
+  }
+
+  const lastCheckedAt = new Date().toISOString();
+  try {
+    const latestVersion = await fetchLatestVersion();
```
kirill-knize-sonarsource: `fetchLatestVersion()` was created for the `self-update`, it's synchronous and has timeout 5000ms. So we introduce a UX degradation for every command that uses this code, because it will wait up to 5 sec. My suggestion is to parametrize and use shorter timeout for this new flow.

kirill-knize-sonarsource (approved):
LGTM

PR #547, opened 2026-07-02
CLI-817 better support for git worktrees during integration

sophio-japharidze-sonarsource:
I have tested the following scenarios:

* Integrate from the main worktree
   * run `sonar context guidelines get` and `sonar analyze agentic` from the main worktree ✅ 
   * run `sonar context guidelines get` and `sonar analyze agentic` from a different worktree ✅ 
* Integrate from a different worktree
   * run `sonar context guidelines get` and `sonar analyze agentic` from the main worktree ✅ 
   * run `sonar context guidelines get` and `sonar analyze agentic` from a different worktree ✅ 
   
This all works in conjunction with a corresponding [change on the CAG binary](https://github.com/SonarSource/sonar-context-augmentation/pull/320)

----
## Summary by Gitar

- **Git Worktree Support:**
  - Added `lib/project-workspace/git-worktree.ts` to map worktrees to the main repository root and resolve candidate paths for state lookups.
  - Updated `resolveSqaaProjectKey` to search across worktree roots, allowing project metadata to be matched even when integrated in a different worktree.
- **State Persistence:**
  - Modified integration flows to persist `repoRoot` in feature attributes, decoupling integration state from physical `targetRoot` for improved cross-worktree stability.
- **Context Augmentation:**
  - Added `SONAR_CONTEXT_WORKSPACE_ROOT` to `ContextAugmentationEnvContext` to provide CAG with the correct git root or target directory for daemon folder placement.
- **Testing:**
  - Added comprehensive integration tests in `passthrough.test.ts` and `analyze-sqaa.test.ts` to verify resolution from linked worktrees and subdirectories.
- **Documentation:**
  - Updated `CLAUDE.md` with detailed logic for worktree-aware resolution in SQAA and Context Augmentation.


michael-jabbour-sonarsource:
Hi @sophio-japharidze-sonarsource,

FYI, I implemented a draft of the necessary changes on our end at SonarSource/sonar-context-augmentation#320.

It should be possible to download the binaries from  https://github.com/SonarSource/sonar-context-augmentation/actions/runs/28927833586

Update: Fixed Windows binaries at https://github.com/SonarSource/sonar-context-augmentation/actions/runs/28939880034

One thing that I noticed while testing locally (by just replacing the cag binary under `~/.sonar/sonarqube-cli/bin/...` with the custom one above) that the daemon started from a non-integrated worktree is wrongly associated with the main worktree instead. My interpretation would be that `SONAR_CONTEXT_WORKSPACE_ROOT` is incorrectly set to the main work tree in that case (I think that `SONAR_CONTEXT_WORKSPACE_ROOT` should instead point to the worktree where the command is invoked, using the project information i.e. `SONAR_CONTECT_PROJECT` from the main/integrated work tree). Here is some analysis from Codex (I haven't checked the code in this PR myself yet):

> In the current branch, for newly recorded Context Augmentation state, `SONAR_CONTEXT_WORKSPACE_ROOT` corresponds to the **main worktree path**, not the linked worktree where `sonar context` is invoked.
> 
> Flow:
> 
> 1. During `sonar integrate`, CAG attrs record:
> 
> `repoRoot: (await resolveMainWorktreeRoot(projectRoot)) ?? projectRoot`
> 
> See [context-augmentation.ts](/home/michael.jabbour/Documents/repos/work-1/sonarqube-cli/src/cli/commands/integrate/_common/context-augmentation.ts:52) and [git-worktree.ts](/home/michael.jabbour/Documents/repos/work-1/sonarqube-cli/src/lib/project-workspace/git-worktree.ts:99). `resolveMainWorktreeRoot()` uses the first entry from `git worktree list --porcelain`, which the code treats as the main working tree.
> 
> 2. During `sonar context`, the current cwd is used only to find a matching recorded integration. If invoked from a linked worktree, it maps cwd to the equivalent path under the main worktree, then matches state against recorded `repoRoot`.
> 
> See [context/index.ts](/home/michael.jabbour/Documents/repos/work-1/sonarqube-cli/src/cli/commands/context/index.ts:119).
> 
> 3. The env value is then set from:
> 
> `workspaceDir: getOptionalStringAttr(match.attrs, 'repoRoot') ?? match.targetRoot`
> 
> See [context/index.ts](/home/michael.jabbour/Documents/repos/work-1/sonarqube-cli/src/cli/commands/context/index.ts:151) and [context-augmentation-env.ts](/home/michael.jabbour/Documents/repos/work-1/sonarqube-cli/src/cli/commands/_common/context-augmentation-env.ts:66).
> 
> So: **main worktree for new state; `targetRoot` only as fallback for older/non-git state without `repoRoot`.** The invocation worktree path is not passed as `SONAR_CONTEXT_WORKSPACE_ROOT`; it is just part of the lookup/matching process.

Is this intended? Can we change so that `SONAR_CONTEXT_ROOT` points to root of the workspace path where the CLI is invoked?

I am also happy to share any additional information as needed.

Thanks for taking care of this!

sophio-japharidze-sonarsource:
@michael-jabbour-sonarsource thanks. I have ran manual tests and all seems to be good. The only thing worth mentioning is that the output seems too verbose to me. Right now each context command will log
```Auto-starting daemon for '/Users/.../sonarqube-cli'...
Daemon process started (PID 54227). Waiting for readiness...
Daemon is ready.
```

michael-jabbour-sonarsource:
> @michael-jabbour-sonarsource thanks. I have ran manual tests and all seems to be good. The only thing worth mentioning is that the output seems too verbose to me. Right now each context command will log

I just saw this. I wasn't expecting this on every context call; Only the first one per-workspaceroot should print this. Are you observing these on every command even on the same workspace?

sophio-japharidze-sonarsource:
No you are right, it's on the first call per workspace

src/cli/commands/_common/context-augmentation-env.ts
```diff
@@ -26,13 +26,21 @@ export interface ContextAugmentationEnvContext {
...
+  workspaceDir?: string;
 }
 
 type ContextAugmentationEnvKey =
   | 'SONAR_CONTEXT_ORGANIZATION'
   | 'SONAR_CONTEXT_PROJECT'
   | 'SONAR_CONTEXT_TOKEN'
-  | 'SONAR_CONTEXT_URL';
+  | 'SONAR_CONTEXT_URL'
+  | 'SONAR_CONTEXT_WORKSPACE_ROOT';
```
vnaskos-sonar, line 43: I noticed that env variable is still in a [Draft PR](https://github.com/SonarSource/sonar-context-augmentation/pull/320) on the CAG repo, so it probably doesn't have any effect at the moment.

This is also interesting because the integration tests didn't catch it, but I am not sure if we could have a deeper integration test 🤔

sophio-japharidze-sonarsource: This is expected, the changes come in conjunction with the CAG binary update.

Screenshot 2026-07-14 at 11 51 01

Without the changes on CAG binary, a call to CAG will still fail with the above error. CAG change is needed to take into account the new env variable, and auto-create a folder for daemon if one does not exist yet.

src/lib/project-workspace/git-worktree.ts
michael-jabbour-sonarsource, line 1: Why is this seen as a binary file here?

src/cli/commands/integrate/_common/context-augmentation.ts
michael-jabbour-sonarsource, line 191:
```suggestion
    workspaceDir: p.projectRoot,
```
Can we add workspaceDir as well? Currently, for simplicity, I have set all the environment variables as strictly required by all functionalities in CAG. I would also suggest making `assignContextEnvValue` fail early in case any of the fields inside "context" is missing, so that this requirement is encoded on sonarqube-cli side as well...
WDYT?

michael-jabbour-sonarsource (changes requested):
When I try to integrate using the CAG binary from my PR (SonarSource/sonar-context-augmentation#320), the process fails:

```
└────────────────────────────────────────────────────────────────────────────────┘
  ›  Press Enter to install…

     sonar-secrets binary already installed
     sonar-context-augmentation binary already installed
     Installing secret scanning hooks...
     Claude PreToolUse hook script already installed
     Claude UserPromptSubmit hook script already installed
     Claude hooks configuration already installed
     Installing Vortex agentic analysis hook...
     Claude PostToolUse hook script already installed
     Claude Vortex agentic analysis hook configuration already installed
     Installing Vortex agentic analysis instructions...
     Claude Vortex agentic analysis instructions already installed
     Installing MCP server...
     Claude MCP configuration already installed
     Installing context augmentation...
     Installing Vortex context augmentation...
  ✗  sonar-context-augmentation 0.17.0.2614
⚠️ sonar-context-augmentation exited with code 1.
  error[invalid_args]: Required environment variable(s) not set: SONAR_CONTEXT_WORKSPACE_ROOT
    hint: Check command flags and retry.

❌ Vortex context augmentation tool integration failed.
```

See my second commit. Let me know your thoughts.

src/lib/project-workspace/git-worktree.ts
sophio-japharidze-sonarsource: ah, interesting ... I have no idea 🤔

michael-jabbour-sonarsource: Apparently it comes from line 47, where the string `cacheKey` contains real NUL characters. Using something like `\0` instead should solve the problem as far as I understand (I am not that knowledgeable in TS though, so I am not sure)

src/cli/commands/integrate/_common/context-augmentation.ts
michael-jabbour-sonarsource: I can see that my suggested change is part of 1b6b2f77af0db11945626dac243479949437e31d, thanks for landing this. I did a quick test and it seems to work as expected :tada:

I am leaving this conversation open because I would like to hear your thoughts on my suggestion regarding `assignContextEnvValue` before closing.

src/cli/commands/analyze/sqaa-auth.ts
```diff
@@ -110,29 +125,36 @@ export function resolveCloudAuth(
...
+    const lookupRoots = (await resolveWorktreeRootCandidates(projectRoot ?? process.cwd())).map(
+      canonicalProjectRoot,
     );
+    let sqaaFeature;
+    for (const root of lookupRoots) {
+      sqaaFeature = claude?.features.find(
+        (feature) =>
+          feature.featureId === SQAA_HOOK_FEATURE_ID &&
+          feature.scope === 'project' &&
+          sqaaFeatureMatchesRoot(feature, root),
```
michael-jabbour-sonarsource: Nitpick: Depending on the order of features, this may overlook a matching targetRoot in favor of a repoRoot-only match, which could end up yielding the wrong project-key (in the corner case of two workspaces being integrated with different project-keys - could be something that monorepo users do).

I would suggest something like this instead, to prefer `targetRoot` matches here, and fallback to `repoRoot` only if we fail to match `targetRoot`:

```ts
  for (const root of lookupRoots) {
    const direct = claude?.features.find(
      (feature) =>
        isSqaaFeature(feature) &&
        canonicalProjectRoot(feature.targetRoot) === root,
    );
    if (direct) return getProjectKey(direct);

    const fallback = claude?.features.find(
      (feature) =>
        isSqaaFeature(feature) &&
        getOptionalStringAttr(feature.attrs, 'repoRoot') !== undefined &&
        canonicalProjectRoot(getOptionalStringAttr(feature.attrs, 'repoRoot')!) === root,
    );
    if (fallback) return getProjectKey(fallback);
  }
```

src/cli/commands/context/index.ts
```diff
@@ -137,6 +141,10 @@ function resolveRecordedContextAugmentationConfig(cwd: string): RecordedContextA
       organization: getOptionalStringAttr(match.attrs, 'orgKey'),
       projectKey: getOptionalStringAttr(match.attrs, 'projectKey'),
       serverUrl: getOptionalStringAttr(match.attrs, 'serverUrl'),
+      // CAG daemon folder: git working-tree root (climbs up from subdirs), or the
+      // physical integrate targetRoot outside a git repo. Project metadata above
+      // comes from recorded state matched via repoRoot / targetRoot.
+      workspaceDir: canonicalizePath(await resolveContextWorkspaceRoot(cwd, match.targetRoot)),
```
michael-jabbour-sonarsource, line 127: Thanks for implementing this logic. I tested on one repository and it seems to work as expected 🎉

Question for my curiousity: I noticed that the resolution in sqaa-auth doesn't utilitze this logic, and they have a different but similarish algorith. Why? I feel that the same resolution could also apply to SQAA here. WDYT?

src/cli/commands/context/index.ts
```diff
@@ -113,17 +109,25 @@ function getOptionalStringAttr(
...
       .integrations.installed.flatMap((integration) =>
         integration.features.filter(isProjectContextAugmentationFeature).map((feature) => ({
           feature,
-          projectRoot: canonicalPath(feature.targetRoot),
+          // Prefer the stable main-working-tree key recorded at integrate time;
+          // fall back to targetRoot for state written by older CLI versions.
+          projectRoot: pathComparisonKey(
+            getOptionalStringAttr(feature.attrs, 'repoRoot') ?? feature.targetRoot,
+          ),
         })),
```
michael-jabbour-sonarsource: As far as I understand, this matching ignores `targetRoot` (in favor or `repoRoot`) for all newly recorded state. I think that this can cause matches across different worktrees (depending on the order of items in `features`).

I asked Claude, and it came with the following suggestion (I have no idea if this is a good solution or not):

```ts
const candidates = (await resolveWorktreeEquivalentPaths(cwd)).map(pathComparisonKey);

// The first candidate is always the physical cwd, canonicalized.
// In a linked worktree this is the linked-worktree path, while later
// candidates may be equivalent paths in the main checkout.
const currentPath = candidates[0];

const matches = features
  .flatMap((feature) => {
    const targetRoot = pathComparisonKey(feature.targetRoot);
    const repoRoot = getOptionalStringAttr(feature.attrs, 'repoRoot');

    if (currentPath && isPathInside(targetRoot, currentPath)) {
      return [
        {
          feature,

          // Rank 0: exact/current install target.
          // This feature was installed in the physical tree we are running from.
          rank: 0,
          projectRoot: targetRoot,
        },
      ];
    }

    if (repoRoot) {
      const repoRootKey = pathComparisonKey(repoRoot);

      if (candidates.some((path) => isPathInside(repoRootKey, path))) {
        return [
          {
            feature,

            // Rank 1: git-aware shared repository match for newly recorded integrations
            // Newer state records attrs.repoRoot as the stable main worktree.
            rank: 1,
            projectRoot: repoRootKey,
          },
        ];
      }

      return [];
    }

    if (candidates.some((path) => isPathInside(targetRoot, path))) {
      return [
        {
          feature,

          // Rank 2: legacy fallback.
          // Older state may not have attrs.repoRoot, so targetRoot is the only
          // recorded root.
          rank: 2,
          projectRoot: targetRoot,
        },
      ];
    }

    return [];
  })
  .sort(
    (a, b) =>
      // Prefer the strongest kind of match first.
      a.rank - b.rank ||
      b.projectRoot.length - a.projectRoot.length ||
      getUpdatedAtTimestamp(b.feature) - getUpdatedAtTimestamp(a.feature),
  );
```

michael-jabbour-sonarsource (approved):
Thanks for the fixing the environment variable during integrate. Seems to work as expected 🎉

Leaving a few minor non-blockers/questions that I have noticed while looking at the implementation. I let you whether they are worth looking into and how to do it. I am also happy to discuss in more detail tomorrow in our meeting.

src/cli/commands/context/index.ts
sophio-japharidze-sonarsource: I have tried to unify them.

src/lib/project-workspace/recorded-feature-resolver.ts
```diff
@@ -0,0 +1,157 @@
...
+ * True when `parent` equals or is an ancestor of `child`. Both arguments must
+ * already be comparison keys (see `pathComparisonKey`) so the string and
+ * `relative` checks are case- and canonicalization-consistent.
+ */
+function isPathInside(parent: string, child: string): boolean {
+  if (parent === child) {
+    return true;
+  }
+  const rel = relative(parent, child);
+  return rel !== '' && !rel.startsWith('..') && !isAbsolute(rel);
```
michael-jabbour-sonarsource: nitpick: `rel.startsWith('..')` can return true for directories whose name really start with '..' (e.g. '..myDirectory'). Doesn't it?

michael-jabbour-sonarsource (approved):
I haven't looked at the newest changes again, but I did some testing with my PR, and it seems to work as expected 🎉
Thanks again for taking care of this, and let me know if you would prefer that I take a closer look at something...

vnaskos-sonar (approved):
Tested the basic scenarios worked as expected 👌

src/lib/project-workspace/recorded-feature-resolver.ts
sophio-japharidze-sonarsource: I have pushed a fix

PR #560, opened 2026-07-07
GROW-213 Provision SonarQube project in sonar import command

kevinmlsilva:
## Summary
- Wires up the final step of `sonar import`: after org/repo selection, creates the bound SonarQube project via the `provision_projects` API and prints its project key.
- Resolves the org's connected DevOps platform to format the `installationKeys` param correctly (`<slug>|<id>` for GitHub, plain `id` otherwise).
- Enforces the org's public/private project visibility rules (`onlyPrivateProjects` + billing entitlement) before provisioning, rejecting mismatched repos.
- Already-imported repos are filtered out of the interactive picker instead of just labeled; fails clearly if none remain to import.
- Command remains hidden while in development.

Continuation of GROW-212.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun test tests/integration/specs/import/import.test.ts` (29 pass, including new provisioning/visibility test cases)

JIRA: GROW-213


src/cli/commands/import/_common/resolve-options.ts
```diff
@@ -18,23 +18,81 @@
...
-): Promise<string> {
+): Promise<ResolvedOrg> {
   if (!client.isCloud) {
     throw new CommandFailedError('sonar import is only supported on SonarQube Cloud.', {
       remediationHint: "Run 'sonar auth login' and connect to SonarQube Cloud, then retry.",
     });
   }
 
-  if (opts.org) return opts.org;
+  if (opts.org) return { key: opts.org };
```
nquinquenel: When org is passed, it only returns `key`. Later we do:
```
  const onlyPrivateProjects: OnlyPrivateProjects = {
    enabled: onlyPrivateProjectsEnabled ?? false,
    available: await client.hasPrivateProjectsEntitlement(orgKey),
  };
```
So it means when org is passed it never loads the org's private projects. Is it expected?

tests/integration/specs/import/import.test.ts
```diff
@@ -532,4 +579,310 @@ describe('sonar import', () => {
...
+
+        expect(result.exitCode).toBe(1);
+        const output = result.stdout + result.stderr;
+        expect(output).toContain("not found in the selected organization's DevOps platform");
+      },
+      { timeout: 15000 },
+    );
+
+    it(
+      'rejects --repo for a public repo when onlyPrivateProjects is unavailable',
```
nquinquenel: I think name is not correct, should be "private repo" no?

src/sonarqube/client.ts
```diff
@@ -480,6 +508,30 @@ export class SonarQubeClient {
...
+      const endpoint = '/billing/entitlements';
+      const result = await this.get<{ entitlements: Array<{ allowedFeatures: string[] }> }>(
+        endpoint,
+        { resourceId: organizationUuid, resourceType: 'organization' },
+        resolveFromEndpoint(this.serverURL, endpoint),
+      );
+      return result.entitlements.some((e) => e.allowedFeatures.includes(entitlement));
+    } catch {
+      return false;
+    }
```
nquinquenel: Should we log something in debug on failure?

src/sonarqube/client.ts
```diff
@@ -569,6 +621,45 @@ export class SonarQubeClient {
...
+
+    try {
+      const endpoint = '/dop-translation/organization-bindings';
+      const result = await this.get<{
+        organizationBindings: Array<{ devOpsPlatform: string }>;
+      }>(endpoint, { organizationId }, resolveFromEndpoint(this.serverURL, endpoint));
+      return result.organizationBindings[0]?.devOpsPlatform;
+    } catch {
+      return undefined;
+    }
```
nquinquenel: Same here, should we log in debug something just in case? Or not worth it/expected

nquinquenel (approved):
Looks ok, one comment about org and private project is important to tackle I think

src/cli/commands/import/_common/resolve-options.ts
kevinmlsilva: Good catch, that wasn't expected. Fixed: added `SonarQubeClient.fetchOrganizationByKey()`, a targeted `/api/organizations/search?organizations=<key>` lookup (same pattern as `checkOrganization`), and `resolveOrg`'s `--org` branch now uses it to populate both `almKey` and `onlyPrivateProjectsEnabled` instead of leaving them undefined. This also means `onlyPrivateProjects.enabled` is now correctly enforced when using `--org` directly, not just in the interactive flow. Added a regression test for it.

tests/integration/specs/import/import.test.ts
kevinmlsilva: You're right, this test title was stale — fixed, it now reads "rejects --repo for a private repo when onlyPrivateProjects is unavailable", matching the private repo used in the test body.

src/sonarqube/client.ts
kevinmlsilva: Added — the catch block now calls `logger.debug(`Failed to check '${entitlement}' billing entitlement`, err)` before returning `false`.

src/sonarqube/client.ts
kevinmlsilva: Added — same treatment: `logger.debug('Failed to fetch organization-bindings for ALM key lookup', err)` in the catch block before returning `undefined`.

PR #566, opened 2026-07-09
CLI-737 CLI-738 CLI-739 Pass network config to mcp

Krosovok:
----
## Summary by Gitar

- **MCP Configuration:**
    - Propagate proxy settings, CA certificates, and client certificates to the MCP container via environment variables and volume mounts.
    - Implement `toJavaNonProxyHosts` to translate standard `NO_PROXY` formats into Java-compatible `http.nonProxyHosts` patterns.
- **Certificate Management:**
    - Add `pemToPkcs12` utility to convert PEM client certificates into PKCS12 format required by the Java runtime inside the MCP container.
    - Automate conversion and temporary file storage for PEM-formatted client certificates.
- **Filesystem and Infrastructure:**
    - Introduce `CLI_TMP_DIR` for storing generated files like converted certificates.
    - Include `CLI_TMP_DIR` in the `system reset` cleanup cycle.
- **WSL Compatibility:**
    - Use `WSLENV` with `/p` translation for certificate paths to ensure correct cross-environment volume mounting.


src/lib/mcp/mcp-helper.ts
```diff
@@ -81,16 +86,106 @@ export function getMcpConfig(
...
 }
 
+// Path where update-ca-certificates picks up custom CA certs inside the MCP container (Debian/Alpine convention).
+export const MCP_CONTAINER_CA_CERT_PATH = '/usr/local/share/ca-certificates/sonar-ca.crt';
+
+// Fixed mount path for the client certificate PKCS12 keystore inside the MCP container.
+export const MCP_CONTAINER_CLIENT_CERT_PATH = '/etc/ssl/mcp/client.p12';
+
+// Temp file used when converting PEM cert+key to PKCS12 for the MCP container mount.
+const MCP_CLIENT_CERT_TEMP_PATH = join(CLI_TMP_DIR, 'mcp-client-cert.p12');
```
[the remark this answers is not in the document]
Krosovok: Concurrency is not likely, since the user is unlikely interact with multiple CLIs trying to set up the same thing. Creating a unique file will quickly pollute the users machine, since the MCPs are likely to be run more in the future. That reveals another issue though: we probably don't want to reconstruct the file. Probably there is a way to give a unique identifier based on the key.

src/lib/mcp/mcp-helper.ts
```diff
@@ -81,16 +86,106 @@ export function getMcpConfig(
...
+      keystorePassphrase: clientCert.passphrase,
+    };
+  }
+  const pkcs12Buffer = pemToPkcs12(clientCert.resolvedCertPem, clientCert.resolvedKeyPem);
+  mkdirSync(dirname(MCP_CLIENT_CERT_TEMP_PATH), { recursive: true });
+  writeFileSync(MCP_CLIENT_CERT_TEMP_PATH, pkcs12Buffer);
+  return {
+    hostCertPath: normalizePath(MCP_CLIENT_CERT_TEMP_PATH),
+    keystorePassphrase: undefined,
+  };
```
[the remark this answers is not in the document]
Krosovok: How do these modes map to non-Unix systems?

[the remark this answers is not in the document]
Krosovok: We need file itself to be accessable by docker user, so we can't restrict the access to the file itself. Best-effort we can do: make the directory non-traversable and delete the file after the MCP exits.

src/lib/mcp/mcp-helper.ts
```diff
@@ -81,16 +87,114 @@ export function getMcpConfig(
...
+    return {
+      hostCertPath: normalizePath(clientCert.certPath),
+      keystorePassphrase: clientCert.passphrase,
+    };
+  }
+  const cachedPath = clientCertCachePath(clientCert);
+  const cacheDir = dirname(cachedPath);
+  mkdirSync(cacheDir, { recursive: true, mode: NON_TRAVERSABLE }); // make the directory non-traversable
+  if (!existsSync(cachedPath)) {
+    const pkcs12Buffer = pemToPkcs12(clientCert.resolvedCertPem, clientCert.resolvedKeyPem);
```
tomshafir-sonarsource: Should the passphrase be threaded through here so that an encrypted PEM private key can be decrypted?

src/cli/commands/run/mcp.ts
```diff
@@ -90,4 +101,13 @@ export async function runMcp(auth: ResolvedAuth, options: McpRunOptions = {}): P
...
   });
+
+  if (network.clientCert?.format === 'pem') {
+    try {
+      rmSync(clientCertCachePath(network.clientCert), { force: true });
+    } catch (err) {
+      // Windows: Docker Desktop may still hold the file; content-addressed so safe to leave behind
+      logger.warn(`Could not remove cached client certificate: ${String(err)}`);
+    }
+  }
```
tomshafir-sonarsource: Nit: This cleanup should be done in a `finally` block so it also runs when the `spawn` fails.

tomshafir-sonarsource (approved):
Looks good! Left two comments.

src/lib/mcp/mcp-helper.ts
Krosovok: Good catch! I tested it without a passphrase for PEM case, so it slipped.

src/lib/crypto/pkcs12.ts
```diff
@@ -86,6 +86,22 @@ function serializeToPem(certBags: forge.pkcs12.Bag[], keyBag: forge.pkcs12.Bag):
   return { cert, key: forge.pki.privateKeyToPem(keyBag.key) };
 }
 
+export function pemToPkcs12(certPem: string, keyPem: string): Buffer {
+  try {
+    const cert = forge.pki.certificateFromPem(certPem);
+    const key = forge.pki.privateKeyFromPem(keyPem);
```
nquinquenel: I think conversion of encrypted PEM key will fail with no passphrase

src/cli/commands/run/mcp.ts
```diff
@@ -44,7 +51,11 @@ function debugLog(message: string): void {
   process.stderr.write(`[sonarqube-cli] DEBUG ${message}\n`);
 }
 
-export async function runMcp(auth: ResolvedAuth, options: McpRunOptions = {}): Promise<void> {
+export async function runMcp(
+  auth: ResolvedAuth,
+  options: McpRunOptions = {},
+  network: ResolvedNetworkConfig = getNetworkConfig(),
```
nquinquenel: Should it be `getNetworkConfig` or `getNetworkConfigOrThrow` ?

Krosovok: Can be `getNetworkConfigOrThrow`, it would make it fail fast.

PR #576, opened 2026-07-13
CLI-825 Migrate `CliCommandExecuted` event to the new telemetry pipeline

tomshafir-sonarsource:
----
## Summary by Gitar

- **Telemetry Migration:**
  - Migrated `CliCommandExecuted` events from the legacy `state.telemetry.events` queue to the `findings.ndjson` telemetry pipeline.
  - Added `migrateLegacyTelemetryEvents` to `runPostUpdateActions` to ensure seamless transition of pending events during updates.
- **Refactoring:**
  - Unified telemetry event handling by moving `CliCommandExecuted` logic into `src/telemetry/findings.ts` and removing it from the state-persistence flow.
  - Updated `StoredAnalysisEvent` type and telemetry storage schemas to accommodate the new event type.
- **Testing Infrastructure:**
  - Introduced `telemetry-helpers.ts` to centralize test utilities for event validation and telemetry state management.
  - Updated existing telemetry unit tests to reflect the transition from state-based to file-based event persistence.
- **Documentation:**
  - Updated `CLAUDE.md` to reflect the consolidation of all telemetry events into the `findings.ndjson` pipeline.


src/lib/post-update.ts
```diff
@@ -84,12 +85,37 @@ export async function runPostUpdateActions(): Promise<void> {
...
+ * `state.telemetry.events` queue into findings.ndjson.
+ */
+export function migrateLegacyTelemetryEvents(): void {
+  const state = loadState();
+  const legacyEvents = state.telemetry.events ?? [];
+  if (legacyEvents.length === 0) {
+    return;
+  }
+
+  const unmigrated = legacyEvents.filter((event) => !appendAnalysisEvent(event));
```
vnaskos-sonar: Initial thought:
small robustness thing on the migration: we append every event to the ndjson first and only then saveState to clear the queue. if the appends succeed but saveState throws here, the outer catch in runPostUpdateActions swallows it and cliVersion never gets bumped, so the next launch re-runs the migration and re-appends the same events → duplicates in the ndjson.

I think it's low impact: it's a narrow window (append to the ndjson works but writing state.json in the same dir fails), and the re-appended events keep their original event_id so the backend should dedupe them.

but then:
do we actually want the migration? 🤔 state.telemery.events only has entries when a previous sent failed, events[] should be empty the great majority of the time.

tests/unit/cli/commands/update/post-update.test.ts
```diff
@@ -36,19 +37,26 @@ import {
...
 import * as migration from '../../../../../src/lib/migration';
 import {
   migrateClaudeCodeHooks,
   migrateDeclarativeIntegrations,
+  migrateLegacyTelemetryEvents,
   runPostUpdateActions,
   updateScaScannerBinaryIfNeeded,
   updateSecretsBinaryIfNeeded,
 } from '../../../../../src/lib/post-update';
 import * as stateRepository from '../../../../../src/lib/repository/state-repository';
```
vnaskos-sonar, line 51: comment for state-repository.ts line 55

minor inconsistency now that getDefaultState dropped telemetry.events: migrateState still injects events: [] when the telemetry block is missing. since events is legacy/optional from now on, I'd drop it here too so both init paths agree, otherwise a state migrated from a missing telemetry block gets a dead events: [] that never goes away.

vnaskos-sonar (approved):
some minor comments 👍

PR #586, opened 2026-07-15
CLI-810 Don't skip hooks silently if unauthenticated or missing binary

tomshafir-sonarsource:
----
## Summary by Gitar

- **Hook behavior:**
  - Modified secret scanning hooks to "fail closed" instead of skipping silently when authentication is missing or the `sonar-secrets` binary is unavailable.
  - Updated `resolveAuthAndSecrets` to throw a `MissingDependenciesError` when environment conditions are unmet.
- **Error handling:**
  - Added `MissingDependenciesError` to propagate auth or binary issues, ensuring hooks explicitly block or notify the user instead of ignoring the scan.
  - Updated `git-pre-commit` and `git-pre-push` hooks to catch these errors and exit with code 1.
  - Updated agent prompt hooks to output a JSON block with the error reason when dependencies are missing.
- **Documentation:**
  - Updated `CLAUDE.md` to clarify that hooks now fail closed and to distinguish between scan-error behavior and dependency-gate behavior.


src/cli/commands/hook/copilot-pre-tool-use.ts
```diff
@@ -63,8 +68,19 @@ export async function copilotPreToolUse(): Promise<void> {
...
   if (!filePath || !existsSync(filePath)) return;
 
-  const deps = await resolveAuthAndSecrets();
-  if (!deps) return;
+  let deps: HookDependencies;
+  try {
+    deps = await resolveAuthAndSecrets();
+  } catch (err) {
+    if (err instanceof MissingDependenciesError) {
+      process.stdout.write(
```
georgii-borovinskikh-sonarsource: I suggest creating the helper methods for producing the structured messages to the agents: the helpers would hide the logic and let the caller just provide the simple reason for deny

src/cli/commands/hook/git-pre-commit-dependency-risks.ts
```diff
@@ -67,7 +67,9 @@ export interface DepRisksStageOptions {
 export async function runDepRisksStage(options: DepRisksStageOptions): Promise<void> {
   const binaryPath = resolveScaScannerBinaryPath();
   if (!binaryPath) {
-    logger.debug('Dependency-risks hook: sca-scanner binary not installed, skipping.');
+    warn(
```
georgii-borovinskikh-sonarsource, line 71: Small suggestion: still keep the warning logged (maybe elevate to log.warn)

src/cli/commands/hook/git-pre-commit.ts
```diff
@@ -45,7 +46,9 @@ export async function gitPreCommit(
   if (stagedFiles.length === 0) return;
 
   const auth = await resolveAuth().catch(() => null);
-  if (!auth) return;
+  if (!auth) {
+    throw new MissingDependenciesError(SECRETS_INACTIVE_UNAUTHENTICATED);
```
georgii-borovinskikh-sonarsource: The message here ideally should not be specific to secrets, auth is needed for all subhooks

georgii-borovinskikh-sonarsource (approved):

kirill-knize-sonarsource (approved):
LGTM

PR #611, opened 2026-07-22
GROW-257 Add --regex selection mode to sonar import

kevinmlsilva:
## Summary
- Adds `--regex <regex>` as a third way to choose which repositories to import, alongside `--all` (every eligible repository) and `--repo` (an explicit list). `--regex` imports only eligible repositories whose DevOps platform **name** (not slug) matches the pattern.
- The three are mutually exclusive, validated eagerly with a combined error message (e.g. `--all, --repo cannot be combined`); `--non-interactive` requires exactly one of them.
- Interactively, omitting all three shows a **Recommended / Manual / By pattern / ← Back** menu — choosing "By pattern" prompts for a **mandatory** (non-skippable) regex: blank or invalid input just re-prompts, and cancelling (Ctrl+C) returns to the menu.
- Supports `/pattern/flags` literal syntax (e.g. `/^archived-/i`) for flags like case-insensitive matching, since JS has no inline `(?i)` modifier the way PCRE/Python do.
  - `g`/`y` flags are silently stripped — the compiled regex is reused via `.test()` across many independent repository names, where `lastIndex` statefulness would otherwise cause alternating false negatives.
  - The `/pattern/flags` detection requires a non-empty pattern body and a valid flag-character suffix, so a plain pattern that merely looks delimited (e.g. `/etc/passwd`) is compiled as a literal instead of being misparsed as invalid flags or silently matching everything (e.g. `//`).

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run test:unit` (2100 pass)
- [x] `bun run test:integration` for `tests/integration/specs/import/import.test.ts` (69 pass)
- [x] Manual smoke test against the compiled binary: `--regex` flag (standalone, bypasses the mode-select menu), interactive "By pattern" selection (typed pattern, mandatory/non-skippable, cancel-returns-to-menu), `/pattern/i` case-insensitive matching, invalid regex and flag-conflict validation (exit 2), and the two ambiguous `/pattern/flags`-parsing edge cases
- [x] SonarQube Agentic Analysis (DEEP) on all changed files in earlier iterations of this branch — clean aside from a pre-existing `command-tree.ts` file-length finding, unrelated to this change


vnaskos-sonar:
I found some weird UX cases

1. The `--exclude` must always be used in combination with `--all` --> This could change to `sonar import --exclude "..."` instead of `sonar import --all --exclude "..."`
2. On interactive mode when passing `--all` the user is asked to whether they want to exclude, similar to the scenario when they don't pass `--all`. --> `--all` looks like I intentionally want all of them, and we could skip the question to exclude.
3. On non-interactive mode the `--exclude` is silently skipped
4. `sonar import --exclude "..." --non-interactive` produces an error message that `--all` or `--repos` must be passed, but `--repos` is not allowed with `--exclude`.

@cliffordgoh-sonar could have a look on this. For reference these are all the possible cases, 

### Interactive (no `--non-interactive`)

| `--all` | `--repo` | `--exclude` | Outcome |
|:--:|:--:|:--:|---|
| – | – | – | Mode menu (Recommended / Manual). Recommended → **optional exclude prompt** → import all eligible; Manual → hand-pick. |
| ✓ | – | – | Imports all eligible, but **shows the optional exclude prompt first** ⚠️ |
| – | ✓ | – | Imports exactly the named repo(s). No menu. |
| – | – | ✓ | Mode menu. Recommended → **all-except-excluded, no extra prompt** ✓; Manual → **`--exclude` silently ignored** ⚠️ |
| ✓ | ✓ | – | ❌ exit 2 — `--all cannot be combined with --repo` |
| ✓ | – | ✓ | Imports all eligible **except excluded**, no prompt ✅ *(the clean path)* |
| – | ✓ | ✓ | ❌ exit 2 — `--exclude cannot be combined with --repo` |
| ✓ | ✓ | ✓ | ❌ exit 2 — `--all cannot be combined with --repo` *(fires first; exclude conflict never evaluated)* |

### Non-interactive (`--non-interactive`)

| `--all` | `--repo` | `--exclude` | Outcome |
|:--:|:--:|:--:|---|
| – | – | – | ❌ exit 2 — `--repo or --all is required in non-interactive mode` |
| ✓ | – | – | Imports all eligible. No prompt. |
| – | ✓ | – | Imports exactly the named repo(s). |
| – | – | ✓ | ❌ exit 2 — `--repo or --all is required…` ⚠️ *(the `--all`/`--repo` guard fires first, so you get this message — **not** an exclude-specific one — even though you did pass a filter)* |
| ✓ | ✓ | – | ❌ exit 2 — `--all cannot be combined with --repo` |
| ✓ | – | ✓ | Imports all eligible **except excluded**. No prompt. ✅ |
| – | ✓ | ✓ | ❌ exit 2 — `--exclude cannot be combined with --repo` |
| ✓ | ✓ | ✓ | ❌ exit 2 — `--all cannot be combined with --repo` *(fires first)* |

src/commands/command-tree.ts
```diff
@@ -226,6 +226,18 @@ COMMAND_TREE.command('import', { hidden: true })
...
+    '--exclude <regex>',
+    'Exclude repositories whose DevOps platform name (not slug) matches this regular expression.\n' +
+      '\n' +
+      'Case-sensitive by default; wrap as /pattern/flags to add flags like case-insensitive\n' +
+      'matching. Only applies when importing all eligible repositories (--all, or interactively\n' +
+      'choosing "Recommended"); has no effect on Manual selection. Cannot be combined with --repo.\n' +
+      '\n' +
+      'Examples:\n' +
+      '    --exclude "^test-"\n' +
+      '    --exclude "/^archived-/i"\n',
```
vnaskos-sonar:
```
Skip repositories whose name matches this pattern.
The name is the bare repo name ("my-repo"), not the
"owner/my-repo" slug.

Examples:
  --exclude "^archived-"      skip names starting with "archived-"
  --exclude "/^archived-/i"   same, but case-insensitive (matches ARCHIVED-…)

Applies only with --all or "Recommended"; ignored for Manual;
cannot be combined with --repo.

File: src/commands/command-tree.ts [hunk 2]
```diff
@@ -226,6 +226,18 @@ COMMAND_TREE.command('import', { hidden: true })
     '--all',
     'Import every eligible repository in the organization (not already imported, and allowed by its project visibility settings). Cannot be combined with --repo.',
   )
+  .option(
+    '--exclude <regex>',
+    'Exclude repositories whose DevOps platform name (not slug) matches this regular expression.\n' +
```
vnaskos-sonar: There is an inconsistency, `--repos` accepts repo slug but `--exclude` accepts names.

cliffordgoh-sonar:
Thanks @vnaskos-sonar cc @kevinmlsilva 
I landed on below UX

- make `--exclude` a standalone scope flag, same tier as `--all`/`--repo`, and always fail loud
- each of `--all`, `--repo`, `--exclude` independently answers "which repos" combining them will fail except `--all`+`--exclude`

| `--all` | `--repo` | `--exclude` | Outcome |
|:---:|:---:|:---:|---|
| – | – | – | Mode menu: Recommended → optional exclude prompt → import all; Manual → hand-pick |
| ✓ | – | – | Imports all eligible, no prompt |
| – | ✓ | – | Imports exactly the named repo(s), no menu |
| – | – | ✓ | Imports all eligible **except excluded** — no menu, `--all` not required |
| ✓ | ✓ | – | ❌ exit 2 — `--all cannot be combined with --repo` |
| ✓ | – | ✓ | Imports all eligible **except excluded** (the clean path) |
| – | ✓ | ✓ | ❌ exit 2 — `--exclude cannot be combined with --repo` |
| ✓ | ✓ | ✓ | ❌ exit 2 — `--all cannot be combined with --repo` *(fires first)* |

src/commands/command-tree.ts
cliffordgoh-sonar: @kevinmlsilva Is there a reason why `--repo` is repo slug but `--exclude` is repo name?

kevinmlsilva:
> Thanks @vnaskos-sonar cc @kevinmlsilva I landed on below UX
> 
> * make `--exclude` a standalone scope flag, same tier as `--all`/`--repo`, and always fail loud
> * each of `--all`, `--repo`, `--exclude` independently answers "which repos" combining them will fail except `--all`+`--exclude`
> 
> `--all`	`--repo`	`--exclude`	Outcome
> –	–	–	Mode menu: Recommended → optional exclude prompt → import all; Manual → hand-pick
> ✓	–	–	Imports all eligible, no prompt
> –	✓	–	Imports exactly the named repo(s), no menu
> –	–	✓	Imports all eligible **except excluded** — no menu, `--all` not required
> ✓	✓	–	❌ exit 2 — `--all cannot be combined with --repo`
> ✓	–	✓	Imports all eligible **except excluded** (the clean path)
> –	✓	✓	❌ exit 2 — `--exclude cannot be combined with --repo`
> ✓	✓	✓	❌ exit 2 — `--all cannot be combined with --repo` _(fires first)_

just a small remark

its not --exclude but --regex

the rest does look perfect and implemented

src/commands/command-tree.ts
```diff
@@ -219,13 +218,24 @@ COMMAND_TREE.command('import', { hidden: true })
...
   )
   .option(
     '--all',
-    'Import every eligible repository in the organization (not already imported, and allowed by its project visibility settings). Cannot be combined with --repo.',
+    'Import every eligible repository in the organization (not already imported, and allowed by its project visibility settings). Cannot be combined with --repo or --regex.',
+  )
+  .option(
+    '--regex <regex>',
+    'Import only eligible repositories whose DevOps platform name (not slug) matches this regular expression.\n' +
+      '\n' +
```
vnaskos-sonar, line 232: new lines of the --regex help are not aligning with the first line.

Image

vnaskos-sonar (approved):
👍  Left just a small visual fix

cliffordgoh-sonar:
Tagging @farah-bouassida for visbility
cc @vnaskos-sonar 
We went with the following UX

sonar import 
--all
--repo (accept multiple values)
--regex regex // (Will only import if regex returns true)

PR #617, opened 2026-07-24
CLI-837 Move remainin /lib/** files to /core/**, fix imports

georgii-borovinskikh-sonarsource:


vnaskos-sonar (commented):
1. Stale path in Cursor rules `.cursor/rules/sonarqube-cli.mdc:26`. Still says "All paths and URLs must be imported from src/lib/config-constants.ts", now points at a deleted file. Should be `src/core/config-constants.ts`. (Not in the PR diff.)
2. Stale comment `src/commands/analyze/sqaa-api.ts:72`. Doc comment reads "Throwing wrapper over lib/fs-utils.toRelativePosixPath"; the module is now core/io/fs-utils.
3. Sentry test fixtures, `tests/unit/core/server/lib/sentry.test.ts:294,308`. Use src/lib/state-manager.ts as sample path strings for frame-path normalization.
4. Some test dirs (e.g. tests/unit/core/server/lib/) don't mirror the new src/core layout

vnaskos-sonar (approved):

PR #618, opened 2026-07-24
CLI-837 Move common error and version models to /core

georgii-borovinskikh-sonarsource:
----
## Summary by Gitar

- **Refactoring:**
    - Moved common error and version models from commands to `src/core/` (`src/core/command-error.ts`, `src/core/version.ts`)


vnaskos-sonar (approved):
Nothing major 👍 

minor notes:
1. Two modules named version.ts inside `core/`. `src/core/version.ts` (the Version class) imports stripBuildNumber / isNewerVersion from `src/core/host/version.ts`. Same basename, one layer apart, one depending on the other, seems ambiguous in import lines. They could be combined into a single one.

2. Three error modules, with the two most similarly-named at the same level:
- `core/errors.ts` - `ApiCallError`, `NetworkConfigError`, `CryptographicError`
- `core/command-error.ts` - the `CliError` with exit codes (imports from the above)
- `core/server/errors.ts` - HTTP errors

    The roles are genuinely distinct so nothing is wrong here, but errors.ts vs command-error.ts can be easily missintepreted especially by a newcomer/non-squad-member deciding where a new error type goes. Probably a follow-up task rather than this PR.

PR #625, opened 2026-07-27
GROW-290 Restrict sonar import to GitHub and Azure DevOps organizations

kevinmlsilva:
## What

`sonar import` now refuses to run unless the connection's organization is bound to GitHub or Azure DevOps. Any other platform (GitLab, Bitbucket) fails with:

```
sonar import supports GitHub and Azure DevOps organizations only, but 'my-org' is bound to GitLab.
💡 Import these repositories from SonarQube Cloud in your browser instead.
```

## How

`assertSupportedAlm()` (`import/_common/resolve-options.ts`) checks the ALM key against a single `KNOWN_ALMS` table that carries both the display name and whether the platform is supported. Unrecognized keys are rejected too — it's an allowlist, not a blocklist.

It's called from `resolveOrgAndRepos()` right after the ALM key resolves (including the `organization-bindings` fallback) and before any repository page is fetched, so an unsupported org stops immediately rather than after a full paginated repo fetch.

An **undefined** ALM key is deliberately not rejected: it means unknown, not known-unsupported. An org with nothing bound has no repositories to import and already falls through to the existing "no repositories found" path, and `getOrganizationAlmKey()` swallows its own errors — so failing there would turn a transient lookup failure into a hard stop for supported orgs too.

## Tests

`tests/integration/specs/import/import.test.ts` — 66 pass, 0 fail. Three new cases: GitLab via `alm.key` exits 1 (asserting `dop-repositories` and `provision_projects` are never called), Bitbucket via the bindings fallback exits 1, Azure via the bindings fallback imports normally. The two existing "non-GitHub" cases now use an Azure fixture, since GitLab would abort before reaching what they assert.

Harness: new `withOrganizationBinding(orgId, devOpsPlatform)` on the fake server, so a test can model an org that has a binding but no `alm` on the org lookup — previously `devOpsPlatform` was always derived from `alm.key` and that path couldn't be exercised on its own.


src/commands/import/_common/resolve-options.ts
```diff
@@ -38,8 +38,22 @@ import {
   type SkippedRepo,
 } from './repository-collection';
 
-/** ALM key used by GitHub-bound organizations, per `Organization.alm.key`. */
+/* ALM keys as the server spells them. */
 const GITHUB_ALM_KEY = 'github';
+const AZURE_DEVOPS_ALM_KEY = 'azure';
```
vnaskos-sonar: Noticed that on SQS/SQC the alm for azure contains more than azure.

`if (almKey === 'microsoft' || almKey.startsWith('azure'))`

Should it also be the case here?

src/commands/import/_common/resolve-options.ts
```diff
@@ -153,6 +167,34 @@ async function resolveOrgByKey(client: SonarQubeClient, orgKey: string): Promise
...
+ * and Azure DevOps are.
+ *
+ * An undefined `almKey` is deliberately not treated as unsupported. It means the platform is
+ * unknown rather than known-unsupported: either the org has nothing bound to it — in which case
+ * it has no repositories to import and falls through to the existing "no repositories found"
+ * path — or `getOrganizationAlmKey`'s lookup failed, and that best-effort call swallows its
+ * errors, so failing here would turn a transient error into a hard stop for supported orgs too.
+ */
+export function assertSupportedAlm(orgKey: string, almKey: string | undefined): void {
+  if (almKey === undefined) {
```
vnaskos-sonar: For an org whose /api/organizations/search record has no alm but which is bound to GitLab, a transient organization-bindings failure makes almKey undefined → the guard passes → fetchDopRepositoriesPage is keyed only on organizationId, so it can still return GitLab repos and provision them.

Can dop-repositories return repositories when the org record carries no alm? If yes, the restriction is silently bypassable on a transient failure.

src/commands/import/_common/resolve-options.ts
```diff
@@ -153,6 +167,34 @@ async function resolveOrgByKey(client: SonarQubeClient, orgKey: string): Promise
...
+    return;
+  }
+
+  // A platform we've never heard of is unsupported too: this is an allowlist, not a blocklist.
+  const alm = KNOWN_ALMS[almKey];
+  if (alm?.supported === true) {
+    return;
+  }
```
vnaskos-sonar: This took a bit of time to understand, and `KNOWN_ALMS` is not used elsewhere, it could change to a simple `if (alm == github || alm == azure)`.

src/commands/import/_common/resolve-options.ts
```diff
@@ -153,6 +167,34 @@ async function resolveOrgByKey(client: SonarQubeClient, orgKey: string): Promise
...
+
+  throw new CommandFailedError(
+    `sonar import supports GitHub and Azure DevOps organizations only, but '${orgKey}' is bound ` +
```
vnaskos-sonar, line 180: It's a bit strange to have `KNOWN_ALMS` and hardcoded values for what is supported. For this PR, we could drop the `KNOWN_ALMS` for simplicity and keep the hardcoded values (to be tackled later) as mentioned on the previous comment.

src/commands/import/_common/resolve-options.ts
```diff
@@ -153,6 +167,34 @@ async function resolveOrgByKey(client: SonarQubeClient, orgKey: string): Promise
...
+ * path — or `getOrganizationAlmKey`'s lookup failed, and that best-effort call swallows its
+ * errors, so failing here would turn a transient error into a hard stop for supported orgs too.
+ */
+export function assertSupportedAlm(orgKey: string, almKey: string | undefined): void {
+  if (almKey === undefined) {
+    return;
+  }
+
+  // A platform we've never heard of is unsupported too: this is an allowlist, not a blocklist.
+  const alm = KNOWN_ALMS[almKey];
```
vnaskos-sonar: To make sure that it always matches regardless of the casing of the api reply

```suggestion
  const alm = KNOWN_ALMS[almKey.trim().toLowerCase()];
```

vnaskos-sonar (commented):
Nothing major but some possible edge cases and simplifications

vnaskos-sonar (approved):

PR #627, opened 2026-07-28
CLI-837 Move integration features framework and dependency installers to src/core/

georgii-borovinskikh-sonarsource:
----
## Summary by Gitar

- **Architecture refactoring:**
    - Moved integration features framework and dependency installers from `src/commands/_common/` and `src/commands/integrate/_common/` to `src/core/framework/` and `src/core/host/`
    - Decoupled `core/host/post-update.ts` from command modules by introducing `PostUpdateDependencies` interface injected via composition root


kirill-knize-sonarsource (approved):
LGTM

Code complexity findings ideally to be tracked and fixed later.

PR #628, opened 2026-07-28
CLI-866 Record env-var auth into state like login, skip the keychain write

kirill-knize-sonarsource:
- Env-var authentication (SONARQUBE_CLI_TOKEN + SONARQUBE_CLI_SERVER/SONARQUBE_CLI_ORG) previously left no trace in state.auth.connections, unlike sonar auth login. resolveAuth()'s env branch now calls a new recordConnectionFromAuth() helper, which records the connection (server/org/region/tokenName) and runs identity enrichment (userUuid/organizationUuidV4/sqsInstallationId) exactly like login does — the keychain write (saveToken()) is the only thing that stays login-exclusive.
- recordConnectionFromAuth() no-ops (no state write, no network call) when the active connection already matches the resolved auth and its identity is already complete; login.ts passes force: true to bypass that check, since an explicit login must always refresh authenticatedAt/tokenName.
- login.ts now reuses recordConnectionFromAuth() instead of duplicating the identity-fetch logic inline.
- telemetry/identity.ts drops its own env-or-state priority re-implementation (isEnvBasedAuth() ? resolveFromEnv() : resolveFromState()) in favor of calling the canonical resolveAuth() — removing the one place in the codebase that reimplemented auth resolution instead of reusing it.
- Extracted identity-fetch.ts (fetch/cache of identity fields, used by both telemetry and the connection recorder) and sonarcloud-region.ts (pure URL/region helpers, re-exported from auth-resolver.ts for compatibility) to avoid an import cycle between auth-resolver.ts and the new recorder.
- Moved authMatchesConnection from telemetry/identity.ts to state-manager.ts — shared by telemetry and auth-connection-recorder.ts.


src/core/state/state.ts
```diff
@@ -78,6 +78,8 @@ export interface AuthConnection {
   organizationUuidV4?: string | null;
   /** Installation ID of the SonarQube Server (fetched at auth time, SQS only) */
   sqsInstallationId?: string | null;
+  /** True when there is no keychain entry behind this connection. */
+  envOnly?: boolean;
```
georgii-borovinskikh-sonarsource, line 88: Does that mean that the case of env && keychain is possible?

src/core/state/state-manager.ts
```diff
@@ -37,6 +38,12 @@ export { tryLoadState } from './state-repository.ts';
 
 import { type AuthConnection, type CliState, type CloudRegion } from './state.ts';
 
+export function authMatchesConnection(auth: ResolvedAuth, conn: AuthConnection): boolean {
```
georgii-borovinskikh-sonarsource, line 41: why not `return auth.serverUrl === conn.serverUrl && (auth.connectionType !== 'cloud' || auth.orgKey === conn.orgKey)`

src/core/server/sonarcloud-region.ts
```diff
@@ -0,0 +1,84 @@
+/*
```
georgii-borovinskikh-sonarsource, line 1: I don't think this file should be in `src/core/host/`, looks more like a `src/core/server` helper

src/core/host/auth-resolver.ts
```diff
@@ -57,11 +58,34 @@ export interface ResolvedAuth {
...
+export interface ResolveAuthOptions {
+  /** Suppress the "partial env vars" warning. */
+  silent?: boolean;
+}
+
+// Env vars are fixed for the process lifetime, so this only needs to run once per process.
+let envAuthRecordAttempted = false;
+
+export async function resolveAuth(options: ResolveAuthOptions = {}): Promise<ResolvedAuth | null> {
+  const envAuth = resolveFromEnv(options);
```
georgii-borovinskikh-sonarsource, line 70: ideally the recording of the state should not happen in this method (in another private method, otherwise this method is mixing orchestration and env-source-specific state recording behavior)

src/core/host/auth-resolver.ts
```diff
@@ -57,11 +58,34 @@ export interface ResolvedAuth {
...
+        logger.debug(`Failed to record env-var connection in state: ${(err as Error).message}`);
+      });
+    }
+    return envAuth;
+  }
+  return await resolveFromState();
+}
+
+/** Test-only: resets the per-process env-auth record guard between test cases. */
+export function resetEnvAuthRecordGuard(): void {
```
georgii-borovinskikh-sonarsource, line 94: This is a bad pattern we have across the whole codebase: a static variable with a special method to reset it just for tests (even though nothing prevents other users to call this and mess up something, as it's still a public method)

src/core/host/auth-connection-recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/** No-ops when `auth` already matches a fully-enriched active connection. */
+export async function recordConnectionFromAuth(
+  auth: ResolvedAuth,
+  options: RecordConnectionOptions = {},
+): Promise<AuthConnection> {
+  const state = loadState();
+  const active = getActiveConnection(state);
+  const matches = active !== undefined && authMatchesConnection(auth, active);
```
georgii-borovinskikh-sonarsource: `isMatchingActiveConnection` or remove it completely since it's only used once

src/core/host/auth-connection-recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+  const seedConn = matches ? active : undefined;
+  const seedIdentity = identityFromConnection(seedConn);
+
+  if (
+    !options.force &&
```
georgii-borovinskikh-sonarsource, line 61: personal preference: logical operators at the end of the line are hard to read

src/core/host/auth-connection-recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+/** No-ops when `auth` already matches a fully-enriched active connection. */
+export async function recordConnectionFromAuth(
+  auth: ResolvedAuth,
+  options: RecordConnectionOptions = {},
+): Promise<AuthConnection> {
+  const state = loadState();
+  const active = getActiveConnection(state);
+  const matches = active !== undefined && authMatchesConnection(auth, active);
+  const seedConn = matches ? active : undefined;
+  const seedIdentity = identityFromConnection(seedConn);
```
georgii-borovinskikh-sonarsource: very weird model: it's called TelemetryIdentity (should be AuthIdentity?) and is filled with nulls when seedConn is undefined

src/core/host/auth-connection-recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/** No-ops when `auth` already matches a fully-enriched active connection. */
+export async function recordConnectionFromAuth(
+  auth: ResolvedAuth,
+  options: RecordConnectionOptions = {},
+): Promise<AuthConnection> {
+  const state = loadState();
+  const active = getActiveConnection(state);
+  const matches = active !== undefined && authMatchesConnection(auth, active);
+  const seedConn = matches ? active : undefined;
```
georgii-borovinskikh-sonarsource: `seedConnection`

src/core/host/auth-connection-recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  const connection = addOrUpdateConnection(state, auth.serverUrl, auth.connectionType, {
+    orgKey: auth.orgKey,
+    region: cloudRegionFromUrl(auth.serverUrl),
+    tokenName: options.tokenName,
+    envOnly: options.envOnly,
+  });
+
+  const identity = await resolveTelemetryIdentity(auth, seedIdentity);
```
georgii-borovinskikh-sonarsource, line 75: this whole bit about applying the telemetry identity to connection model deserves its own method

georgii-borovinskikh-sonarsource (approved):

PR #630, opened 2026-07-28
CLI-902 Send telemetry events through the resolved network configuration

vnaskos-sonar:
The telemetry flush was the only fetch in the CLI that did not pass `buildFetchNetworkOptions`, so it ignored the proxy, custom CA, and client certificate settings every other request honors. Behind an enforced corporate proxy it failed for every event, so those installations reported no telemetry at all; where direct egress happened to be open it succeeded by bypassing the configured policy.

The options are now resolved once per batch, the batch is requeued rather than sent when they cannot be resolved, and each request is capped at 20s so one stalled request no longer consumes the whole flush deadline.

Verified end to end with a proxy pointed at a closed port:

| | `telemetry-events.ndjson` after flush |
|---|---|
| before | absent — drained, so the POST succeeded despite the unreachable proxy |
| after | present, 1 event queued — proxy honored, nothing sent |

Three unit tests added, each verified to fail without the corresponding source change.

### DIFF
--- a/src/core/server/fetch-guarded.ts
+++ b/src/core/server/fetch-guarded.ts
@@ -23,12 +23,17 @@
 import { buildFetchNetworkOptions } from '@/core/host/connectivity/network-config.ts';
 import type { FetchNetworkOptions } from '@/core/host/connectivity/types.ts';
 
+/**
+ * `networkOptions` is required so a new call site cannot silently skip the resolved
+ * proxy/TLS configuration. Pass `buildFetchNetworkOptions(url)` for the URL being fetched.
+ * `body` stays positional but explicit: pass `undefined` for bodyless requests.
+ */
 export function buildFetchInit(
   method: string,
   headers: Record<string, string>,
   timeoutMs: number,
-  body?: string,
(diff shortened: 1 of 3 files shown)


Krosovok (approved):
Since we are on it, can you please also make the `networkOptions` in `buildFetchInit` non-optional? There should be no other places now without it and it will help us to prevent it in the future.

vnaskos-sonar:
Done in 70bb513. `networkOptions` is now required on `buildFetchInit`.

You were right that every call site already passes it — `client.ts` (5 sites), the telemetry flush, and one test — so `tsc` came back clean without touching any of them.

One wrinkle worth flagging for review: `body` sat before `networkOptions` and TypeScript forbids a required parameter after an optional one, so `body?: string` became `body: string | undefined`. It stays positional and every existing call site already passes a value there (`undefined` for the bodyless GET), so nothing else changed. Added a short comment on the signature explaining why both are the way they are.

Also rebased onto master to pick up ff8ca2ec.

Krosovok:
Yes, I think we can keep the parameter order as-is for now.

PR #632, opened 2026-07-28
CLI-908 Replace `--skip-context` flag with `_SKIP_CONTEXT` env var

tomshafir-sonarsource:


src/commands/integrate/_common/context-augmentation.ts
```diff
@@ -32,6 +32,13 @@ import { SonarQubeClient } from '@/core/server/client.ts';
 import type { IntegrationStateAttribute } from '@/core/state/state.ts';
 import { discreetSuccess, type OutputChannel, print, text, warn, withSpinner } from '@/core/ui';
 
+export const ENV_SKIP_CAG = '_SKIP_CAG';
```
georgii-borovinskikh-sonarsource: the name looks weird:
1) does it start with _ to indicate that it's internal?
2) some sort of indication that this is our variable would be nice, like `_SQCLI_SKIP_CAG`, maybe with a combination of words like 'internal', 'dev' or 'test'

georgii-borovinskikh-sonarsource (approved):

PR #633, opened 2026-07-29
CLI-859 Add project_uuid to CLI telemetry events

vnaskos-sonar:
## What

Reports `project_uuid` on the `CliCommandExecuted` telemetry event: SonarQube's legacy internal project id (`projects.uuid`, from `GET /api/navigation/component`). Despite the name it is **not** an RFC-4122 UUID. It behaves identically on SonarQube Cloud and Server.

`CliAnalysisCompleted` and `CliIntegrationConfigured` intentionally do **not** carry the field — every event already shares an `invocation_id`, so they join to the command event on it. One field, one write site.

`project_uuid` is `null` for commands that never resolve a project, for `integrate --global` and project-less integrations, and when resolution fails.

## How

`src/core/telemetry/project-uuid.ts` owns two halves:

- **`resolveProjectUuid(auth, projectKey)`** — cache-then-API. Permanent disk cache (`project-uuid-cache.json`) keyed by `serverUrl::projectKey` rather than by auth fingerprint, since the id belongs to the project, not the caller. A resolved-but-empty response is cached as `null`; transient failures are not cached, so the next run retries. Never rejects, and short-circuits when telemetry is disabled.
- **`noteProject()` / `currentProjectUuid()`** — the project is a per-invocation fact, so it lives in a module-level slot instead of being passed down call signatures. Writes are synchronous and do no I/O; resolution happens once per process, memoized, in the `postAction` hook — off the command's critical path.

`noteProject` is called wherever a project key resolves and auth is in hand (auth supplies the server URL, and the anonymous hook commands resolve it inside the handler):

| Site | Covers |
| --- | --- |
| `analyze/sqaa-auth.ts` — `resolveCloudAuthAndProject` | `analyze`, `analyze agentic`, `verify` |
| `analyze/dependency-risks.ts` | `analyze dependency-risks` |
| `remediate/index.ts` | `remediate` |
| `hook/agent-post-tool-use.ts`, `hook/codex-post-tool-use.ts`, `hook/git-pre-commit.ts` | the three project-bearing hooks |
| `framework/features/install-integration.ts` | every `integrate` subcommand (from the first non-empty `attrs.projectKey`) |

`sonar run mcp` deliberately opts out (comment in place): it starts a long-running server, so its `CliCommandExecuted` only fires at exit, if ever.

## Notes for reviewers

- **`SonarQubeClient.getSafe` gains an optional `timeoutMs`** (default unchanged). The resolver passes 3s instead of the 30s default because resolution runs in `postAction`, after the command has printed its result — a slow server would otherwise look like the CLI hanging at exit, worst in `hook git-pre-commit` where it would stall a commit. The cap is enforced by the fetch's own `AbortSignal`: bounding only the `await` would leave the socket open and keep the process alive anyway.
- **`emitSqaaAnalysisTelemetry` is now wrapped in try/catch**, mirroring `emitScaAnalysisTelemetry`. This is unrelated to `project_uuid` and is included because it is a small safety gap in the same code: it was the only `emit*` path with no exception guard anywhere in its call chain, so a telemetry failure could turn a successful SQAA analysis into a reported command failure.
- **The analysis→command join is 1→N.** Bare `sonar analyze` emits two `CliAnalysisCompleted` rows for one `CliCommandExecuted`, so per-project aggregates must count on the analysis side or `DISTINCT` on `invocation_id`.
- `project_uuid` is written on the last event of a run, so a hard `process.exit()` before `postAction` would lose it for the whole invocation. No current path does this with a project in scope.
- Tests must call `resetProjectUuidContextForTests()` in `beforeEach` — module state outlives individual tests in the same file.

## Testing

Unit tests cover the resolver (cache hit/miss, transient vs confirmed-absent caching, telemetry-disabled short-circuit) and the ambient context (memoization to one API call per process, a later note superseding an earlier one, never rejecting on failure).

Integration tests assert `project_uuid` on `CliCommandExecuted` — and `invocation_id` equality with the analysis event — across the analyze, hook, remediate and integrate paths, plus `null` for a secrets-only `git-pre-commit`.

Also verified manually against SonarQube Cloud on every note site except `remediate`, including auto-detection with no explicit `--project`.

### DIFF
--- a/src/core/framework/features/install-integration.ts
+++ b/src/core/framework/features/install-integration.ts
@@ -30,6 +30,7 @@ import type {
 } from '@/core/state/state.ts';
 import { loadState, saveState } from '@/core/state/state-repository.ts';
 import { emitIntegrationConfiguredTelemetry } from '@/core/telemetry/integrate-telemetry.ts';
+import { noteProject } from '@/core/telemetry/project-uuid.ts';
 import { text, warn } from '@/core/ui';
 
 import { renderCompletionSummary } from './completion-summary.ts';
@@ -142,6 +143,12 @@ export async function installIntegration<TOptions>({
 
     renderCompletionSummary(integration, installedFeatures, removedFeatures);
 
+    // Publish the project for `project_uuid` on CliCommandExecuted. Project scope only:
+    // `--global` installs have no project key recorded, which is why they report null.
+    if (scope === 'project' && auth) {
+      noteProject(auth, findProjectKeyFromFeatures(installedFeatures));
+    }
+
     const stateSaved = saveInstalledFeatures(state);
     if (stateSaved) {
       await emitIntegrationConfiguredTelemetry({
@@ -185,6 +192,23 @@ export function makeContext(
     attrs,
     resolvedDependencies: new Map(),
   };
+}
+
+/**
+ * Project key for telemetry `project_uuid`: the first `attrs.projectKey` recorded on any
+ * installed feature, or undefined when none carries one.
+ *
+ * `integrate git` records `{ projectKey: options.project ?? null }`, and the agent
+ * integrations record the discovered key on their SQAA/CAG features — so the non-empty-string
+ * check is what makes a secrets-only git install (which records an explicit `null`) report
+ * `project_uuid: null` rather than a bogus value.
+ */
+function findProjectKeyFromFeatures(features: InstalledIntegrationFeature[]): string | undefined {
+  for (const feature of features) {
+    const value = feature.attrs?.projectKey;
+    if (typeof value === 'string' && value.length > 0) return value;
+  }
+  return undefined;
 }
 
 /**
--- a/src/core/framework/features/install-integration.ts
+++ b/src/core/framework/features/install-integration.ts
@@ -146,7 +146,7 @@ export async function installIntegration<TOptions>({
     // Publish the project for `project_uuid` on CliCommandExecuted. Project scope only:
     // `--global` installs have no project key recorded, which is why they report null.
     if (scope === 'project' && auth) {
-      noteProject(auth, findProjectKeyFromFeatures(installedFeatures));
+      noteProject(auth, projectKeyFromAttrs(attrs));
     }
 
     const stateSaved = saveInstalledFeatures(state);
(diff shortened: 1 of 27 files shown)


src/core/telemetry/project-uuid.ts
```diff
@@ -0,0 +1,132 @@
...
+      typeof parsed === 'object' &&
+      'entries' in parsed &&
+      typeof (parsed as ProjectUuidCacheFile).entries === 'object'
+    ) {
+      return parsed as ProjectUuidCacheFile;
+    }
+  } catch {
+    return { entries: {} };
+  }
+  return { entries: {} };
```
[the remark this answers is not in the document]
vnaskos-sonar: Fixed in f2e5fdc5. `readDiskCache` now reads `entries` as `unknown` and explicitly rejects a `null`/non-object value (`typeof null === 'object'`) before returning, so a `{ "entries": null }` file falls back to an empty cache instead of throwing on the later `in` check. Added a regression test ("treats a cache file with a null entries value as empty").

src/core/telemetry/project-uuid.ts
```diff
@@ -0,0 +1,241 @@
...
+    return { entries: {} };
+  }
+  try {
+    const parsed: unknown = JSON.parse(readFileSync(path, 'utf-8'));
+    if (parsed !== null && typeof parsed === 'object' && 'entries' in parsed) {
+      // `entries` is untyped at runtime: guard against `null` (typeof null ===
+      // 'object') and non-object values so callers can safely index into it.
+      const entries: unknown = parsed.entries;
+      if (entries !== null && typeof entries === 'object') {
+        return parsed as ProjectUuidCacheFile;
```
kirill-knize-sonarsource, line 75: `project_uuid: null` (set in `src/core/telemetry/index.ts:75`) gets stripped from the outbound payload by the null-stripping replacer at `src/core/telemetry/telemetry-events.ts:254`. Verified: `JSON.stringify({project_uuid:null}, nullReplacer)` drops the key. So "not project-scoped" and "resolution failed" are indistinguishable in the warehouse, contradicting [CLI-859 AC #3](https://sonarsource.atlassian.net/browse/CLI-859) ("explicit empty value, not an ambiguous null"). Use `''` instead, like `details: ''` in `sca-analysis-telemetry.ts`?

src/core/framework/features/install-integration.ts
```diff
@@ -142,6 +143,12 @@ export async function installIntegration<TOptions>({
 
     renderCompletionSummary(integration, installedFeatures, removedFeatures);
 
+    // Publish the project for `project_uuid` on CliCommandExecuted. Project scope only:
+    // `--global` installs have no project key recorded, which is why they report null.
+    if (scope === 'project' && auth) {
+      noteProject(auth, findProjectKeyFromFeatures(installedFeatures));
```
kirill-knize-sonarsource: only reads `toInstall` features. If a run only removes features (`toInstall` empty, `toRemove` non-empty), this returns `undefined` and `project_uuid` reports `null`, even though the removed feature's `attrs.projectKey` is known. Intentional, or should it fall back to `toRemove` too?

kirill-knize-sonarsource (commented):
Some questions.

kirill-knize-sonarsource (approved):

src/core/telemetry/project-uuid.ts
vnaskos-sonar: This was on purpose left out, it will cause more trouble rather than good while extracting data from the tables and would blow up the complexity of the implementation. I will update the specs accordingly.

PR #637, opened 2026-07-29
CAG-910 Add E2E tests for shell compression hooks

michael-jabbour-sonarsource:
Improve E2E coverage on newly introduced CAG hooks: forwarding Claude hook invocations through `sonar context` and compressing Gradle-shaped Bash output. The compression assertions are expected to fail until #649 is merged.

Tests added:

- `cag-passthrough.test.ts` forwards `sonar context __hook Claude --help` to the real CAG binary. Rationale: prove the CLI wrapper accepts and forwards the hidden Claude hook entry point instead of treating it as an invalid subcommand.

- `cag-passthrough.test.ts` sends a Claude `PostToolUse` Bash payload for `./gradlew build` to the real CAG binary. Rationale: provide a lightweight contract test for Gradle compression without invoking Claude Code; it asserts `hookSpecificOutput.updatedToolOutput`, preserved tool-response fields, removed noisy task lines, and the `sonar context distillate restore` hint. Note that I also changed these tests to seed the recorded CAG/Vortex context with an explicit fake SonarQube Cloud server. This keeps non-help sonar context passthroughs authenticated against the same fake connection and lets us exercise CAG functionality that requires Cloud context, project access, and entitlement checks, such as `__hook Claude`.

- `claude-integration.test.ts` installs the CAG hook through post-update, creates a fake `./gradlew build`, and has real Claude Code run it. Rationale: prove the generated Claude settings and launcher script work end to end, so Claude receives compressed Gradle output and the restore hint from the installed hook.

----
## Summary by Gitar

- **CI / Test Infrastructure:**
    - Added a workflow step on Windows to map `api.localhost` to loopback for E2E tests in `.github/workflows/build.yml`


michael-jabbour-sonarsource:
This change is part of the following stack:

- #569
    - #637 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

kirill-knize-sonarsource (approved):

tests/e2e/context/cag-passthrough.test.ts
```diff
@@ -66,14 +104,61 @@ describe('sonar-context-augmentation passthrough behaviors (offline, real binary
...
+        hook_event_name: 'PostToolUse',
+        tool_name: 'Bash',
+        tool_input: { command: './gradlew build', description: 'Run Gradle build' },
+        tool_response: {
+          stdout: buildCompressibleGradleStdout(),
+          stderr: '',
+          interrupted: false,
+          isImage: false,
+          noOutputExpected: false,
+        },
```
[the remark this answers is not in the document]
michael-jabbour-sonarsource: #649 was already merged at this point.

tests/e2e/context/_helpers.ts
```diff
@@ -53,6 +54,9 @@ export const STALE_CLI_VERSION = '0.0.1';
 export const STALE_SKILL_VERSION = '0.0.0';
 export const SEEDED_PROJECT_KEY = 'offline-test-project';
 export const SEEDED_ORG_KEY = 'offline-test-org';
+// Not 'sonarsource': the fake CAG stub renders an extra dogfooding-tools section
+// in the skill only for that exact org, which would pollute unrelated assertions.
+export const ALLOWLISTED_CAG_ORG_KEY = 'denis-troller-sonar';
```
michael-jabbour-sonarsource: @kirill-knize-sonarsource I noticed that you changed this from "sonarsource" to "denis-troller-sonar". For my own curiosity, could you elaborate a bit on what assertions where polluted with "sonarsource"? I remember that the tests were passing just fine...

michael-jabbour-sonarsource: Was the change somehow motivated by this Gitar finding? https://github.com/SonarSource/sonarqube-cli/pull/637#discussion_r3720884887

michael-jabbour-sonarsource:
@kirill-knize-sonarsource This ticket is ready for review from my side. I also did an intentional failing test run in #669 to ensure that the test is working as expected.

.github/workflows/build.yml
```diff
@@ -459,6 +459,18 @@ jobs:
         if: matrix.os == 'linux'
         run: SONARQUBE_CLI_USE_COVERAGE=1 bun run test:integration
 
+      # Workaround for CAG-933: CAG deduces the fake SonarQube Cloud API endpoint
```
michael-jabbour-sonarsource, line 462: @kirill-knize-sonarsource Alternatively, I am also happy to drop this workaround and disable the test on Windows until CAG-933 is fixed. Let me know your thoughts

kirill-knize-sonarsource: Do you plan to drop it before the release?

michael-jabbour-sonarsource: To be clear, as discussed, I tend to prefer the current workaround as we gain more confidence by making sure the test runs on more platforms...

tests/e2e/context/_helpers.ts
michael-jabbour-sonarsource: I am trying to restore the sonarsource organization in c0a347fad780c820f3db7e7981c2506919d87eb1. I don't agree with fixing the Gitar finding by changing to the denis org. I think that the same reasoning applied by Gitar there applies to denis org as well:

> ALLOWLISTED_CAG_ORG_KEY = 'sonarsource' couples the test fixtures to a value baked into the CAG binary's allowlist. If the binary's allowlist changes, these E2E tests silently break with no obvious pointer to the cause. Consider documenting the coupling with a comment referencing where the binary defines this allowlist so future maintainers can keep them in sync.

The allowlist is temporary anyway. I don't expect to ever remove SonarSource from there. We may only remove the whole gating, and when that happens, SonarSource would remain a valid test org, and it is more intuitive IMO

PR #639, opened 2026-07-29
CLI-901 Unify Vortex features for Claude

tomshafir-sonarsource:
---
# Summary
* Unifies Claude Code’s SQAA hooks/instructions and Context Augmentation under one project-scoped `vortex` container with a single installation prompt and entitlement check.
* Refactors shared SQAA/CAG declarations into reusable subfeatures while preserving other agent integrations.
* Updates SQAA and Context Augmentation runtime lookup to recognize the new Vortex container while remaining backward-compatible with legacy standalone records.
* Adds rollback protection to post-update migration, restoring deprecated feature records if Vortex installation fails.
* Updates integration and E2E tests for Vortex installation, migration, failure recovery, and runtime lookup.


src/core/host/post-update.ts
```diff
@@ -294,6 +296,37 @@ function migrateReplacedFeatures(
...
+      (feature) =>
+        feature.featureId === application.feature.id &&
+        feature.scope === application.scope &&
+        feature.targetRoot === application.targetRoot,
+    );
+    if (successorWasInstalled) {
+      continue;
+    }
+
+    const predecessorsToRestore = originalFeatures.filter(
```
georgii-borovinskikh-sonarsource, line 325: This only restores it in state - when can the replacement fail to install? Does the replacement itself produce side effects during the failed install? Do we intentionally ignore them when we restore the original?

src/core/server/client.ts
```diff
@@ -505,27 +505,31 @@ export class SonarQubeClient {
    * enable both capabilities when only one is available, which would fail later.
    */
   async hasVortexEntitlement(organizationKey?: string): Promise<VortexEntitlementStatus> {
-    if (!organizationKey || !isSonarQubeCloud(this.serverURL)) {
```
georgii-borovinskikh-sonarsource, line 508: god github diffs are awful - took me 5 minutes to understand that you just wrapped the method in try/catch. Though, was that necessary?

src/commands/integrate/_common/sqaa-entitlement.ts
```diff
@@ -85,3 +98,39 @@ export async function resolveSqaaSetup(params: ResolveSqaaSetupParams): Promise<
   // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
   return status === 'enabled' || status === 'over_consumption';
 }
+
+export const SQAA_INSTRUCTIONS_SUBFEATURE_ID = 'sqaa-instructions';
+const SQAA_INSTRUCTIONS_MARKER = 'sonarqube-agentic-analysis-protocol';
+
+/** End-of-turn SQAA instructions, written by each agent into its own rules format. */
+export function createSqaaInstructionsSubfeature<TOptions>(
```
georgii-borovinskikh-sonarsource: move feature declarations to ./features/

src/core/host/post-update.ts
georgii-borovinskikh-sonarsource: TODO: Document why the rollback logic here is correct

georgii-borovinskikh-sonarsource (approved):

src/core/server/client.ts
tomshafir-sonarsource: Yes, that try/catch used to be outside, handled by the caller, and it was not intuitive.

PR #664, opened 2026-08-04
CLI-940 Stop test runs leaking telemetry to the production endpoint

vnaskos-sonar:
## Problem

Test fixtures were being transmitted to the production telemetry endpoint. Production data contains `CliCommandExecuted` rows with `project_uuid=AYdemolegacy` — a value that exists only in two integration specs.

## Root cause

`isTelemetryEnabled(state)` was the single input to three different decisions: whether to collect events, whether to transmit them, and whether to enable error reporting.

That makes **"collect but do not transmit" impossible to express** — which is exactly what a telemetry test needs. A test has to enable telemetry to assert that an event was recorded, and doing so also arms the detached `flush-telemetry` worker that POSTs it. Seven integration specs are in that position, and none can use `__SQ_CLI_TELEMETRY_FLUSH__` as a workaround because that flag no-ops `storeEvent`, the very thing they assert on. Sentry leaked through the same gate.

## Change

A second, independent input: `resolveTelemetryEgress()` (`src/core/telemetry/egress.ts`), driven by `__SQ_CLI_TELEMETRY_EGRESS`. Unset means production; any other value means off. The fallthrough is fail-safe by design, so a misspelled override over-isolates rather than silently restoring production egress.

It is read in three places:

| Site | With egress off |
|---|---|
| `storeEvent` | skips `spawnFlushWorker()` — no detached process is created |
| `flushTelemetry` | returns before the drain — the queue is left on disk |
| `initSentry` | requires production egress — error reporting stays off |

`flushTelemetryEvents` still transmits unconditionally; its guards sit one level up in `flushTelemetry` so it remains directly callable by its unit tests.

## Enforcement

- ESLint `no-restricted-imports` confines `TELEMETRY_ENDPOINT`, `TELEMETRY_API_KEY` and `SENTRY_DSN` to the two modules that own the outbound call. Matched as a glob against the import string, so every spelling is covered (`@/core/…`, `./`, `../`, `../../`).
- `applyIsolatedSpawnEnv()` replaces duplicated overlay logic across the harness spawn paths and throws if egress is ever not `off`.
- Two guards: a unit test (no worker spawned, event still queued) and an integration test (queue written, never drained).
- Comments at the two hand-rolled CLI spawns in `install-scripts.test.ts`, where the `ISOLATED_CLI_SPAWN_ENV` spread is the only protection and no rule can catch its removal.

## For test authors

`withTelemetryEnabled()` is now safe on its own; the previous advice to pair it with `__SQ_CLI_TELEMETRY_FLUSH__=1` has been removed. Unit tests that need the spawn or drain to run clear the variable per file — 4 assertions in `telemetry.test.ts`, 22 in `sentry.test.ts`. The remaining pitfalls are documented in the telemetry section of `CLAUDE.md`.

## Known gap

The egress mode protects *processes*. It cannot stop a unit test that enables telemetry without isolating `SONAR_USER_HOME` from writing events into the developer's real queue, which their next genuine `sonar` command would then drain and transmit. That path is safe by convention today; making it structural (unconditional `SONAR_USER_HOME` in the test preload) is deliberately left out of this PR.


damien-urruty-sonarsource (approved):
LGTM. Another thing we could also do is introduce an env var to tweak the URL where events are posted. We could configure a fake server from the harness. This would also help us verify that events are properly posted by the worker, which I don't think we do today. But this can be done separately.

PR #666, opened 2026-08-04
CLI-905 Handle Vortex entitlement checks across CLI flows

Krosovok:
----
## Summary by Gitar

- **Vortex entitlement checks:**
    - Added `recheckVortexEntitlement` to re-query combined entitlement status and reinterpret runtime 403 errors
    - Created `vortexUnavailableHookMessage` and `vortexUnavailableCommandMessage` for context-aware user messaging
- **Hook & command integration:**
    - Updated agent and codex post-tool-use hooks to emit entitlement loss notices when analysis returns forbidden errors
    - Added 24-hour throttle in state for Vortex entitlement-loss hook warnings
    - Updated `sonar analyze agentic` flows to gracefully handle forbidden responses using entitlement re-checks


src/core/vortex/vortex-entitlement-notice.ts
```diff
@@ -0,0 +1,45 @@
+/*
```
georgii-borovinskikh-sonarsource, line 1: Files related to vortex entitlement notice should not be in `src/core/state`. This folder is now only responsible for generic state operations and the model

src/core/time/cooldown.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+export const ONE_DAY_MS = 24 * 60 * 60 * 1000;
+
+/**
+ * True when `isoTimestamp` is a valid past instant within `cooldownMs` of now.
+ * A missing, unparseable, or future timestamp is treated as "not in cooldown"
+ * so the guarded action is allowed to run.
+ */
+export function isWithinCooldown(isoTimestamp: string | undefined, cooldownMs: number): boolean {
```
georgii-borovinskikh-sonarsource, line 28: same, shouldn't be in `/state/`

src/core/vortex/entitlement.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Re-query the combined Vortex entitlement for the resolved connection. Used to
+ * reinterpret a runtime 403 (entitlement loss vs usage-limit exhaustion) without
+ * coupling to any command or output layer.
+ */
+export async function recheckVortexEntitlement(
+  auth: ResolvedAuth,
+): Promise<VortexEntitlementStatus> {
+  return new SonarQubeClient(auth.serverUrl, auth.token).hasVortexEntitlement(auth.orgKey);
```
georgii-borovinskikh-sonarsource, line 32: Is there a way to reuse the client from outside and not re-instantiate it?

src/commands/analyze/sqaa-analysis.ts
```diff
@@ -300,6 +300,9 @@ function recordFileFailure(
   error: Error,
 ): void {
   progress.update(fileIndex, 'failed');
+  if (isGlobalSqaaError(error)) {
```
georgii-borovinskikh-sonarsource: Not a fan of this move: I don't think the file handler should be the one setting the global error when we could do it once at the beginning

src/commands/analyze/sqaa-run.ts
```diff
@@ -156,10 +176,14 @@ export async function fetchSingleFileReport(
...
   } catch (err) {
     const error = err as Error;
-    return {
-      report: singleFileFailureReport(filePath, error.message, displayDepth),
-      error,
-    };
+    const report = singleFileFailureReport(filePath, error.message, displayDepth);
+    if (isGlobalSqaaError(error)) {
+      report.globalError = {
+        kind: error instanceof SqaaForbiddenError ? 'forbidden' : 'unavailable',
```
georgii-borovinskikh-sonarsource: this `? 'forbidden' : 'unavailable'` expression is duplicated, introduce a helper

src/commands/analyze/sqaa-run.ts
```diff
@@ -108,6 +113,23 @@ async function finishSqaaRun(
...
+ * Resolve which of the two 403 causes (entitlement lost vs. usage limit) actually
+ * happened and replace the report's generic 403 text with that resolved message,
+ * mirroring what the text-mode branches already show via `printVortexUnavailable`.
+ */
+async function enrichForbiddenGlobalError(
+  report: SqaaJsonReport,
+  auth: ResolvedAuth,
+): Promise<void> {
+  if (!report.globalError) return;
+  const status = await recheckVortexEntitlement(auth);
```
georgii-borovinskikh-sonarsource, line 139: what if it's intermittent and the status re-check return no error?

src/commands/analyze/sqaa-run.ts
```diff
@@ -274,7 +305,11 @@ export async function runSqaaAnalysisOnFiles(
     const { result: tally, durationMs } = await timed(() =>
       runSqaaAnalysesTallyForResolved(files, allPaths, resolved, branch, wireDepth, displayDepth),
     );
-    printJsonReport(tally, ignored, allPaths, repoRoot, displayDepth);
+    const report = buildJsonReport(tally, ignored, allPaths, repoRoot, displayDepth);
+    if (tally.globalError instanceof SqaaForbiddenError) {
```
georgii-borovinskikh-sonarsource: probably out of scope of this PR, but it would be great to have common error handling done centrally and not called by every type of sqaa run separately

georgii-borovinskikh-sonarsource: actually we should improve duplication of 

```
const report = buildJsonReport(tally, ignored, allPaths, repoRoot, displayDepth);
if (tally.globalError instanceof SqaaForbiddenError) {
  await enrichForbiddenGlobalError(report, options.auth);
}
print(JSON.stringify(report, null, 2));
```

tests/integration/specs/analyze/analyze-sqaa.test.ts
```diff
@@ -2267,6 +2269,76 @@ describe('analyze agentic — API error codes', () => {
   );
 });
 
+describe('analyze agentic — SQAA 403 (Vortex unavailable)', () => {
```
georgii-borovinskikh-sonarsource, line 2281: Feedback from agent

So a bug specifically in the runSqaaAnalysisOnExplicitFiles or runSqaaAnalysisOnFiles 403-handling branches (e.g. someone forgets enrichForbiddenGlobalError when copy-pasting, or breaks the printVortexUnavailable call in the text-mode change-set path) would pass the whole suite. Given how much this PR leans on that copy-pasted 4-way duplication, I'd add at least one 403 case that forces multi-file-explicit or change-set mode.

tests/integration/specs/hook/hook-agent-post-tool-use.test.ts
```diff
@@ -90,8 +99,67 @@ describe('sonar hook claude-post-tool-use', () => {
...
+      });
+
+      expect(result.exitCode).toBe(0);
+      expect(result.stdout.trim()).toBe('');
+    },
+    { timeout: 15000 },
+  );
+
+  it(
+    'exits 0 and shows the usage-limit message every run on an over_consumption 403',
```
georgii-borovinskikh-sonarsource, line 142: From agent:

"shows the usage-limit message every run" doesn't call run twice
In both hook-agent-post-tool-use.test.ts and hook-codex-post-tool-use.test.ts, the test titled "exits 0 and shows the usage-limit message every run on an over_consumption 403" only calls harness.run(...) once. It proves the message appears on a fresh state, but not that it's un-throttled across repeated runs — which is the entire point of the title (contrasting with the not_entitled case, which is throttled and does get a dedicated "stays silent when warned within 24h" test). Either rename it to something like "shows the usage-limit message on a fresh run" or actually assert the no-throttle behavior — e.g. run twice and check the message appears both times, or assert state.config.vortexEntitlementLossNotice stays unset after the run.

georgii-borovinskikh-sonarsource (approved):
Functionally correct, some comments regarding code organization and tests. You can merge without another reivew once all are resolved.

src/commands/analyze/sqaa-run.ts
Krosovok: It would give a generic error:
`'Vortex analysis is temporarily unavailable. Please try again later.'`
which would fit for most scenarios I think.

src/core/vortex/entitlement.ts
Krosovok: There is not much to reuse and auth is already passed around as a part of `SqaaBatchRunOptions`. Reusing it to group submitting SQAA and error-handling invocations doesn't make a ton of sense to me.

PR #667, opened 2026-08-04
CLI-933 Allow contributors to declare Alpha commands

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Commands:**
    - Added `alpha()` method to `SonarCommand` to declare alpha commands conditioned on `SONARQUBE_CLI_ALPHA`
    - Appended `[ALPHA]` tags to alpha command descriptions and summaries
    - Grouped alpha commands together in custom root help output and warned users on execution

### DIFF
--- a/src/commands/sonar-command.ts
+++ b/src/commands/sonar-command.ts
@@ -20,8 +20,8 @@
 
 // SonarCommand — Commander Command subclass with built-in error handling and auth support
 
-import type { CommandOptions } from 'commander';
-import { Command } from 'commander';
+import type { CommandOptions, Option } from 'commander';
+import { Command, Help } from 'commander';
 
 import type { ResolvedAuth } from '@/core/auth/auth-resolver.ts';
 import { resolveAuth } from '@/core/auth/auth-resolver.ts';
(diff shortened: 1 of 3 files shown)


src/commands/sonar-command.ts
```diff
@@ -106,6 +110,58 @@ export class SonarCommand extends Command {
...
+   * Mark this command as alpha. Alpha commands are registered only when the
+   * SONARQUBE_CLI_ALPHA environment variable is set.
+   */
+  alpha(): this {
+    if (this._isAlpha) {
+      return this;
+    }
+
+    this._isAlpha = true;
+    if (process.env[ALPHA_ENV_VAR] === undefined) {
```
vnaskos-sonar: Currently `SONARQUBE_CLI_ALPHA=false` will be treated as alpha enabled.

For consistency it could be checked in the same way as the env vars, `=== '1'` instead of `=== undefined`.

vnaskos-sonar (approved):
Looks and works great.

Possibly expected but raising it for awareness, currently only top level alpha commands are grouped together on the --help. For subcommands it will look like this 

Image

src/commands/sonar-command.ts
damien-urruty-sonarsource: Good catch! I made it so we support values 'true' and '1'

damien-urruty-sonarsource:
@vnaskos-sonar we only have grouping for the top-level help menu. We don't do anything special for sub-help menus, and commands are displayed all together, in the declaration order. I don't know if we need to to something about that, @cliffordgoh-sonar do you have any opinion? Should we enforce alpha sub-commands to appear at the bottom?

cliffordgoh-sonar:
> @vnaskos-sonar we only have grouping for the top-level help menu. We don't do anything special for sub-help menus, and commands are displayed all together, in the declaration order. I don't know if we need to to something about that, @cliffordgoh-sonar do you have any opinion? Should we enforce alpha sub-commands to appear at the bottom?

Yes, lets enforce that for subcommand as well, to maintain predictability across CLI
We can let contributor change the order within the bottom group if they wish to (if they have >1 alpha subcommand)

PR #682, opened 2026-08-12
CLI-935 Extract createCommandTree from module-level registration

damien-urruty-sonarsource:
## Summary
- Move command registration from a module-level `COMMAND_TREE` side effect into `createCommandTree()`
- Update CLI entrypoint, docs generator, and update-notification tests to call the factory
- No behavior change — prepares for per-invocation Private Beta gating in a follow-up PR

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,9 +34,9 @@ bun run test:e2e          # end-to-end tests
 
 ## Commands
 
-Each command lives in `src/commands/`. The command tree is defined in `src/commands/command-tree.ts` and the entry point is `src/index.ts`.
+Each command lives in `src/commands/`. The command tree is built by `createCommandTree()` in `src/commands/command-tree.ts` and the entry point is `src/index.ts`.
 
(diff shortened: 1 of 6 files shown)


vnaskos-sonar (approved):
👍 

Just a minor comment from claude 
```
Stale comment left behind in src/commands/sonar-command.ts:152-157 — the updateNotifier getter's doc comment still says "command-tree.ts builds the whole tree as top-level side effects when the module loads, registering every .showUpdateNotification() call into this instance in the process." That's no longer true: registration now happens inside createCommandTree(), invoked per-call, not as a module-load side effect. This file wasn't touched by the PR, so it was probably just missed. Low severity (doesn't affect behavior), but worth a one-line fix since it actively misleads about how the mechanism works now
```

damien-urruty-sonarsource:
Updated the stale \`updateNotifier\` comment in \`src/commands/sonar-command.ts\` to describe \`createCommandTree()\` instead of module-load side effects.

PR #688, opened 2026-08-12
CLI-935 Cache Private Beta LaunchDarkly flag decisions for 2 hours

damien-urruty-sonarsource:
## Summary
- Add a 12-hour on-disk cache for Private Beta flag decisions under `~/.sonar/sonarqube-cli/launch-darkly/`
- Reuse fresh cache entries to avoid LaunchDarkly round-trips on every CLI invocation
- Treat missing LaunchDarkly keys as cached `false`

----
## Summary by Gitar

- **Caching:**
    - Shorten Private Beta flag decision cache TTL from 12 hours to 2 hours
    - Harden cache entry validation against corrupted or malformed on-disk files


src/core/launch-darkly/cache.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/** Persists boolean decisions for the identity, replacing any previous entry. */
+export function writeFlagDecisions(
+  identity: FeatureFlagIdentity,
+  flags: Record<string, boolean>,
+  clientSideId: string,
+  nowMs: number = Date.now(),
+): void {
+  const cache = readCacheFile();
```
vnaskos-sonar, line 151: This read-then-write isn't atomic. Two sonar commands running at once could both read the same snapshot and the second write mess up the first's entry. Probably fine since it's just a cache and worst case is one extra LD fetch next time.

src/core/launch-darkly/cache.ts
```diff
@@ -0,0 +1,158 @@
...
+function isCacheEntry(value: unknown): value is CacheEntry {
+  if (typeof value !== 'object' || value === null) {
+    return false;
+  }
+  const record = value as { fetchedAt?: unknown; flags?: unknown };
+  return (
+    Number.isFinite(record.fetchedAt) &&
+    typeof record.flags === 'object' &&
+    record.flags !== null &&
+    !Array.isArray(record.flags)
```
vnaskos-sonar: This checks flags is an object but never checks the values inside are actually booleans.

A corrupted or hand-edited cache file could have "cli.beta.private": "yes" and we'd happily trust it as a real decision. Can we add something like `Object.values(record.flags).every(v => typeof v === 'boolean')` here? cheap fix, matches the === true discipline you already use in toBooleanFlagMap.

vnaskos-sonar (approved):
I think it could/should be less than 12 hours, so we can have more control over it. Maybe 1-3 hours?

src/core/launch-darkly/cache.ts
damien-urruty-sonarsource: Agreed — leaving it as-is. Worst case is a lost cache entry and one extra LaunchDarkly fetch.

src/core/launch-darkly/cache.ts
damien-urruty-sonarsource: Good catch — `isCacheEntry` now requires every flag value to be a boolean, otherwise the entry is ignored. Done in c8479fd3.

damien-urruty-sonarsource:
@vnaskos-sonar Re the TTL: switched to **2 hours** in c8479fd3 so we keep more control over enable/disable without hitting LaunchDarkly on every invocation.

PR #697, opened 2026-08-18
CLI-978 Forward custom user home to the MCP server declaration

damien-urruty-sonarsource:
## Summary
- Bake a custom `SONAR_USER_HOME` into the generated MCP server `env` block so `sonar run mcp` can load the same CLI state the user authenticated against
- Leave default `~/.sonar` installs unchanged (no `env` key); re-running integrate refreshes an older entry that was missing it
- Shared `getMcpConfig()` writer, so Claude, Copilot, Cursor, Codex, and Antigravity all pick this up

### DIFF
--- a/tests/integration/specs/integrate/codex.test.ts
+++ b/tests/integration/specs/integrate/codex.test.ts
@@ -26,6 +26,7 @@
 import { isAbsolute } from 'node:path';
 
 import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
+import { parse as parseToml } from 'smol-toml';
 
 import { CONTEXT_AUGMENTATION_FEATURE_ID } from '../../../../src/commands/integrate/_common/features/context-augmentation-feature';
 import {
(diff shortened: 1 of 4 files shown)


tests/integration/specs/integrate/codex.test.ts
```diff
@@ -294,6 +297,33 @@ describe('integrate codex', () => {
       { timeout: 30000 },
     );
 
+    it(
+      'forwards SONAR_USER_HOME into [mcp_servers.sonarqube.env] and refreshes a prior entry',
```
tomshafir-sonarsource, line 302: Nit: The Codex and Cursor tests can't actually fail on a wrong path. `harness.sonarUserHome.path` is `<userHome>/.sonar`, and the harness already points `HOME`/`USERPROFILE` at `userHome`, so the custom home is identical to the default.

tomshafir-sonarsource (approved):
Looks good! Left one comment.

tests/integration/specs/integrate/codex.test.ts
damien-urruty-sonarsource: Addressed in the last commit

PR #699, opened 2026-08-19
CLI-982 Introduce CommandInvocationContext for command handlers

damien-urruty-sonarsource:
## Summary
- Introduce `CliContext` / `CliAuthenticatedContext` (classes in `cli-context.ts`) with resolved `isAlpha` / `isBeta` for the current command invocation
- `anonymousAction()` passes `CliContext`; `authenticatedAction()` passes `CliAuthenticatedContext` (adds `auth`)
- Migrate authenticated entry handlers to take the context; internal helpers that only need credentials keep `ResolvedAuth`
- Leaves room for CLI-958 / CLI-959 to add agent-session methods on `CliContext`


src/commands/analyze/secrets.ts
```diff
@@ -193,8 +194,9 @@ interface ScanDisplayContext {
 
 export async function analyzeSecrets(
   options: AnalyzeSecretsOptions,
-  auth: ResolvedAuth,
+  ctx: CliAuthenticatedContext,
```
damien-urruty-sonarsource: I'm not too sure about the names CliContext and CliAuthenticatedContext. It's maybe more the context about a specific command invocation? `CommandInvocationContext` and `CommandAuthenticatedInvocationContext`? A bit longer, let me know if you have any idea

vnaskos-sonar (approved):
👍

PR #700, opened 2026-08-20
CLI-983 Allow contributors to stage command options as alpha or beta

damien-urruty-sonarsource:
## Summary
- Add `SonarOption.stage()` so contributors can mark options Alpha, Open Beta, or Private Beta
- Hide unentitled options from help and treat them as unknown; collect Private Beta flag keys from options
- Include Open Beta options in generated docs with a beta badge; omit Alpha and Private Beta options


src/commands/sonar-command.ts
```diff
@@ -140,6 +233,13 @@ class SonarHelp extends Help {
     }
     return super.formatItemList(heading, items, helper);
   }
+
+  override optionDescription(option: Option): string {
```
vnaskos-sonar, line 237: Small gap: the root `sonar --help` menu (root-help.ts) reads option text directly instead of through the new SonarHelp.optionDescription(), so it skips the [ALPHA]/[BETA] tagging.

Example: if someone later stages a root-level option like
`COMMAND_TREE.addOption(new SonarOption('--foo', 'Do the thing').stage(Stage.Beta()))`
then sonar foo --help would correctly show Do the thing [BETA], but the root sonar --help menu would show plain Do the thing — no tag.

Not an issue today (the only root options are -h/--help and -v/--version, never staged), just flagging in case a root option gets staged down the line.

vnaskos-sonar (approved):
👌

src/commands/sonar-command.ts
damien-urruty-sonarsource: Fixed

PR #703, opened 2026-08-21
CLI-997 Send SonarQube Cloud enterprise UUID to LaunchDarkly

damien-urruty-sonarsource:
## Summary
- Persist SonarQube Cloud `enterpriseUuid` on the connection (and identity disk cache), resolved via `/organizations/organizations` then `/enterprises/enterprise-organizations`
- Include an `enterprise` LaunchDarkly context for Private Beta flag evaluation when the UUID is known; orgs not in an enterprise still evaluate with user + organization
- Existing Cloud connections with user + org but no enterprise field are enriched once; transient lookup failures are not stored

### DIFF
--- a/src/core/telemetry/identity-fetch.ts
+++ b/src/core/telemetry/identity-fetch.ts
@@ -18,7 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-// Fetches and disk-caches server-side identity fields (user/org/installation UUIDs) for a resolved auth.
+// Fetches and disk-caches server-side identity fields (user/org/enterprise/installation UUIDs) for a resolved auth.
 
 import { createHash } from 'node:crypto';
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
@@ -31,15 +31,20 @@ import { resolveFromEndpoint } from '@/core/server/sonarcloud-region.ts';
 import { getTelemetryDir } from '../config-constants.ts';
 import type { AuthConnection, ServerType } from '../state/state.ts';
 
+const ORGANIZATIONS_ENDPOINT = '/organizations/organizations';
+const ENTERPRISE_ORGANIZATIONS_ENDPOINT = '/enterprises/enterprise-organizations';
+
 export interface TelemetryIdentity {
   user_uuid: string | null;
   organization_uuid_v4: string | null;
+  enterprise_uuid: string | null;
   sqs_installation_id: string | null;
 }
 
 export const EMPTY_IDENTITY: TelemetryIdentity = {
   user_uuid: null,
   organization_uuid_v4: null,
+  enterprise_uuid: null,
   sqs_installation_id: null,
 };
 
@@ -47,6 +52,7 @@ export function identityFromConnection(conn: AuthConnection | undefined): Teleme
   return {
     user_uuid: conn?.userUuid ?? null,
     organization_uuid_v4: conn?.organizationUuidV4 ?? null,
+    enterprise_uuid: conn?.enterpriseUuid ?? null,
     sqs_installation_id: conn?.sqsInstallationId ?? null,
   };
 }
@@ -67,6 +73,11 @@ function connectionUserUuidResolved(conn: AuthConnection | undefined): boolean {
   return conn?.userUuid !== undefined;
 }
 
+/** Login persists `enterpriseUuid` (string or null); undefined means we have not tried yet. */
+function connectionEnterpriseUuidResolved(conn: AuthConnection | undefined): boolean {
+  return conn?.enterpriseUuid !== undefined;
+}
+
 export function needsIdentityEnrichment(
   identity: TelemetryIdentity,
   connectionType: ServerType,
@@ -75,12 +86,16 @@ export function needsIdentityEnrichment(
   if (!isIdentityCompleteForConnection(identity, connectionType)) {
     return true;
   }
-  return !identity.user_uuid && !connectionUserUuidResolved(conn);
+  if (!identity.user_uuid && !connectionUserUuidResolved(conn)) {
+    return true;
+  }
+  return connectionType === 'cloud' && !connectionEnterpriseUuidResolved(conn);
 }
 
 interface CacheEntry {
   userUuid?: string | null;
   organizationUuidV4?: string | null;
+  enterpriseUuid?: string | null;
   sqsInstallationId?: string | null;
 }
 
@@ -91,6 +106,7 @@ interface IdentityCacheFile {
 interface IdentityFetchPlan {
   user: boolean;
   org: boolean;
+  enterprise: boolean;
   sqs: boolean;
 }
 
@@ -104,6 +120,7 @@ export function mergeIdentity(
   return {
     user_uuid: partial.user_uuid ?? base.user_uuid,
     organization_uuid_v4: partial.organization_uuid_v4 ?? base.organization_uuid_v4,
+    enterprise_uuid: partial.enterprise_uuid ?? base.enterprise_uuid,
     sqs_installation_id: partial.sqs_installation_id ?? base.sqs_installation_id,
   };
 }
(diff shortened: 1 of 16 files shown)


src/core/telemetry/identity-fetch.ts
```diff
@@ -194,22 +228,42 @@ async function fetchUserUuid(client: SonarQubeClient): Promise<FieldFetchResult>
...
+    const { response, value } = await client.getSafe<Array<{ enterpriseId?: string }>>(
+      ENTERPRISE_ORGANIZATIONS_ENDPOINT,
+      { organizationId },
+      resolveFromEndpoint(serverUrl, ENTERPRISE_ORGANIZATIONS_ENDPOINT),
     );
     if (!response.ok) {
       return { value: null, resolved: false };
     }
-    return { value: value?.[0]?.uuidV4 ?? null, resolved: true };
+    return { value: value?.[0]?.enterpriseId ?? null, resolved: true };
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: Not applying this. On SonarQube Cloud the enterprise mapping endpoint returns 200 in both cases we care about (org in an enterprise, or empty list when it is not), so a persistent 4xx is not a realistic production path.

Treating every 4xx as confirmed-absent would also freeze transient client errors such as 401, 408, and 429 as "no enterprise" until cache expiry, which is worse than retrying on the next command.

src/core/telemetry/identity-fetch.ts
```diff
@@ -67,6 +73,11 @@ function connectionUserUuidResolved(conn: AuthConnection | undefined): boolean {
   return conn?.userUuid !== undefined;
 }
 
+/** Login persists `enterpriseUuid` (string or null); undefined means we have not tried yet. */
+function connectionEnterpriseUuidResolved(conn: AuthConnection | undefined): boolean {
```
georgii-borovinskikh-sonarsource, line 77: you can set up return type to `conn is AuthConnection` (type predicate)

georgii-borovinskikh-sonarsource: Although maybe it wouldn't work here because of `needsIdentityEnrichment`

src/core/telemetry/identity-fetch.ts
```diff
@@ -104,6 +120,8 @@ export function mergeIdentity(
   return {
     user_uuid: partial.user_uuid ?? base.user_uuid,
     organization_uuid_v4: partial.organization_uuid_v4 ?? base.organization_uuid_v4,
+    enterprise_uuid:
+      partial.enterprise_uuid !== undefined ? partial.enterprise_uuid : base.enterprise_uuid,
```
georgii-borovinskikh-sonarsource, line 125: why doesn't `??` work here?

src/core/telemetry/identity-fetch.ts
```diff
@@ -116,16 +134,20 @@ function cacheKey(auth: ResolvedAuth): string {
...
 
-function cacheEntryToIdentity(entry: CacheEntry): TelemetryIdentity {
-  return {
+function cacheEntryToIdentity(entry: CacheEntry): Partial<TelemetryIdentity> {
+  const identity: Partial<TelemetryIdentity> = {
     user_uuid: entry.userUuid ?? null,
     organization_uuid_v4: entry.organizationUuidV4 ?? null,
     sqs_installation_id: entry.sqsInstallationId ?? null,
   };
+  if ('enterpriseUuid' in entry) {
```
georgii-borovinskikh-sonarsource, line 144: Type predicate guard would work here though

src/core/telemetry/identity-fetch.ts
```diff
@@ -116,16 +134,20 @@ function cacheKey(auth: ResolvedAuth): string {
...
+    identity.enterprise_uuid = entry.enterpriseUuid ?? null;
+  }
+  return identity;
 }
 
 function fetchPlanIsComplete(plan: IdentityFetchPlan): boolean {
-  return !plan.user && !plan.org && !plan.sqs;
+  return !plan.user && !plan.org && !plan.enterprise && !plan.sqs;
```
georgii-borovinskikh-sonarsource: For readability `!(x || y || z)` would be better IMO. Although I don't really understand why there is negation in the first place. The method is called IsComplete, but it returns true if ALL properties are not set (if at least one is set, it will return false)

src/core/telemetry/identity-fetch.ts
```diff
@@ -116,16 +134,20 @@ function cacheKey(auth: ResolvedAuth): string {
...
     organization_uuid_v4: entry.organizationUuidV4 ?? null,
     sqs_installation_id: entry.sqsInstallationId ?? null,
   };
+  if ('enterpriseUuid' in entry) {
+    identity.enterprise_uuid = entry.enterpriseUuid ?? null;
+  }
+  return identity;
 }
 
 function fetchPlanIsComplete(plan: IdentityFetchPlan): boolean {
```
georgii-borovinskikh-sonarsource, line 151: I know it's not in the style of the current repo to do it, but encoding incomplete types as separate from complete (immutable) types would improve readability

src/core/telemetry/identity-fetch.ts
```diff
@@ -140,6 +162,11 @@ function planFieldsToFetch(
       !!auth.orgKey &&
       !identity.organization_uuid_v4 &&
       !(entry && 'organizationUuidV4' in entry),
+    enterprise:
+      auth.connectionType === 'cloud' &&
+      !!auth.orgKey &&
```
georgii-borovinskikh-sonarsource, line 169: Double negation?

src/core/telemetry/identity-fetch.ts
```diff
@@ -238,18 +296,33 @@ async function fetchMissingFromApi(
...
+      if (!org.resolved) {
+        enterprise_uuid = undefined;
+        resolved.enterprise = false;
+      } else if (!org.id) {
+        enterprise_uuid = null;
+        resolved.enterprise = true;
+      } else {
+        const enterprise = await fetchEnterpriseUuid(client, auth.serverUrl, org.id);
+        enterprise_uuid = enterprise.resolved ? enterprise.value : undefined;
+        resolved.enterprise = enterprise.resolved;
```
georgii-borovinskikh-sonarsource: this line looks really funny

georgii-borovinskikh-sonarsource: (I suspect the layout of the types/interfaces could be improved)

georgii-borovinskikh-sonarsource (approved):

src/core/telemetry/identity-fetch.ts
damien-urruty-sonarsource: Agreed — a `conn is AuthConnection` predicate would be wrong here. The helper can return `false` for a defined connection whose `enterpriseUuid` is still `undefined` (not resolved yet). Leaving it as a boolean.

src/core/telemetry/identity-fetch.ts
damien-urruty-sonarsource: `??` coalesces `null` as well as `undefined`. `null` here is confirmed-absent and must be kept; only `undefined` (not resolved yet) should fall back to `base`. Added a comment next to the assignment.

src/core/telemetry/identity-fetch.ts
damien-urruty-sonarsource: The `'enterpriseUuid' in entry` check is the runtime distinction we need (key missing vs explicit `null`). A type predicate would just wrap that without changing behavior, so I left it.

src/core/telemetry/identity-fetch.ts
damien-urruty-sonarsource: Rewrote as `!(plan.user || plan.org || plan.enterprise || plan.sqs)`. The name means the fetch plan has no remaining work — each flag is “should we fetch this field?”, so all-false is complete. Added a short JSDoc to make that explicit.

src/core/telemetry/identity-fetch.ts
damien-urruty-sonarsource: Same boolean coercion as the existing `org` plan field just above. Not changing only the new line.

src/core/telemetry/identity-fetch.ts
damien-urruty-sonarsource: Agreed it would be clearer, but that’s a wider typing refactor than this PR and not current repo style. Leaving the existing `TelemetryIdentity` / `Partial` split.

src/core/telemetry/identity-fetch.ts
damien-urruty-sonarsource: Pulled the unresolved / no-org-id / fetch-enterprise branches into `resolveEnterpriseUuid` so the call site is just assign value + resolved. Not introducing a complete/incomplete type hierarchy in this PR.

PR #706, opened 2026-08-24
CLI-969 Add installed-agent discovery module

vnaskos-sonar:
## Summary
- Adds `detectInstalledAgents()` (`src/core/host/environment/installed-agent-detector.ts`): detects which of the 5 supported AI agents (Claude Code, Copilot, Codex, Cursor, Antigravity) are installed on the current machine, via home-directory markers. Distinct from `agent-detector.ts`'s `detectCallerAgent()`, which identifies only the single agent currently invoking the CLI.
- Ported from the `vn/cli-931-integrate` PoC with one correction: the Antigravity marker `~/.gemini/antigravity` had no basis anywhere in the codebase; replaced with the confirmed real path, `ANTIGRAVITY_GLOBAL_CONFIG_DIR` (`~/.gemini/config`).
- This is the detection primitive only. Wiring it into `sonar integrate`'s bulk-install flow is out of scope here (tracked in [CLI-990](https://sonarsource.atlassian.net/browse/CLI-990)).

Ref: [CLI-969](https://sonarsource.atlassian.net/browse/CLI-969)

## Test plan
- [x] `bun test tests/unit/core/host/environment/installed-agent-detector.test.ts` — 5/5 pass
- [x] `bun run typecheck` — no new errors (pre-existing `launch-darkly` errors unrelated)
- [x] `bun x prettier --write` / `bun x eslint --fix` — clean, no changes needed
- [x] Vortex analysis (DEEP) on all changed files — no issues

[CLI-990]: https://sonarsource.atlassian.net/browse/CLI-990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLI-969]: https://sonarsource.atlassian.net/browse/CLI-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core/host/environment/installed-agent-detector.ts
```diff
@@ -0,0 +1,67 @@
...
+ * once. Deliberately excludes shared cross-tool directories such as
+ * `~/.agents`, which several agents write to and so identifies none of them.
+ */
+const AGENT_HOME_MARKERS: Record<DetectedAgentId, string[]> = {
+  cursor: ['.cursor'],
+  claude: ['.claude', '.claude.json'],
+  codex: ['.codex'],
+  copilot: ['.copilot'],
+  antigravity: [join('.gemini', 'config')],
+};
```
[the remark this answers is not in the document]
vnaskos-sonar: Fixed in f5eb482d — cursor's marker now uses `CURSOR_CONFIG_DIR` from `config-constants.ts`. Left the other markers (`.claude`, `.codex`, `.copilot`) as-is: their equivalent constants live in each agent's own `src/commands/integrate/<agent>/declaration.ts`, and importing those into `src/core/host/environment/` would invert the dependency direction (core depending on the commands layer). Cursor's is the only one that's already in the core-layer `config-constants.ts`, so it's the only one safe to reuse here.

src/core/host/environment/installed-agent-detector.ts
```diff
@@ -0,0 +1,67 @@
...
+// this answers "what should a machine-wide `sonar integrate` set up?", which
+// is a superset.
+
+import { existsSync } from 'node:fs';
+import { homedir } from 'node:os';
+import { join } from 'node:path';
+
+export const DETECTED_AGENT_IDS = ['cursor', 'claude', 'codex', 'copilot', 'antigravity'] as const;
+
+export type DetectedAgentId = (typeof DETECTED_AGENT_IDS)[number];
```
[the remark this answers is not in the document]
vnaskos-sonar: Fixed in f5eb482d — DetectedAgentId is now a direct alias of CallerAgent, and AGENT_LABELS is typed as Record<CallerAgent, string>, so the compiler now rejects a missing or extra agent between the two files. DETECTED_AGENT_IDS derives from Object.keys(AGENT_LABELS) instead of being a second hand-written literal array.

src/core/host/environment/installed-agent-detector.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const AGENT_LABELS: Record<CallerAgent, string> = {
+  cursor: 'Cursor',
+  claude: 'Claude Code',
+  codex: 'Codex',
+  copilot: 'Copilot',
+  antigravity: 'Antigravity',
+};
+
+export const DETECTED_AGENT_IDS = Object.keys(AGENT_LABELS) as CallerAgent[];
```
georgii-borovinskikh-sonarsource: at this point, it's not detecting anything and should be called `SUPPORTED_AGENT_IDS`

src/core/host/environment/installed-agent-detector.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export function detectInstalledAgents(home: string = homedir()): DetectedAgentId[] {
+  return DETECTED_AGENT_IDS.filter((agentId) =>
+    AGENT_HOME_MARKERS[agentId].some((marker) => existsSync(join(home, marker))),
+  );
+}
+
+export function isDetectedAgentId(value: string): value is DetectedAgentId {
+  return (DETECTED_AGENT_IDS as readonly string[]).includes(value);
```
georgii-borovinskikh-sonarsource: this looks incorrect - as it is now, `DETECTED_AGENT_IDS` is a static array, meaning this will return true

src/core/host/environment/installed-agent-detector.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+const AGENT_HOME_MARKERS: Record<DetectedAgentId, string[]> = {
+  cursor: [CURSOR_CONFIG_DIR],
+  claude: ['.claude', '.claude.json'],
+  codex: ['.codex'],
+  copilot: ['.copilot'],
+  antigravity: [join('.gemini', 'config')],
+};
+
+export function detectInstalledAgents(home: string = homedir()): DetectedAgentId[] {
```
georgii-borovinskikh-sonarsource, line 64: Will detecting by configs-only work with machine-wide installations on windows?

georgii-borovinskikh-sonarsource (changes requested):

src/core/host/environment/installed-agent-detector.ts
vnaskos-sonar: Good catch — renamed to SUPPORTED_AGENT_IDS in 2e1a283b (also fixed isDetectedAgentId -> isSupportedAgentId for the same reason, see the other thread).

src/core/host/environment/installed-agent-detector.ts
vnaskos-sonar: You're right — the name was misleading even though the logic itself was intentional (it's a type guard against the known/supported set, not a detection-result check). Renamed to isSupportedAgentId in 2e1a283b, alongside DETECTED_AGENT_IDS -> SUPPORTED_AGENT_IDS. isSupportedAgentId('claude') now reads correctly as "is this one of the ids we know how to handle", not "was claude detected on this machine" — that's what detectInstalledAgents()'s return value is for.

src/core/host/environment/installed-agent-detector.ts
vnaskos-sonar: Checked this rather than assuming: these exact paths (`.claude`, `.cursor`, `.codex`, `.copilot`, `.gemini/config`) are the same ones this codebase's own already-shipped, Windows-CI-tested `--global` integrations already read/write via homedir() on Windows (e.g. Claude's global settings path, CURSOR_CONFIG_DIR) — those flows already pass on the Windows integration job, so these locations are correct on Windows too, not just Unix.

The real caveat isn't Windows-specific though: config-directory detection only catches "has run at least once" (already called out in the module's own doc comment) — an agent that's installed (binary present, e.g. via a machine-wide/all-users install) but never launched by this particular user won't have created its config dir yet, on any platform. So on a machine where an admin installed an agent for all users but this user never ran it, detection would (correctly, per its current scope) say "not detected for this user."

The ticket's own scope allowed for "config directory presence and/or binary on PATH" — I went with config-only for this first cut since it's simpler and already tested. If the installed-but-never-run case matters for your rollout, happy to add a PATH-based fallback check as a second signal. Let me know.

src/core/host/environment/installed-agent-detector.ts
vnaskos-sonar: Correction to my last reply — I fixed the wrong thing. Renaming to isSupportedAgentId addressed the symptom but missed your actual point: we still need a function that answers "was this specific agent detected as installed", not just "is this a name we recognize". Fixed properly in 3fd8f1ef: isDetectedAgentId(value, home) now checks membership in detectInstalledAgents(home)'s real result, so it correctly returns false for a supported-but-not-installed agent. isSupportedAgentId stays as a separate, genuinely useful thing (a plain type-guard against the known set) — both now exist with names that match what they actually do.

georgii-borovinskikh-sonarsource (approved):

PR #707, opened 2026-08-24
CLI-971 Chain global git hooks to pre-existing local hooks

vnaskos-sonar:
## Summary
- Decision 9's answer: `sonar integrate git --global` sets `core.hooksPath` globally, which is a single-value override — it silently stops git from reading `.git/hooks/<name>` in every repo on the machine with no local override of its own (Husky is safe; plain hand-written hooks and `pre-commit`-framework hooks would otherwise go dark).
- `getHookScript()` now prepends a chaining block **for global-scope hooks only**: `git rev-parse --git-dir` resolves the physical `.git` directory regardless of any `core.hooksPath` override (`git rev-parse --git-path hooks` follows the override instead, so it's not usable here), letting the script find and run whatever hook already existed in that repo.
  - Old hook runs first; non-zero exit aborts immediately with the same code (preserves prior abort semantics, and lets a formatter/linter in the old hook run before Sonar's scan sees final content).
  - `pre-push` delivers its ref list on stdin, unlike `pre-commit`; a naive chain would drain that stdin before Sonar's own scan reads it, silently skipping the secrets check. Fixed by capturing stdin once into a temp file and replaying it to both readers — only `pre-push` pays this cost, `pre-commit`'s script is unchanged.
  - Reuses the existing overwrite/removal markers (`getRecognizedNativeMarkers`) as the double-chain guard, so an old per-repo Sonar install isn't executed twice.
- Repo-scope hooks are untouched — `wholeFile`'s existing overwrite refusal already protects that case.

Verified manually end-to-end, not just by test: installed the released CLI globally against a repo with a plain hand-written pre-commit hook — the old hook silently never ran, no warning. Rebuilt from this branch, reinstalled — the old hook now runs every time, before Sonar's own check.

Ref: [CLI-971](https://sonarsource.atlassian.net/browse/CLI-971)

## Test plan
- [x] `bun test tests/unit/commands/integrate/git/git-native-resource.test.ts` — 12/12 pass
- [x] `bun test tests/integration/specs/integrate/git.test.ts` — 48/48 pass (happy path / abort-on-failure / double-chain avoidance / pre-push stdin fix)
- [x] `bun run typecheck` / `bun run lint` — clean
- [x] Vortex analysis (DEEP) on all changed files — no issues
- [x] Manual end-to-end verification: released binary vs. this branch, same repo, same pre-existing hook — confirmed the behavior difference by hand (see PR comments for the walkthrough)

[CLI-971]: https://sonarsource.atlassian.net/browse/CLI-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,12 +51,38 @@ function nativeBinBlock(): string {
...
+  return [
+    'SONAR_GIT_DIR=$(git rev-parse --git-dir 2>/dev/null || :)',
+    'if [ -n "$SONAR_GIT_DIR" ]; then',
+    `  SONAR_LOCAL_HOOK="$SONAR_GIT_DIR/hooks/${hook}"`,
+    `  if [ -x "$SONAR_LOCAL_HOOK" ] && ! grep -qF ${markerArgs} "$SONAR_LOCAL_HOOK" 2>/dev/null; then`,
+    '    "$SONAR_LOCAL_HOOK" "$@" || exit $?',
+    '  fi',
+    'fi',
+  ].join('\n');
+}
```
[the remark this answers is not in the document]
vnaskos-sonar: Confirmed and fixed in f72c699d — good catch, this was a real bug, not a false positive. Verified by hand first: a naive chain left the second stdin reader (Sonar's own git-pre-push handler) with 0 lines, which getFileGroupsToScan silently treats as "nothing to scan" rather than an error. Went with the capture-and-replay approach rather than restricting chaining to pre-commit only, since pre-push is exactly the case a repo is most likely to have a real pre-existing hook for (and the original point of this ticket was to stop those from going dark under global scope) — punting on pre-push would leave the actual problem half-solved. Only pre-push pays the capture cost; pre-commit's generated script is unchanged.

tests/integration/specs/integrate/git.test.ts
```diff
@@ -892,6 +892,112 @@ describe('integrate git (native hooks)', () => {
...
+      mkdirSync(join(harness.cwd.path, '.git', 'hooks'), { recursive: true });
+      writeFileSync(hookPath, script, { mode: 0o755 });
+      chmodSync(hookPath, 0o755);
+    }
+
+    it(
+      'runs the pre-existing hook before Sonar’s own check, then still blocks a secret',
+      async () => {
+        await setupAuthenticated(harness, { withSecretsBinary: true });
+        initGitRepo(harness);
```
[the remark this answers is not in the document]
vnaskos-sonar: Added in f72c699d — a pre-push integration test where the chained hook actually reads stdin (not just argv), then pushes a secret to prove Sonar's scan still runs correctly afterward. This is exactly the test that would have caught the stdin-drain bug from the other comment.

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,73 @@ function nativeBinBlock(): string {
...
+ */
+function stdinCaptureBlock(): string {
+  return [
+    // A failed mktemp leaves SONAR_STDIN_CACHE empty; `cat >` into that empty path would
+    // otherwise fail with a raw, confusing shell error instead of a clear one.
+    'SONAR_STDIN_CACHE=$(mktemp) || { echo "sonarqube-cli: mktemp failed, skipping secrets scan"; exit 0; }',
+    'trap \'rm -f "$SONAR_STDIN_CACHE"\' EXIT',
+    'cat > "$SONAR_STDIN_CACHE"',
+  ].join('\n');
+}
```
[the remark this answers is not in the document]
vnaskos-sonar: Confirmed and fixed in 6133a387 — you were right, this was a real regression, not a false positive. Verified by hand first: `stdinCaptureBlock()` runs before `nativeChainBlock()`, so exiting on a failed `mktemp` skipped the chained pre-existing hook too, silently, while looking like success. Applied your suggested fix as-is: fall back to `${TMPDIR:-/tmp}/sonar-stdin-$$` when `mktemp` fails, and only bail (reusing `SONAR_HOOK_SKIP_SECRETS_MESSAGE` for consistency with the other skip path) if the capture write itself fails.

tests/unit/commands/integrate/git/git-native-resource.test.ts
```diff
@@ -140,3 +140,58 @@ describe('native git hook resource (wholeFile)', () => {
...
+    expect(script).not.toContain('git rev-parse --git-common-dir');
+    expect(script).not.toContain('SONAR_LOCAL_HOOK');
+  });
+
+  it('uses --git-common-dir, not --git-dir, so chaining also finds hooks from a linked worktree', () => {
+    const script = getHookScript('pre-commit', context({ scope: 'global' }));
+
+    expect(script).toContain('git rev-parse --git-common-dir');
+    expect(script).not.toContain('git rev-parse --git-dir');
+  });
```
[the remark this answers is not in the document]
vnaskos-sonar: Fair — confirmed by hand that the removed assertion could never fail (`--git-common-dir` never contains `--git-dir` as a substring), so it was testing nothing beyond the duplicate `toContain` check already in the earlier test. Replaced with your suggested exact-assignment assertion in 6133a387, which does regress if the flag ever reverts.

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,75 @@ function nativeBinBlock(): string {
...
+    '  fi',
+    'fi',
+  ].join('\n');
+}
+
 /** Returns the hook script. For pre-commit, bakes `--dependency-risks -p <key>` when `context` carries dep-risks attrs. */
 export function getHookScript(hook: GitHookType, context: IntegrationContext): string {
   const depRisksArgs = hook === 'pre-commit' ? resolveDepRisksArgs(context) : '';
+  const chains = context.scope === 'global';
+  // Only pre-push reads stdin, so only pre-push needs the capture-and-replay dance.
```
georgii-borovinskikh-sonarsource, line 112: Is this optimization important enough to keep these if-s in the code? Or would it be fine to always capture stdin?

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,75 @@ function nativeBinBlock(): string {
...
+  const needsStdinCapture = chains && hook === 'pre-push';
   return [
     '#!/bin/sh',
     `# ${getNativeHookMarker(hook)}`,
+    ...(needsStdinCapture ? [stdinCaptureBlock()] : []),
+    ...(chains ? [nativeChainBlock(hook, needsStdinCapture)] : []),
     nativeBinBlock(),
-    `"$SONAR_BIN" hook ${resolveSonarHookCommand(hook)}${depRisksArgs}`,
+    needsStdinCapture
+      ? `"$SONAR_BIN" hook ${resolveSonarHookCommand(hook)}${depRisksArgs} < "$SONAR_STDIN_CACHE"`
```
georgii-borovinskikh-sonarsource, line 121: some potential to reduce the copypasta

georgii-borovinskikh-sonarsource: stdinRedirect is already there in the other method

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,75 @@ function nativeBinBlock(): string {
...
+ * When `stdinFromCache` is set (pre-push only — see `stdinCaptureBlock`), the chained hook reads
+ * from the captured copy instead of the hook's own stdin, which was already drained by the capture.
+ */
+function nativeChainBlock(hook: GitHookType, stdinFromCache: boolean): string {
+  const markerArgs = getRecognizedNativeMarkers(hook)
+    .map((marker) => `-e ${shellQuoteBash(marker)}`)
+    .join(' ');
+  const stdinRedirect = stdinFromCache ? ' < "$SONAR_STDIN_CACHE"' : '';
+  return [
+    'SONAR_GIT_DIR=$(git rev-parse --git-common-dir 2>/dev/null || :)',
```
georgii-borovinskikh-sonarsource, line 98: Is this the DIR for other hooks? If so, SONAR_ is a bit misleading

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,75 @@ function nativeBinBlock(): string {
...
+ * a strict superset, not a behavior change there. The old hook runs first; a non-zero exit aborts
+ * immediately with the same code, preserving the abort semantics the repo had before Sonar's hook
+ * existed. The marker grep (reusing the same markers `wholeFile` already checks for
+ * overwrite/removal) skips chaining when the "pre-existing" hook is actually a Sonar hook from an
+ * earlier install, avoiding a double scan.
+ *
+ * When `stdinFromCache` is set (pre-push only — see `stdinCaptureBlock`), the chained hook reads
+ * from the captured copy instead of the hook's own stdin, which was already drained by the capture.
+ */
+function nativeChainBlock(hook: GitHookType, stdinFromCache: boolean): string {
```
georgii-borovinskikh-sonarsource, line 92: I remember you were talking about the need to modify some git config property for this to work. Is that not needed?

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,75 @@ function nativeBinBlock(): string {
...
+    // here would skip the chained pre-existing hook too, not just Sonar's own scan.
+    'SONAR_STDIN_CACHE=$(mktemp 2>/dev/null || :)',
+    '[ -z "$SONAR_STDIN_CACHE" ] && SONAR_STDIN_CACHE="${TMPDIR:-/tmp}/sonar-stdin-$$"',
+    'trap \'rm -f "$SONAR_STDIN_CACHE"\' EXIT',
+    `cat > "$SONAR_STDIN_CACHE" || { echo "${SONAR_HOOK_SKIP_SECRETS_MESSAGE}"; exit 0; }`,
+  ].join('\n');
+}
+
+/**
+ * Chains to a pre-existing local hook that a global `core.hooksPath` override would otherwise
```
georgii-borovinskikh-sonarsource, line 76: Pls no poems in comments

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,75 @@ function nativeBinBlock(): string {
...
+ * scan", not an error: `src/commands/hook/git-pre-push.ts`). Captured once into a temp file here,
+ * replayed to each reader below.
+ */
+function stdinCaptureBlock(): string {
+  return [
+    // mktemp can fail on an unwritable/full TMPDIR. Falling back to a pid-scoped path (rather
+    // than exiting here) matters because this block runs before nativeChainBlock() — bailing out
+    // here would skip the chained pre-existing hook too, not just Sonar's own scan.
+    'SONAR_STDIN_CACHE=$(mktemp 2>/dev/null || :)',
+    '[ -z "$SONAR_STDIN_CACHE" ] && SONAR_STDIN_CACHE="${TMPDIR:-/tmp}/sonar-stdin-$$"',
```
georgii-borovinskikh-sonarsource, line 69: do we really need to write it to a file? can't we store it in a variable?

src/commands/integrate/git/tools/native/shell-fragments.ts
```diff
@@ -50,14 +51,75 @@ function nativeBinBlock(): string {
...
+    'trap \'rm -f "$SONAR_STDIN_CACHE"\' EXIT',
+    `cat > "$SONAR_STDIN_CACHE" || { echo "${SONAR_HOOK_SKIP_SECRETS_MESSAGE}"; exit 0; }`,
```
georgii-borovinskikh-sonarsource, line 71: _SECRETS_MESSAGE ?

georgii-borovinskikh-sonarsource (approved):

src/commands/integrate/git/tools/native/shell-fragments.ts
vnaskos-sonar: Done — switched to `SONAR_STDIN_CACHE=$(cat)` (a shell variable) instead of a temp file, replayed via `printf '%s\n' "$SONAR_STDIN_CACHE" | ...`. This also removes the whole mktemp-failure fallback path from an earlier round, since there's no longer any disk I/O that can fail.

src/commands/integrate/git/tools/native/shell-fragments.ts
vnaskos-sonar: This whole branch is gone now that stdin is captured into a variable instead of a temp file — there's no more write-failure path to report, so the message-reuse question is moot.

src/commands/integrate/git/tools/native/shell-fragments.ts
vnaskos-sonar: A comment's a whisper, not a song,
best kept short, never too long —
so here's my one and final rhyme:
the docstring's trimmed, back in line.

Trimmed both doc comments to the one non-obvious "why" each — dropped the verification narrative and restated behavior.

src/commands/integrate/git/tools/native/shell-fragments.ts
vnaskos-sonar: I think this is about `core.hooksPath` — that's already set by the existing `--global` install flow (predates this PR), not something the chaining mechanism itself needs. Chaining doesn't touch git config at all: it just resolves the physical `hooks/` dir via `git rev-parse --git-common-dir`, which ignores whatever `core.hooksPath` points to, and runs whatever script is sitting there. So no extra config property is needed for this mechanism specifically — let me know if you meant something else.

src/commands/integrate/git/tools/native/shell-fragments.ts
vnaskos-sonar: Fair point — renamed to `SONAR_GIT_COMMON_DIR` so it's clear this is just the repo's shared `.git` dir (from `--git-common-dir`), not something Sonar-specific. Kept the `SONAR_` prefix for the same namespacing reason as the sibling vars (`SONAR_BIN`, `SONAR_LOCAL_HOOK`, `SONAR_STDIN_CACHE`).

src/commands/integrate/git/tools/native/shell-fragments.ts
vnaskos-sonar: Fixed — `sonarCommand` is now built once in `getHookScript` and reused for both branches, with only the `printf | ` prefix varying.

src/commands/integrate/git/tools/native/shell-fragments.ts
vnaskos-sonar: Kept the pre-push-only condition — it's not just an optimization but a correctness guard: pre-push's stdin is always terminated by git itself (the ref list, then EOF), so `cat` is safe there. pre-commit's stdin is whatever `git commit` inherited (often a terminal), and unconditionally reading it would hang an interactive commit waiting for input that never arrives. Added an explicit comment on `needsStdinCapture` to spell this out so it's not read as a pure performance call next time.

PR #709, opened 2026-08-24
CLI-1003 Clean up hook stdin listeners on timeout

damien-urruty-sonarsource:
## Summary
- `readRawStdin()` now removes `data`/`end`/`error` listeners when the read finishes
- Pause and destroy stdin so an idle pipe cannot keep the process alive after the 5s timeout
- Shared by all hooks; found during CLI-958 / #695 review but pre-existing


vnaskos-sonar (approved):
👌

PR #714, opened 2026-08-25
CLI-961 `get quality-gate` command

Krosovok:
----
## Summary by Gitar

- **New features:**
    - Added `get quality-gate` command to retrieve quality gate details using `QualityGatesClient`
- **Refactoring:**
    - Updated command registration and output rendering for quality gates


src/commands/get/quality-gate.ts
```diff
@@ -0,0 +1,78 @@
...
+  const [projectStatus, scope] = await Promise.all([
+    qualityGatesClient.getProjectStatus({ projectKey, ...queryParams }),
+    resolveDisplayScope(client, projectKey, options),
+  ]);
+
+  const metricsClient = new MetricsClient(client);
+  const metrics = projectStatus?.conditions.length ? await metricsClient.searchMetrics() : [];
+
+  const verdict = toVerdict(projectStatus?.status);
+  const conditions = selectConditions(projectStatus?.conditions, metrics, options.all);
```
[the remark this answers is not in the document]
Krosovok: This one is intentional: if we can not find it on the server, it probably is NOT_COMPUTED. Most likely that would mean analysis in progress for a fresh branch. Quality gate is not failing but is not green either.

src/commands/command-tree.ts
```diff
@@ -266,6 +269,27 @@ function buildCommandTree(runtime: CliRuntime): SonarCommand {
...
+    .option('-p, --project <project>', 'Project key')
+    .addOption(
+      new SonarOption('--format <format>', 'Output format')
+        .choices(QUALITY_GATE_VALID_FORMATS)
+        .default('json'),
+    )
+    .option('--branch <branch>', 'Branch name. Cannot be combined with --pull-request.')
+    .option('--pull-request <pull-request>', 'Pull request ID. Cannot be combined with --branch.')
+    .option('--all', 'Also show passing conditions. By default only failing conditions are shown.')
+    .authenticatedAction((ctx, options: GetQualityGateOptions) => getQualityGate(options, ctx));
```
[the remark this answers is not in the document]
Krosovok: I think it's fine. Not to show it in non-zero exit code is a design decision so showing it best effort (when we can) is good enough. Quality gates nature is to see them failing until they turn green. So a user supposedly will see the update notification eventually, just a few commands later.

src/commands/get/quality-gate/index.ts
```diff
@@ -0,0 +1,94 @@
+/*
```
damien-urruty-sonarsource, line 1: Nitpick: I find it a bit strange to have a quality-gate.ts file and a quality-gate-helpers folder. One suggestion would be to have a quality-gate folder inside get/, and rename quality-gate.ts to index.ts in this folder

src/core/server/quality-gates.ts
```diff
@@ -0,0 +1,53 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// SonarQube Quality Gates API wrapper
+
+import { type SonarQubeClient } from './client.ts';
+import type { ProjectStatus, ProjectStatusParams, ProjectStatusResponse } from './types.ts';
+
+export class QualityGatesClient {
```
damien-urruty-sonarsource, line 26: I don't know if this should be in `src/core`. This is supposed to be the "infra" or building blocks for other commands. Here I think it's about a specific API of a specific domain, so I wonder if we should leave it in the command directly, not sure

src/core/update/notification.ts
```diff
@@ -34,6 +34,15 @@ import { BACKGROUND_UPDATE_CHECK_TIMEOUT_MS, fetchLatestVersion } from './check.
 /** When to show the post-command update notice for an opted-in command. */
 export type UpdateNotificationCondition = (opts: Record<string, unknown>) => boolean;
 
+/**
+ * Condition for commands with a `--format` option: only notify for `table`,
+ * since JSON (and other machine-readable formats) stdout must stay pure.
+ */
+export const isTableFormatOption: UpdateNotificationCondition = (opts) => {
```
damien-urruty-sonarsource: This seems oddly placed, it's not related to the update? Or is it?

src/core/server/branches.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { ProjectBranch, ProjectBranchesResponse } from './types.ts';
+
+export class BranchesClient {
+  private readonly client: SonarQubeClient;
+
+  constructor(client: SonarQubeClient) {
+    this.client = client;
+  }
+
+  async listBranches(projectKey: string): Promise<ProjectBranch[]> {
```
damien-urruty-sonarsource, line 33: Same question about the placement

src/core/server/metrics.ts
```diff
@@ -0,0 +1,55 @@
...
+
+// SonarQube Metrics API wrapper
+
+import { type SonarQubeClient } from './client.ts';
+import type { Metric, MetricsSearchResponse } from './types.ts';
+
+/** `api/metrics/search` has no key/name filter, so a full-catalog fetch is the only option. */
+const METRICS_PAGE_SIZE = 500;
+
+export class MetricsClient {
```
damien-urruty-sonarsource, line 29: And same here too

src/commands/get/quality-gate/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      ? formatQualityGateTable({ verdict, project: projectKey, scope, conditions })
+      : formatQualityGateJson({ verdict, project: projectKey, scope, conditions });
+  print(message);
+
+  process.exitCode = exitCodeFor(verdict);
+}
+
+async function assertProjectExists(client: SonarQubeClient, projectKey: string): Promise<void> {
+  if (!(await client.checkComponent(projectKey))) {
+    throw new CommandFailedError(`Project '${projectKey}' does not exist or not accessible.`, {
```
damien-urruty-sonarsource, line 90: or `is` not accessible?

damien-urruty-sonarsource (commented):
Some suggestions for improvement. There are also Gitar findings to address

src/core/server/quality-gates.ts
Krosovok: I think it's the best place for now. Because it's better to have the Clients in the same place. Splitting them across several modules would be weird for me. Maybe they do have to be separate from core but then I'd rather put them into a separate module. For now it's basically an API access layer mixed up with core, but it's still better then having it tangled up with the commands.

src/core/update/notification.ts
Krosovok: Yes, it's a bit weird. Just a bit of refactoring from `command-tree.ts` to avoid duplication for the new command. It felt weird to have it in already over saturated command file, so I ended up moving it to what it's used for: `UpdateNotificationCondition`.

Might have been a separate file. Or `sonar-command.ts` where `showUpdateNotification` lives.

src/commands/get/quality-gate/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  process.exitCode = exitCodeFor(verdict);
+}
+
+async function assertProjectExists(client: SonarQubeClient, projectKey: string): Promise<void> {
+  if (!(await client.componentExists(projectKey))) {
+    throw new CommandFailedError(`Project '${projectKey}' does not exist or not accessible.`, {
+      remediationHint: 'Check the project key and your access to the project on the server.',
+    });
+  }
+}
```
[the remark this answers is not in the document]
Krosovok: It actually gets `Access denied (HTTP 404). Check that the supplied token and organization are valid.`
403 is used for unauthenticated users here. 404 shows for both non-existing and inaccessible.
I think it's straightforward enough.

src/core/server/quality-gates.ts
damien-urruty-sonarsource: We don't have to address this now and can think a bit more about it, there are anyway other HTTP clients in the core module, so I think it's fine for now.
I think it shouldn't be in src/core, but it's also not directly related to a specific command I agree, and some web APIs could be invoked by different commands. Maybe another module would make sense yes

src/core/update/notification.ts
damien-urruty-sonarsource: it's used for the update, but semantically it is about checking an option. Some ideas:
* move it to a separate file
* find a way to attach some metadata to options or option values, so the update check would simply call `option.canSafelyPrintOnStdOut`, and this could be done at option declaration time somehow
* this is about not showing a message in the case of JSON or other structured format because we don't want to break parsing. One thing I did for alpha/beta stages is to print messages/warnings on stderr. This way we make sure we don't pollute the output and we don't need to check this kind of things

I think we could use the last idea in a separate ticket, so we can also keep it like you did for now

Krosovok: I think I will move it to a separate file for now then.

damien-urruty-sonarsource (approved):

PR #719, opened 2026-08-26
CLI-985 Pass invocation context to command handlers

damien-urruty-sonarsource:
## Summary
- Thread `CommandInvocationContext` through secrets scans (hooks, `analyze`, SCA manifest pre-scan) so later work can defer telemetry from those call sites


tomshafir-sonarsource (approved):
LGTM!

PR #720, opened 2026-08-26
CLI-985 Make core telemetry event-agnostic

damien-urruty-sonarsource:
## Summary
- Make core telemetry event-agnostic: `emitTelemetryEvent(name, fields)` prefixes `Analytics.Cli.`, merges identity, and appends to ndjson
- Move domain payload types and analyzer/integrate helpers next to their producers; they still emit immediately
- Keep `storeEvent` in `postAction` with the existing `(command, success, agentSessionId)` signature, and emit integrate telemetry from an `onSuccess` callback so the framework never imports event types


src/commands/integrate/git/index.ts
```diff
@@ -227,9 +228,17 @@ async function installGitFeatures(
     auth,
     force: options.force,
     nonInteractive: options.nonInteractive,
-    isFromRouter: options.isFromRouter,
     // Attrs are project-scope only; global hooks do not support a project key.
     attrs: scope === 'project' ? { projectKey: options.project ?? null } : undefined,
+    onSuccess: (facts) =>
```
damien-urruty-sonarsource: These `onSuccess` callbacks are temporary. They are replaced by something else in the next PR

src/core/framework/features/install-integration.ts
```diff
@@ -47,6 +46,14 @@ import type {
 } from './types.ts';
 import { isFeatureContainer } from './types.ts';
 
+/** Facts from a successful install; commands turn these into telemetry. */
+export interface InstallIntegrationSuccessFacts {
```
georgii-borovinskikh-sonarsource, line 50: subfeatures?

georgii-borovinskikh-sonarsource (approved):

src/core/framework/features/install-integration.ts
damien-urruty-sonarsource: Hmm no it's not specifically about subfeatures, but when an integration was successfully installed

georgii-borovinskikh-sonarsource: @damien-urruty-sonarsource sorry I should've been more clear. I meant wouldn't we want to record subfeatures too?

damien-urruty-sonarsource: That would probably be interesting data, yes, I would need to check with Farah. This PR is just a refactoring, so I would address it as a separate ticket

PR #725, opened 2026-08-26
CLI-984 Record installed binaries only in `dependencies.installed`

tomshafir-sonarsource:
----
## Summary by Gitar

- **State Management:**
    - Retired legacy `tools.installed` registry and migrated entries to `dependencies.installed`
    - Added `saveStateKeepingInstalledDependencies` to prevent overwriting dependency records during state updates


docs/state-management.md
```diff
@@ -118,6 +119,8 @@ The `agentExtensions` registry stores per-project or global artifacts installed
 
 ### Tools Section
 
+Legacy. Entries are folded into `dependencies.installed` on load and this list is deleted.
```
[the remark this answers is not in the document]
tomshafir-sonarsource: Redundant information

docs/state-management.md
```diff
@@ -30,7 +30,8 @@ The state file persists configuration across CLI invocations and stores:
 | `auth`            | AuthState             | Authentication and server connections               |
 | `agents`          | AgentsState           | Configuration for each agent (Claude Code, etc.)    |
 | `config`          | CliConfig             | CLI configuration metadata                          |
-| `tools`           | ToolsState (optional) | Installed tools and binaries                        |
+| `tools`           | ToolsState (optional) | Legacy, superseded by `dependencies`                |
+| `dependencies`    | DependenciesState     | Installed binaries shared by declarative features   |
```
damien-urruty-sonarsource, line 34: There are some examples below showing usage of tools, and no usage of dependencies. We could also update them

src/core/host/install/binary.ts
```diff
@@ -110,6 +110,8 @@ async function downloadAndInstall(
   const binaryPath = join(resolvedBinDir, binaryName);
 
   if (!options.force && existsSync(binaryPath)) {
+    // Heal state when the binary is on disk but not (or no longer) recorded.
+    recordInstalledDependency(spec.name, spec.version, binaryPath);
```
damien-urruty-sonarsource: Is this necessary? I think we read and write the state every time, which is maybe a bit too much.

src/core/state/state-repository.ts
```diff
@@ -160,6 +196,22 @@ export function saveState(state: CliState): void {
   }
 }
 
+/**
+ * Save state without overwriting `dependencies.installed`: binary installers record there
+ * themselves, writing straight to disk after the caller loaded its snapshot. Callers that
+ * install binaries between their `loadState()` and their save must use this.
+ */
+export function saveStateKeepingInstalledDependencies(state: CliState): void {
```
damien-urruty-sonarsource, line 204: I find it a bit strange that we have to save a partial state, and also pay the cost of loading the state before saving. Also, if the loading throws, we still end up replacing the dependencies in the file. WOuld there be a better way to achieve this?

tomshafir-sonarsource: This is a pre-existing patch extracted to a helper in this PR. After discussion, we decided to fix it in a future PR.

docs/state-management.md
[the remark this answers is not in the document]
tomshafir-sonarsource: Redundant information.

src/core/host/install/binary.ts
```diff
@@ -146,7 +146,7 @@ async function downloadAndInstall(
   );
   print(`     ${spec.name} ${installedVersion}`, channel);
 
-  recordInstallationInState(spec.name, installedVersion, binaryPath);
+  recordInstalledDependency(spec.name, spec.version, binaryPath);
```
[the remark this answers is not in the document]
tomshafir-sonarsource: This gap is acceptable for now. It was never explicitly tested and was essentially a side effect, not an intended feature. Plus the binaries would self-heal on their next version update.

damien-urruty-sonarsource (approved):

PR #730, opened 2026-08-28
CLI-129 Check the organization exists during login

charafsalmi:
On SonarQube Cloud, `--org` was already validated. Three paths were not:

- an organization key read from `sonar-project.properties` or `.sonarlint`
- a key typed at the manual entry prompt (the `hgfjhgfhj` case in the report)
- a failed lookup, which was reported to the user as "organization not found"

This extends the existing check to the first two, and makes the check itself honest.

`/api/organizations/search` answers an unknown key with `200` and an empty list, so an error never means the organization is missing. `resolveOrganizationAccess()` now returns `accessible | not_found | check_failed`, and login gives the last two their own message.

`checkOrganization()` keeps its boolean contract by delegating to the same resolver, so
the `integrate` preflight is unchanged.

`listUserOrganizations()` no longer swallows errors either: an empty membership list sends
the flow to the manual prompt, so an outage must not look like one.

## Two notes for review

- Validation stays **after** token generation. Resolving a private organization needs a  token, so checking earlier would break the members-only case. Instead, a rejected organization now revokes the token this login just minted. A reused keychain token, or one pasted by hand, is left alone.
- The `organizations=` filter is not membership-scoped, so a public organization the user is not a member of still validates. That is the point of manual entry. The fake server now models this (`withVisibleOrganizations`).

## On the diff size

Most of the change in `login.ts` is moved code: `resolveOrganization()` scopes the token revoke, `promptForOrganizationKey()` replaces two duplicated manual-entry blocks.
Behaviour for the membership menu and for on-premise is **unchanged**.

Tests cover each validation path, plus one unit test for releasing stdin on failure — the integration harness runs non-TTY, so that branch is unreachable there.

The second commit is unrelated .gitignore housekeeping : `bun build --compile` drops a ~60 MB copy of the Bun runtime in the working directory on every build.


src/commands/auth/login.ts
```diff
@@ -53,57 +62,96 @@ export async function authLogin(options: AuthLoginOptions): Promise<void> {
...
+async function resolveOrganization(
+  server: string,
+  isCloud: boolean,
+  orgOption: string | undefined,
+  auth: BrowserAuthResult & { reusedExistingToken: boolean },
+): Promise<string | undefined> {
+  try {
+    return isCloud
+      ? await validateOrSelectOrganization(new SonarQubeClient(server, auth.token), orgOption)
+      : await setupOnPremiseOrganization(orgOption);
```
damien-urruty-sonarsource: It's not related to your changes but this branch does not make sense. There is no such thing as "on-premise organization", SonarQube Server does not have the concept of organizations. I think we should return `undefined` in this case

src/commands/auth/login.ts
```diff
@@ -145,9 +193,106 @@ async function getOrGenerateToken(
...
+ * missing organization.
+ */
+function organizationAccessError(
+  org: string,
+  access: Extract<OrganizationAccess, { status: 'not_found' | 'check_failed' }>,
+): CommandFailedError {
+  if (access.status === 'not_found') {
+    return new CommandFailedError(`Organization '${org}' not found or not accessible.`, {
+      remediationHint:
+        "Check the organization key (keys are lowercase) and your access, then rerun 'sonar auth login'.",
```
damien-urruty-sonarsource: > keys are lowercase

It's maybe true today but I don't think there is any explicit contract about the organization key being lowercase. I would suggest not to mention that here

src/core/server/client.ts
```diff
@@ -859,22 +860,33 @@ export class SonarQubeClient {
...
+      return { status: 'check_failed', reason: (error as Error).message };
     }
   }
 
+  /**
+   * Check if organization exists and is accessible.
+   *
+   * Use `resolveOrganizationAccess` to tell a missing organization from a failed lookup.
+   */
+  async checkOrganization(organizationKey: string): Promise<boolean> {
```
damien-urruty-sonarsource: Suggestion: rename to `isOrganizationAccessible`, this tells more than `checkOrganization`

damien-urruty-sonarsource (approved):
LGTM, some suggestions

PR #740, opened 2026-09-01
CLI-1037 Add runInteractive to the test harness

damien-urruty-sonarsource:
## Summary
- Adds `harness.runInteractive()` and an `InteractiveSession` (`waitText`, keystrokes, `waitFinish`, `kill`)
- Leaves `run()` and `stdin` / `stdinChunks` in place so existing tests keep working

### DIFF
--- a/tests/integration/harness/index.ts
+++ b/tests/integration/harness/index.ts
@@ -37,13 +37,15 @@ import { FakeGitLabServer, FakeGitLabServerBuilder } from './fake-gitlab-server.
 import { FakeSonarQubeServer, FakeSonarQubeServerBuilder } from './fake-sonarqube-server.js';
 import { FakeUpdateScriptServer } from './fake-update-script-server.js';
 import { File } from './file';
+import { type InteractiveSession, startInteractiveSession } from './interactive-session.js';
 import { buildHomeEnv, IS_WINDOWS } from './platform';
-import type { CliResult, RunOptions } from './types.js';
+import type { CliResult, RunInteractiveOptions, RunOptions } from './types.js';
 
 export { FakeGitLabServer, FakeGitLabServerBuilder } from './fake-gitlab-server.js';
 export { FakeSonarQubeServer, FakeSonarQubeServerBuilder } from './fake-sonarqube-server.js';
+export { InteractiveSession } from './interactive-session.js';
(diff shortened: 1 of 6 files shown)


tests/integration/harness/index.ts
```diff
@@ -231,6 +234,23 @@ export class TestHarness {
...
+      cwd: options?.cwd ?? this.cwd.path,
+      timeoutMs: options?.timeoutMs,
+      waitTimeoutMs: options?.waitTimeoutMs,
+      binaryPath: options?.binaryPath,
+      browserToken: options?.browserToken,
+      browserTokenName: options?.browserTokenName,
+    });
+    this.sessions.push(session);
+    return session;
+  }
```
tomshafir-sonarsource, line 252: Are there any actual tests that need more than a single interactive session?

tests/integration/harness/interactive-session.ts
```diff
@@ -0,0 +1,386 @@
...
+        this.rawStderr += chunk;
+        this.notify();
+      }),
+    ]).then(() => undefined);
+
+    void proc.exited.then((code) => {
+      this.exitCode = code;
+      clearTimeout(this.timer);
+      this.notify();
+    });
```
tomshafir-sonarsource, line 206: > Refactor this asynchronous operation outside of the constructor.sonarqube(typescript:S7059)

I get a very "loud" SonarQube warning on these lines. Can we fix it cleanly? Maybe do those async operations after the constructor?

tomshafir-sonarsource (approved):
Looks good! I think this would be a very good change for the future. Left two questions.

tests/integration/harness/index.ts
damien-urruty-sonarsource: No, it's not the case today. But this new method returning a session, nothing prevents tests from doing so in the future, so I need to support that. I think it's just a little bit of overhead, and it should be fine

tests/integration/harness/interactive-session.ts
damien-urruty-sonarsource: Where did you get the issue? I didn't get it with Vortex analysis or on this PR

damien-urruty-sonarsource: Fixed

PR #743, opened 2026-09-01
CLI-1041 Persist known-server-project mappings and rework mapping conflict resolution

georgii-borovinskikh-sonarsource:
----
## Summary by Gitar

- **Mapping persistence:**
    - Persisted `state.knownServerProjectMappings` table to cache project mappings across upgrades
    - Added post-update migration and live fallback in `discoverProject()` for mapping resolution
    - Updated mapping conflict and completion rules in `src/core/known-server-project-mappings.ts`
- **System reset:**
    - Clears persisted known-server-project mappings on `sonar system reset`


src/core/project-info.ts
```diff
@@ -363,7 +370,7 @@ function collectKnownRoots(known: KnownMappingsResult | undefined): string[] {
   );
 }
 
-/** Matches against mappings derived live from installed features, caller-loaded once above. */
+/** Matches against the combined persisted + live-derived mapping table, caller-loaded once above. */
 function applyKnownServerProjectMapping(
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: Confirmed the connection backfill (`resolveMappingConnection`) only reaches into the caller's current auth/active connection, not into a sibling candidate's own recorded `serverUrl` — so the gap is real. That said, triggering it needs all of these at once: (1) a genuine `targetRoot` conflict actually exists (transient, e.g. mid-migration after re-integrating with a new project key), (2) the winning candidate specifically lacks a recorded `serverUrl`, and (3) both `resolveAuth()` and `getActiveConnection(state)` come back empty — i.e. this machine has no working *and* no previously-recorded connection at all. I checked every `discoverProject()` call site; the only one where `auth` can be `null` at match time is the pre-commit `--dependency-risks` fallback in `git-pre-commit.ts`, and even there `getActiveConnection` independently covers the common case (a connection recorded in state but with an since-invalidated token). Given how narrow the full stack is, leaving this as a known limitation for now rather than adding candidate-iteration logic for it.

src/core/host/recorded-feature-resolver.ts
```diff
@@ -22,7 +22,6 @@
 // is the sole caller, so SQAA and CAG project lookups share one selection policy.
 
 import { pathComparisonKey } from '../io/fs-utils.ts';
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: This is deleted, no?

src/core/known-server-project-mappings.ts
```diff
@@ -18,48 +18,60 @@
...
+/** True when `incoming` has server info that `current` lacks — the only case a duplicate gets replaced. */
+function isMoreComplete(
+  incoming: KnownServerProjectMapping,
+  current: KnownServerProjectMapping,
+): boolean {
+  return incoming.serverUrl !== undefined && current.serverUrl === undefined;
+}
+
+/** Adds `mapping` into `mappings` in place; only collapses with an entry sharing both `targetRoot` and `projectKey` (a true duplicate) — a different `projectKey` at the same `targetRoot` is a conflict, kept as a separate candidate for `selectFeatureForLookupPaths` to resolve. */
+function addMapping(
```
tomshafir-sonarsource: maybe mergeMapping or upsertMapping would be better?

src/core/known-server-project-mappings.ts
```diff
@@ -18,48 +18,60 @@
...
+  mappings: KnownServerProjectMapping[],
+  mapping: KnownServerProjectMapping,
+): void {
+  const duplicateIndex = mappings.findIndex(
+    (entry) =>
+      pathComparisonKey(entry.targetRoot) === pathComparisonKey(mapping.targetRoot) &&
+      entry.projectKey === mapping.projectKey,
+  );
```
tomshafir-sonarsource, line 53: Look correct to me!

src/core/known-server-project-mappings.ts
```diff
@@ -18,48 +18,60 @@
...
-  orgKey?: string;
-  /** ISO timestamp of the contributing feature's last update; used to resolve conflicts when merging. */
-  updatedAt: string;
+/** True when `incoming` has server info that `current` lacks — the only case a duplicate gets replaced. */
+function isMoreComplete(
+  incoming: KnownServerProjectMapping,
+  current: KnownServerProjectMapping,
+): boolean {
+  return incoming.serverUrl !== undefined && current.serverUrl === undefined;
+}
```
tomshafir-sonarsource, line 42: Don't know how necessary this really is but I don't think it hurts either.

src/commands/system/reset.ts
```diff
@@ -112,6 +112,8 @@ export async function systemReset(options: SystemResetOptions): Promise<void> {
 
       applyCleanedState(state, mergeCleanedFields(results.map((r) => r.cleaned)));
       clearLegacyState(state);
+      // Not tied to any step above — would otherwise survive reset indefinitely.
+      state.knownServerProjectMappings = [];
```
tomshafir-sonarsource, line 116: Maybe put inside `clearLegacyState`?

src/core/known-server-project-mappings.ts
tomshafir-sonarsource: Technically we'll persist some mappings we'll basically never read, as on a targetRoot conflict we just take the first match. But I think it's fine. We don't lose data this way and we can tighten it later.

tomshafir-sonarsource (approved):
Looks good! Left a few small comments.

src/core/known-server-project-mappings.ts
georgii-borovinskikh-sonarsource: Good call — renamed to `upsertMapping`.

src/commands/system/reset.ts
georgii-borovinskikh-sonarsource: I'd rather keep this separate — clearLegacyState resets agents/agentExtensions, fields that only exist for pre-declarative-framework compatibility. knownServerProjectMappings isn't legacy, it's the live table discoverProject() actively reads today, so folding the wipe in there would mislabel it as something on its way out. Leaving it as its own explicit line with a comment felt more honest about why it's here.

src/core/known-server-project-mappings.ts
georgii-borovinskikh-sonarsource: Agreed, that's a fine tradeoff to accept for now — we're not losing data and can revisit if it turns out to matter.

src/core/update/known-project-mappings-migration.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// One-time historical backfill of `state.knownServerProjectMappings`, consumed by the
+// post-update mechanism that runs automatically after CLI upgrades. The actual attrs-parsing
+// derivation (buildKnownServerProjectMappings) lives in core/known-server-project-mappings.ts,
+// shared with discoverProject()'s live fallback. Must run before `migrateDeclarativeIntegrations`
+// (see post-update.ts): once integrations move to global scope they stop carrying per-folder
+// `projectKey`/`repoRoot` attrs, so this is the last point where that data is still readable
+// from the pre-reconciliation feature set.
```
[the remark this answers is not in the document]
georgii-borovinskikh-sonarsource: it's fine, if it throws it means it won't succeed ever, so it's fine to skip

PR #744, opened 2026-09-01
CLI-795 Enforce fetching with the network options

charafsalmi:
## TL;DR

**The problem.** The CLI already knew how to go through a corporate proxy and trust a company SSL certificate. But applying that configuration was left to each place in the code that makes a network call. Forget it once, and the CLI silently connects directly instead: it fails, or worse, it works on the developer's laptop and fails only at the customer's. Three calls had already forgotten it.

**What this does.** One single file is now allowed to make network calls, and it applies the configuration itself. Nobody can forget it any more, because the parameter to forget no longer exists. A lint rule blocks any new direct network call and tells the author which helper to use, so the CI catches it instead of a reviewer.

**The limits.** This covers the requests the CLI makes itself. Crash reports go out through the Sentry SDK, which uses its own transport: it does not see our proxy or certificate settings, and no lint rule can reach inside a dependency. So behind a mandatory corporate proxy, crash reports still do not leave the machine. That gap is now written down instead of implied, and needs its own ticket.

---

Ticket: [CLI-795](https://sonarsource.atlassian.net/browse/CLI-795) (epic [CLI-907](https://sonarsource.atlassian.net/browse/CLI-907))

## What changed

`src/core/server/fetch-guarded.ts` is now the only module allowed to call the runtime `fetch`, and exports two wrappers that both resolve the network options themselves:

- `fetchGuarded(url, init)` - the default. Applies the configuration **per hop** (which makes the old `schemeChanged` recomputation block redundant) and keeps blocking cross-origin redirects so a bearer token is never forwarded.
- `fetchWithNetworkConfig(url, init)` - new, for credential-free requests that must keep following CDN redirects: binary download, `stable.version` check, and `fetchServerVersion`. These were the three remaining direct `fetch` calls.

Call sites can no longer skip **or** override the configuration: `buildFetchInit()` cannot carry proxy/TLS options anymore (parameter dropped), and both wrappers strip any `proxy`/`tls` keys found on `init` before spreading the resolved ones. As a side effect, 8 repetitions of `buildFetchNetworkOptions(url)` disappear from the two clients.

An ESLint `no-restricted-syntax` rule rejects `fetch(...)` and the qualified forms (`globalThis.fetch`, `Bun.fetch`) across `src/**`, disabled only in the wrapper module.

The rule cannot tell the two wrappers apart, so `fetchWithNetworkConfig` **throws** on a credential header (`authorization`, `cookie`, `private-token`, `x-api-key`): it lets the runtime follow redirects, which would hand that header to another origin. That keeps the invariant executable rather than a comment.

The module stays where it is: moving and splitting it is [CLI-842](https://sonarsource.atlassian.net/browse/CLI-842).

## Behavior change worth a look

`flushTelemetryEvents()` no longer pre-resolves the options once per batch, so with a custom CA the root certificate list is copied per event instead of per batch (negligible for batches of a few events). The important part is preserved: on `NetworkConfigError` the whole batch is abandoned and requeued rather than sent without the configuration. `sendTelemetryEvent()` was extracted for that, to keep the nesting depth acceptable.

Also renamed the `fetch` parameter of `with503Retry` (`sqaa-api.ts`) to `attemptRequest`, otherwise the new rule flags it.

## Verification

- `typecheck` clean; `lint` reports 0 errors and the same 16 pre-existing warnings as `master`.
- `test:unit` 2578 pass, `test:integration` 941 pass. The integration suite covers the converted paths against a fake CDN (binary download and `stable.version`).
- Mutation-checked the five new guarantees (not applying the options, not stripping a caller-supplied proxy, not aborting the telemetry batch, not rejecting a credential header, and a raw `fetch` added under `src/`): each one is caught by a test or by the ESLint rule.
- `fetchGuarded`'s redirect handling is unchanged line for line apart from the two edits above, and stays covered by the existing suite: same-origin follow, 301 chain, cross-origin reject, HTTP→HTTPS upgrade with proxy switch, POST→GET downgrade with `Content-Type` stripping, redirect limit, and 304 passthrough.
- No manual run: the changes are only observable under a proxy or custom CA, which the integration tests cover.

[CLI-795]: https://sonarsource.atlassian.net/browse/CLI-795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CLAUDE.md
```diff
@@ -169,6 +169,15 @@ The two inputs are separate because the single boolean made _collect but do not
...
+- `fetchWithNetworkConfig(url, init)` — applies the network configuration and lets the runtime follow redirects normally. Only for credential-free requests that need to follow a CDN redirect: `downloadBinary` (`core/host/install/sonarsource-releases.ts`), the `stable.version` check (`core/update/check.ts`), and `fetchServerVersion` (`core/server/server-info.ts`).
+
+Call sites never pass proxy/TLS options: `buildFetchInit(method, headers, timeoutMs, body)` deliberately cannot carry them, and `fetchWithNetworkConfig` drops any `proxy`/`tls` keys found on `init` before spreading the resolved ones, so the configuration cannot be overridden locally. `fetchGuarded` resolves options **per hop**, so a followed HTTP→HTTPS upgrade never reuses the options computed for the original scheme. An unusable configuration surfaces as `NetworkConfigError` rather than a silent direct connection — `flushTelemetryEvents` catches it and requeues the whole batch instead of retrying per event.
```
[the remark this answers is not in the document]
charafsalmi: Fixed in 3f534875, before this comment landed - an independent review pass had reached the same conclusion, which is a good sign the point is real.

I did not take the suggested wording verbatim, because the finding understates itself. Scoping the claim to first-party requests is necessary but not sufficient: a reader still needs to know that crash reports specifically do not reach us behind a mandatory proxy. So `CLAUDE.md` now scopes the guarantee *and* names the gap, with the reason the ESLint rule cannot close it (the transport lives in `node_modules`).

The gap itself is a real defect, not a documentation issue: behind a corporate proxy with a custom CA, no Sentry event leaves the machine, and nobody currently knows. It is out of scope here and gets its own ticket, following the precedent of CLI-902 (telemetry had the same defect).

charafsalmi:
<details><summary>[Retracted] Addressed by @damien-urruty-sonarsource  in /pull/742</summary>@damien-urruty-sonarsource heads-up on the one Gitar finding still open here, so it does not slow this review down.

**It is not about this pull request.** Gitar flagged `CLAUDE.md:228`, the `stdinChunks` sentence. That line is not in this diff. `gh pr diff 744` has zero occurrences of `stdinChunks`. Gitar read the whole file instead of the diff, and my rebase shifted that line down by about ten lines.

**But the finding is correct.** The line groups `stdinChunks` with dump-all stdin. It is the opposite of dump-all. It writes one chunk per prompt, paced by `stdinChunkDelayMs`, because a single dump makes readline discard the bytes meant for later prompts. Its own doc in `types.ts` says exactly that, and the specs use it that way.

The practical risk: a contributor reads that line, concludes `harness.run()` cannot drive prompts, and reaches for `runInteractive()` for cases the suite already covers with chunked stdin.

**Fixed separately in #751.** One line, docs only. It comes from #740, so it belongs there rather than in a pull request about network configuration. Nothing to do here.

One more thing, unrelated to the review. Gitar's CI section on this pull request says "CI failed: All test jobs passed successfully; no CI failures were detected". That is self-contradictory and wrong. The `Build` run is red, on 5 `real Claude Code` e2e tests that time out at 180s. They reproduce on a re-run of the same commit with no code change, and the log shows `Claude did not emit JSON (exit 143)`, so the agent never answers and our harness kills it. That points away from this diff and towards the credentials or the API, which matches what you suspected. Separately, the Windows leg failed once on `analyze dependency-risks > auto-installs sca-scanner-cli` with a hook timeout, and that test never touches the Anthropic API, so it looks like a second, unrelated problem.</details>

src/core/telemetry/telemetry-events.ts
```diff
@@ -167,6 +166,30 @@ function parseValidEvents(content: string, now: number): StoredTelemetryEvent[]
...
 }
 
+/**
+ * POSTs one event. `requeue` leaves it for the next flush attempt; `abort` means the
+ * proxy/TLS configuration is unusable, so the rest of the batch is not worth attempting.
+ */
+async function sendTelemetryEvent(
+  event: StoredTelemetryEvent,
+  timeoutMs: number,
+): Promise<'sent' | 'requeue' | 'abort'> {
```
damien-urruty-sonarsource: Nitpick: The return values are not consistent: one is past tense (sent), other ones are verbs. I would suggest to use past tense everywhere or nouns (something like success, network_error, unknown_error)

src/core/telemetry/telemetry-events.ts
```diff
@@ -167,6 +166,30 @@ function parseValidEvents(content: string, now: number): StoredTelemetryEvent[]
...
+        'POST',
+        { 'Content-Type': 'application/json', 'x-api-key': TELEMETRY_API_KEY },
+        timeoutMs,
+        JSON.stringify(event, (_key, value) => (value === null ? undefined : value)),
+      ),
+    );
+    return 'sent';
+  } catch (err) {
+    return err instanceof NetworkConfigError ? 'abort' : 'requeue';
+  }
```
damien-urruty-sonarsource, line 201: We should do something with this error, or the root cause will be lost. Maybe logging it would be sufficient? see `logger.error()` in the codebase

src/core/server/fetch-guarded.ts
```diff
@@ -56,21 +100,29 @@ const REDIRECT_STATUSES = new Set([
...
+ * Network options are resolved per hop, so a followed HTTP→HTTPS upgrade never
+ * reuses the proxy/TLS options computed for the original scheme.
  */
 export async function fetchGuarded(url: string, init: RequestInit): Promise<Response> {
   let currentUrl = url;
   let currentInit = init;
 
   for (let redirectCount = 0; ; redirectCount++) {
-    const response = await fetch(currentUrl, { ...currentInit, redirect: 'manual' });
+    const response = await applyNetworkConfigAndFetch(currentUrl, {
```
damien-urruty-sonarsource: What is the difference between `applyNetworkConfigAndFetch` and `fetchWithNetworkConfig`? Those 2 functions seem to have similar contracts. Could we find a better name to differentiate the 2, or merge them if possible?

src/core/server/fetch-guarded.ts
```diff
@@ -18,24 +18,68 @@
...
 import type { FetchNetworkOptions } from '@/core/host/connectivity/types.ts';
 
 /**
- * `networkOptions` is required so a new call site cannot silently skip the resolved
- * proxy/TLS configuration. Pass `buildFetchNetworkOptions(url)` for the URL being fetched.
+ * Builds a `RequestInit` for the wrappers below. It deliberately carries no
+ * proxy/TLS options: those are owned by the wrappers, not by call sites.
  * `body` stays positional but explicit: pass `undefined` for bodyless requests.
  */
 export function buildFetchInit(
```
damien-urruty-sonarsource: Unrelated to your changes, but I find the name of this function a bit confusing. Shouldn't it be something like `buildRequest()`? It's not really about initializing the fetch, it's more about the request that is being sent

src/core/server/fetch-guarded.ts
```diff
@@ -18,24 +18,68 @@
...
+/**
+ * Fetch with the resolved proxy/TLS configuration applied for `url`, following
+ * redirects the way the runtime normally would.
+ *
+ * For requests that carry no credentials only. Since the runtime follows redirects
+ * here, a credentialed request could hand its token to another origin, so a
+ * credential header is rejected outright rather than left to review: use
+ * `fetchGuarded` for those.
+ */
+export async function fetchWithNetworkConfig(
```
damien-urruty-sonarsource: 2 comments: 
* I think this is some new behavior. I would have put it in a separate ticket and PR, or at least different commit. FYI, the release notes is automatically built by an agent based on tickets and commits, and this change would deserve to be mentioned in the release notes
* the namings are a bit confusing. The 2 entrypoints are `fetchGuarded` and `fetchWithNetworkConfig`. This kind of implies the first one does not come with network config. I'm also not sure what "guarded" means here

I think what you want to achieve here is make some requests anonymous and some authenticated. Couldn't we have `fetchAnonymous` and `fetch`/`fetchWithAuth`/`fetchAuthenticated`? Instead of checking the absence of a header, we could instead rely on the caller calling the right one based on their situation. We could even imagine the `fetch` taking the token and transforming it to a header.

damien-urruty-sonarsource (approved):
I think the main goal of the ticket is achieved but I have some concerns about naming and clarity for callers.

src/core/telemetry/telemetry-events.ts
charafsalmi: Good catch, the union mixed one state with two instructions. It is now a named SendOutcome with three states: success, send_error, config_error. The caller maps them to actions. I kept config_error rather than unknown_error/network_error split because the reason to stop the batch is specific: the proxy/TLS config is unusable, so no other event in the batch can succeed either.

src/core/telemetry/telemetry-events.ts
charafsalmi: Agreed, and it is worse than it looks: the flush worker is spawned detached with stdio: ['ignore','ignore','ignore'], so stderr goes nowhere. logger.error() writes to the log file only, never to the console, so it is the one channel where the cause survives. Thank you.

src/core/server/fetch-guarded.ts
charafsalmi: Done, buildRequest. I first meant to defer it to CLI-842 to keep this PR narrow, but that ticket rewrites client.ts, which holds 12 of the sites the other renames already touch. Deferring would have bought nothing.

src/core/server/fetch-guarded.ts
charafsalmi: Renamed, your suggestion is the right one: fetchAuthenticated and fetchAnonymous, in src/core/server/fetch.ts. Neither adds the credential, the caller still supplies the header, so the doc comment says so explicitly. Taking the token as a parameter is a bigger change and belongs in CLI-842.

On the separate commit: we squash-merge, so a second commit here does not reach master and cannot feed the release-notes agent. Only a separate ticket and PR would. In this case I do not think it earns one: the assertion is a developer guardrail, no current call site trips it, and no user-visible behavior changes. Happy to split it out if you would rather have the trace.

src/core/server/fetch-guarded.ts
charafsalmi: I renamed it to sendRequest. It now pairs with buildRequest: build the request, send it. The two public wrappers are the only names a call site has to reason about.

src/core/server/fetch-guarded.ts
damien-urruty-sonarsource: We don't necessarily squash-merge, you can also rebase-merge if you prefer. It depends on the case, if you want to have several commits in the PR and keep them on master, you can rebase

charafsalmi: You are right ! I was stuck with the assumption, but we can definitely rebase-merge. For this here, i'll squash, but in the next PRs I'll take care about the commits that really do split along behavior lines.
Thank you.

PR #745, opened 2026-09-01
CLI-1012 Add coverage category enrichment

Krosovok:
Show a per-file breakdown of the worst offenders under each failing coverage-related quality gate condition (top 3 files by default), in both the table and JSON output of `quality-gate status`.

### DIFF
--- a/src/commands/quality-gate/status/breakdown.ts
+++ b/src/commands/quality-gate/status/breakdown.ts
@@ -1,0 +1,137 @@
+/*
+ * SonarQube CLI
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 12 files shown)


src/commands/quality-gate/status/breakdown.ts
```diff
@@ -0,0 +1,137 @@
...
+  new_line_coverage: 'coverage',
+};
+
+export interface QualityGateBreakdownEntry {
+  path: string;
+  value: string;
+  formattedValue: string;
+}
+
+export interface QualityGateMetricBreakdown {
```
kirill-knize-sonarsource: The ticket AC and the MMF-5796 refinement doc both specify `{ condition: { metric, status }, totalCount, worstFiles: [{ file, value }] }` for the breakdown entry, but the current shape is `{ metric, totalCount, fetchedCount, entries: [{ path, value, formattedValue }] }`. Since this JSON is documented as a stable contract, worth checking this is an intentional revision (and update the ticket/doc) or should be aligned to spec before it merged.

Krosovok: Yes, I made some adjustments along the way. I will double-check with Clifford and will update PR or ticket.

Krosovok: I confirmed with Clifford the change and updated the ticket.
I also noticed that our JSON structure was a bit redundant: a separate array for metric details and a separate array for breakdowns for the same metrics. So I moved the breakdown under related metric instead of hanging separately.

kirill-knize-sonarsource (approved):
LGTM

PR #741, opened 2026-09-01
CLI-1037 Drive interactive tests with runInteractive

damien-urruty-sonarsource:
## Summary
- Moves interactive and dump-all stdin call sites to `runInteractive()`
- `run()` still accepts `stdin` / `stdinChunks` so this change is test-only

PR #751, opened 2026-09-01
CLI-1037 Describe stdinChunks as scripted prompt writes, not dump-all stdin

charafsalmi:
## TL;DR

One sentence in `CLAUDE.md` puts `stdinChunks` in the wrong category. It calls it dump-all stdin. It is the opposite of dump-all.

The risk is a contributor going the wrong way. They read the line. They conclude `harness.run()` cannot drive prompts. They reach for `runInteractive()` for cases the existing suite already handles with chunked stdin.

One line changed, docs only.

## What the line said

> Use `harness.run()` for non-interactive commands and dump-all stdin (`stdin` / `stdinChunks`).

## Why that is wrong

`stdinChunks` exists *because* dump-all breaks interactive prompts. Its own doc says so, in `tests/integration/harness/types.ts`:

> Use this when the CLI shows multiple sequential interactive prompts: sending all bytes at once causes readline to buffer and discard later bytes before the next prompt has started listening.

It writes one chunk per prompt, spaced by `stdinChunkDelayMs`. The specs use it exactly that way. `integrate-bare.test.ts` sends a Ctrl+C and arrow keys. `claude.test.ts` sends `['\r', '\r', '\r']` with a delay.

So `harness.run()` does drive prompts. It just drives them blind.

## The distinction the doc now makes

The useful split is not interactive versus non-interactive. It is whether the test needs to read the output before deciding what to send.

- `harness.run()` writes a scripted sequence, blind, with pacing.
- `harness.runInteractive()` reads the output first, then decides.

## Where this came from

Found while reviewing #744. Gitar flagged the line there, even though it is not in that pull request's diff. The line comes from #740 (CLI-1037), merged earlier today. It is split out here to keep #744 focused on its own subject.

Verified before writing: the `stdinChunks` doc in `types.ts`, and every use of it under `tests/integration/specs/`.


damien-urruty-sonarsource:
@charafsalmi I'm currently getting rid of this mechanism in another PR (#742). This sentence will disappear completely. You can close this PR

EDIT: the exact change is here: https://github.com/SonarSource/sonarqube-cli/pull/741/changes#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7R217

PR #758, opened 2026-09-02
CLI-1049 Cap harness temp-dir removal at 2s

damien-urruty-sonarsource:
## Summary
- Windows integration `afterEach` was timing out because `TestHarness.dispose()` waited on `fs.rm` of a just-used CAG/secrets PE (DeleteFile can block ~6–8s; Bun's hook budget is 5s).
- Stop waiting after 2s: recursive `rm` is best-effort, continues in the background, leftovers stay in OS temp.
- Investigation (traces, Handle, per-file unlink) and alternatives are on CLI-1049.


kirill-knize-sonarsource (approved):
LGTM

PR #766, opened 2026-09-03
CLI-1056 Stop baking the project key into Vortex hook scripts

tomshafir-sonarsource:
----
## Summary by Gitar

- **CLI hooks:**
    - Removed baked-in project keys from Vortex hook templates for Claude Code and Codex integration
    - Made `--project` option optional on `claude-post-tool-use` and `codex-post-tool-use` hook commands


src/commands/integrate/claude/hook-templates.ts
```diff
@@ -51,10 +46,10 @@ export function getPostToolUseFailureTemplateWindows(): string {
   return windowsTemplate('sonar hook claude-post-tool-use-failure');
 }
 
-export function getSqaaPostToolTemplateUnix(projectKey: string): string {
-  return unixTemplate(formatSqaaPostToolHookCommandUnix('claude-post-tool-use', projectKey));
+export function getSqaaPostToolTemplateUnix(): string {
```
georgii-borovinskikh-sonarsource, line 49: Not a template anymore

georgii-borovinskikh-sonarsource: Same for other methods in this file. None of them get the template, they return the fully formed value FROM template, which is just an implementation detail

src/commands/integrate/claude/hooks.ts
```diff
@@ -230,21 +228,18 @@ export interface InstallHooksOptions {
...
-   * SQAA remains installed project-locally because it is always project-scoped.
    */
   skipSecretsHooks?: boolean;
 }
 
 /**
- * Install all hooks (cross-platform).
- * Secrets hooks install to globalDir (if provided), SQAA hook installs to projectRoot.
- * SQAA hook is only installed when installSqaa is true (requires cloud connection + entitlement).
+ * Reinstall the secrets hooks (cross-platform) into globalDir when provided,
```
georgii-borovinskikh-sonarsource, line 236: I don't understand why we drop the full hook feature like that

src/commands/command-tree.ts
```diff
@@ -765,7 +765,7 @@ function buildCommandTree(runtime: CliRuntime): SonarCommand {
   hookCommand
     .command('claude-post-tool-use')
     .description('PostToolUse handler: run Vortex analysis after agent edits or writes a file')
-    .requiredOption('--project <key>', 'SonarQube project key')
+    .option('--project <key>', 'SonarQube project key')
```
georgii-borovinskikh-sonarsource: if we don't bake it in the hook anymore, what is the point of keeping this parameter?

src/commands/integrate/claude/hooks.ts
```diff
@@ -230,21 +228,18 @@ export interface InstallHooksOptions {
...
+ * otherwise projectRoot. The Vortex analysis hook is deliberately not installed
+ * here: it needs an entitlement check, which the post-update caller cannot do.
  */
 export async function installHooks(
```
georgii-borovinskikh-sonarsource, line 240: Is this function needed at all now?

georgii-borovinskikh-sonarsource (changes requested):

src/commands/integrate/claude/hooks.ts
tomshafir-sonarsource: This is a legacy migration that is no longer in use. It called the SQAA script generators and would have required rework to match the changes in this PR. However, since no one actually used the SQAA branch with this legacy method in practice, I removed it. The long-term solution is to evaluate whether we still need this legacy method at all and, if not, remove it entirely.

src/commands/command-tree.ts
tomshafir-sonarsource: There's no need for it. I left it in case the migration fails for whatever reason and the user still has the old scripts in place. It is now optional.

src/commands/integrate/claude/hooks.ts
tomshafir-sonarsource: This is the legacy migration method. See explanation above. We should re-evaluate whether this is still needed or can be removed entirely.

georgii-borovinskikh-sonarsource (approved):

src/commands/integrate/claude/hook-templates.ts
tomshafir-sonarsource: Will be handled in a follow-up PR.

src/commands/command-tree.ts
```diff
@@ -765,7 +765,9 @@ function buildCommandTree(runtime: CliRuntime): SonarCommand {
   hookCommand
     .command('claude-post-tool-use')
     .description('PostToolUse handler: run Vortex analysis after agent edits or writes a file')
-    .requiredOption('--project <key>', 'SonarQube project key')
+    // Tolerates the removed `--project` from stale scripts.
+    .allowUnknownOption()
+    .allowExcessArguments()
     .anonymousAction(handleHookInvocation(agentPostToolUse));
```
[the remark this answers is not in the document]
tomshafir-sonarsource: @gitar-bot Replaced with `.option('--project <project>', 'Deprecated; the project key is resolved by discovery') // Accepted and ignored`. Please re-assess this issue.

PR #773, opened 2026-09-03
CLI-842 Move the shared API domains onto their own clients

charafsalmi:
Layer 2 of 3 — replaces #752. Depends on #772.

Moves the six shared API domains off `SonarQubeClient` and onto their own clients, each constructed from a `SonarHttpClient`: `OrganizationsClient`, `ProjectBindingsClient`, `ComponentsClient`, `UsersClient`, `SystemClient`, `ScaClient`. `VortexEntitlementClient` moves to `src/core/vortex/entitlement.ts` at the same time, since it is a pure consumer of `OrganizationsClient`.

Every caller of those domains is migrated in the same PR, and the corresponding methods are deleted from `SonarQubeClient`, which drops from **1162 to 278 lines**.

### Why the migration is in the same PR as the clients

Adding the clients first and migrating callers later was the original plan, and it does not work: the extracted method bodies are **byte-identical** to the ones still in `client.ts`, doc comments included. Two copies of the same 200-line file is exactly what the duplication detector is for, so that split would have opened with a failing quality gate on our own product. A domain therefore only appears here by leaving `client.ts` in the same change — no duplication at any point in the stack.

### The one deliberate piece of scaffolding

Six methods (`getOrganizationLegacyId`, `fetchOrganizationByKey`, `getOrganizationAlmKey`, `hasPrivateProjectsEntitlement`, `getProjectSettings`, `getComponentId`) are still called by the command-level surfaces that move in #774. Rather than leave their bodies behind, `SonarQubeClient` composes the new clients and forwards:

```ts
private readonly organizations = new OrganizationsClient(this);
private readonly components = new ComponentsClient(this);
```

Forwarding methods are normally the thing to avoid — see the API-layer rules added in #774 — but they exist here for exactly one layer, they carry no logic, and they are what keeps the duplication at zero. They are deleted along with the class in #774.

### Review effort

The line count is 3650, but it is not 3650 lines of new logic:

| | |
|---|---|
| Genuinely new source | 6 client files, ~630 lines — mostly methods moved verbatim |
| Caller migration | ~300 lines across 46 files, typically 2–6 lines each (constructor + call site) |
| New tests | ~1100 lines, one test file per client |
| Pure deletion | 871 lines from `client.ts`, 798 from `client.test.ts` |

The two deletions are trivially checkable: the removed methods have no remaining caller, which is what makes the rest of the diff compile.

### Verification

Typecheck clean · lint 0 errors (16 pre-existing warnings) · **2614 unit tests, 0 fail** · **946 integration tests, 0 fail**.

### Stack

Stack #775: #772 → **#773 (this)** → #774. Merges bottom-up.

----
## Summary by Gitar

- **Client refactoring:**
    - Extracted domain-specific API logic from `SonarQubeClient` into dedicated clients like `OrganizationsClient`, `ComponentsClient`, and `ProjectBindingsClient`
    - Updated core entitlement checks and CLI commands to use the modular client architecture


src/core/vortex/entitlement.ts
```diff
@@ -18,16 +18,169 @@
...
+  'enabled' | 'over_consumption' | 'not_entitled' | 'check_failed' | 'not_applicable';
+
+export interface VortexEntitlementResult {
+  status: VortexEntitlementStatus;
+  consumption?: { consumed: number; limit: number };
+}
+
+/**
+ * Server has no organizations, but entitlement lives on `/…/{id}` — omitting the
+ * segment 404s. The nil UUID is a valid UUID the CLI can send without knowing
```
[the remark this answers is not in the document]
charafsalmi: It's normal, because it's an iterative move. We need that duplication for the next PRs and it will be removed eventually in the last one.

src/core/server/sca.ts
```diff
@@ -0,0 +1,68 @@
+/*
```
damien-urruty-sonarsource, line 1: I have a bit the same remark here about the placement. Eventually `src/core` should contain only foundational code, building blocks for features. Web APIs interaction is a bit in between, but some APIs clearly belong to other teams and specific features. I'm not sure what's the best approach. Maybe it can stay there

src/core/telemetry/identity-fetch.ts
```diff
@@ -287,14 +283,14 @@ async function resolveEnterpriseUuid(
   if (!org.id) {
     return { value: null, resolved: true };
   }
-  const enterprise = await fetchEnterpriseUuid(client, serverUrl, org.id);
+  const enterprise = await fetchEnterpriseUuid(client, org.id);
```
damien-urruty-sonarsource, line 286: I think this class could be split further, the fetch methods should be also part of dedicated clients. We could have one client per API path? I think that's the direction we are taking. Can be done as a follow-up.

src/commands/quality-gate/status/index.ts
```diff
@@ -53,7 +54,7 @@ export async function qualityGateStatus(
   const projectKey = await resolveProjectKey(options.project, auth, true);
   noteProject(auth, projectKey);
 
-  const client = new SonarQubeClient(auth.serverUrl, auth.token);
+  const client = new SonarHttpClient(auth.serverUrl, auth.token);
```
damien-urruty-sonarsource, line 81: An idea for a further improvement maybe. Each command is reponsible to build the client, based on auth (URL and token). We could imagine providing it via the context that is passed to every command handler. it would also maybe easier to provide a fake for unit tests

damien-urruty-sonarsource (approved):
LGTM, some suggestions for potential follow-ups (don't have to address them now)

src/core/server/sca.ts
charafsalmi: Leaving it, as you suggest. For the record: **ScaClient** is a thin wrapper over one **/feature-enabled endpoint**, and it's consumed both from **core** (**sca-availability.ts**) and from commands. 

**core/server** is the only place both can import from without creating a commands -> commands dependency. If we later split **core** into foundational vs. feature-owned APIs, this moves with the rest rather than on its own.

src/core/telemetry/identity-fetch.ts
charafsalmi: Agreed, and it's the natural next step. Worth noting it's mostly a deletion: identity-fetch.ts calls four endpoint families, and three already have clients after this PR (UsersClient, OrganizationsClient, SystemClient). Only /enterprises/enterprise-organizations would need a new one. 

I'll create a ticket for that.

src/commands/quality-gate/status/index.ts
charafsalmi: This is the one I'd act on first, and it's bigger than it looks in a good way: it also includes a Gitar finding on #774, where sqaa-api builds a fresh SonarHttpClient per analysis chunk. 

That's harmless today only because the transport holds no state. 

Putting the client on the invocation context answers the lifecycle question once for every command instead of per call site, and makes faking it in unit tests trivial as you say. 

I'll create a ticket for that.

src/core/server/projects.ts
```diff
@@ -17,13 +17,23 @@
...
 
+/**
+ * The `ps` (page size) ceiling on nearly every paginated SonarQube list endpoint - not specific
+ * to any one resource. Confirmed identical across `components/search_projects`,
+ * `measures/component_tree`, `issues/search`, and `metrics/search`'s own param docs ("must be
+ * greater than 0 and less or equal than 500"). The unrelated, narrower
+ * `DOP_REPOSITORIES_MAX_PAGE_SIZE` on `ImportApiClient` is a real exception specific to that one
+ * API.
+ */
+export const MAX_PAGE_SIZE = 500;
```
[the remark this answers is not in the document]
charafsalmi: Good catch, and it's my wording — I adapted this comment when relocating `MAX_PAGE_SIZE` here to resolve a semantic clash with CLI-962, which had moved the constant into `client.ts` (the opposite direction of this refactor). `ImportApiClient` indeed only arrives with #774.

I didn't take the suggested fix, though: pointing at `SonarQubeClient` is accurate at this commit but wrong from #774 on, which deletes that class — it would just move the stale reference one layer up. I dropped the class name instead, so the sentence holds at every layer of the stack and after it.

Fixed in b0371390.

stanislavhh (approved):

PR #774, opened 2026-09-03
CLI-842 Move the command API surfaces onto their own clients and retire SonarQubeClient

charafsalmi:
Layer 3 of 3 — replaces #752. Depends on #773.

Moves the four command-level API surfaces onto their own clients next to the command that drives them — `ImportApiClient`, `RemediateApiClient`, `SqaaAnalysisClient` (with the wire shapes in `sqaa-wire-types.ts`), and the `ScaScanApi` port — then **deletes `SonarQubeClient`**, its 1162-line source and its 986-line test, since nothing calls it any more.

Also closes the two transitional seams opened earlier in the stack:

- `raiseForStatus` goes back to `private` on `SonarHttpClient`.
- `tests/unit/core/server/fetch.test.ts` (142 lines) takes over the redirect and `fetchAuthenticated` coverage that lived in `client.test.ts`, so deleting that file costs no coverage.
- `CLAUDE.md` documents the resulting layer and the three rules that hold across all fifteen clients.

### Equivalence with the reviewed PR

This is the proof that matters for the whole stack:

```
$ git diff <this branch> <#752 rebased onto master>
(empty)
```

The end state is **byte-identical** to #752 as already reviewed — by me, and by Gitar across five findings. Splitting into three layers introduced no new change; it only decided where the seams fall.

### Dead code: 3 removed, 4 deliberately kept

Seven methods had no caller left after the refactor. Only three are deleted (`getCurrentUser`, `getSystemStatus`, `checkQualityProfiles`). The other four — `hasProvisionProjectsPermission`, `hasProjectBeenAnalyzed`, and two organization lookups — are still referenced by the open #723, so removing them would hand that PR a conflict for no gain. They stay until it lands.

### ⚠️ Overlap with #723

@stanislavhh's #723 adds methods to `src/core/server/client.ts`, which **this PR deletes**. The split actually helps here: #772 and #773 can land first and #723 can rebase onto them without losing work, and this PR is the only one that needs to wait for #723 to merge. Whichever of the two goes second, the work moves into `ProjectBindingsClient` / `UsersClient` / `ComponentsClient` rather than being redone.

### Review effort

| | |
|---|---|
| Genuinely new source | 4 client files + wire types, ~380 lines, methods moved verbatim |
| Caller migration | 37 files; the `sqaa-*` cluster is a one-line type-import change each |
| Pure deletion | 2148 lines (`client.ts` + `client.test.ts`) |
| Docs | `CLAUDE.md` |

### Verification

Typecheck clean · lint 0 errors (16 pre-existing warnings) · **2568 unit tests, 0 fail** · **946 integration tests, 0 fail**.

### Stack

Stack #775: #772 → #773 → **#774 (this)**. Merges bottom-up. Once this lands, #752 can be closed unmerged.


CLAUDE.md
```diff
@@ -179,6 +179,53 @@ Call sites never pass proxy/TLS options: `buildRequest(method, headers, timeoutM
...
+`ScaScanApi` port (`src/commands/analyze/dependency-risk-helpers/sca-api.ts`), and
+`VortexEntitlementClient` (`src/core/vortex/entitlement.ts`, owning `VortexEntitlementResult` /
+`VortexEntitlementStatus` and `SERVER_ORGANIZATION_ID_PLACEHOLDER`).
+
+Three rules hold across all fifteen of them, with no exception — keep it that way when adding one.
+
+**Every API client is constructed from a `SonarHttpClient`**, never from a `(serverUrl, token)` pair
+it turns into one itself. The command handler builds the transport client once and passes it in, so
+a single instance can be shared by every domain client in a run — which is what keeps
+`OrganizationsClient`'s organization cache effective instead of one cache per caller.
```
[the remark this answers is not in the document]
charafsalmi: This behaviour was already there in master. It's out of the scope of CLI-842. And as you said : « behaviourally harmless today ».
This is a refactoring ticket, not a change in the behaviour of the app.

src/commands/import/_common/import-api.ts
```diff
@@ -0,0 +1,113 @@
...
+  readonly organizations: OrganizationsClient;
+  private readonly client: SonarHttpClient;
+
+  constructor(client: SonarHttpClient) {
+    this.client = client;
+    this.organizations = new OrganizationsClient(client);
+  }
+
+  get isCloud(): boolean {
+    return this.client.isCloud;
```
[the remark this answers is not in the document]
charafsalmi: This is **fidelity** to master, not an oversight. sqaa-analysis-client.test.ts and remediate-api.test.ts are not new coverage: they carry over the createAnalysis and scheduleAgentJob describes that lived in client.test.ts (lines 1192 and 1372 on master), which this PR deletes.

fetchDopRepositoriesPage, provisionProject, requestAutoscanEligibility and checkAiRemediationEntitlement had no unit test on master, so there was nothing to carry over.

Coverage is neither added nor removed anywhere in this PR.

They are not untested either: tests/integration/specs/import/import.test.ts holds 66 tests driving sonar import against the fake server, which exercise all three methods end to end — including the page.total unwrapping and the swallow-on-failure path. Per CLAUDE.md, integration tests are the default here and unit tests are the exception, so adding one would go against the stated policy rather than follow it.

No change.

src/commands/import/_common/import-api.ts
```diff
@@ -0,0 +1,113 @@
+/*
```
damien-urruty-sonarsource, line 1: I just notice it now, and it was already there, so not related to your changes, but I'm not sure this `_common` folder serves a lot here. It's not shared by several (sub)commands, it's only used by the import command. Those files could be moved directly to the import/ folder

src/commands/remediate/remediate-api.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  async checkAiRemediationEntitlement(
+    orgKey: string,
+  ): Promise<{ status: AiRemediationEntitlement }> {
+    try {
+      // Not OrganizationsClient.getOrganizationLegacyId: it maps a failed lookup to null,
+      // which would report an unreachable server as 'not_eligible' instead of 'unknown'.
+      const orgsEndpoint = '/organizations/organizations';
+      const orgs = await this.client.get<Array<{ id: string; uuidV4: string; name?: string }>>(
```
damien-urruty-sonarsource, line 59: It seems we are calling the orgs APIs, shouldn't we use the org client?

src/commands/analyze/sqaa-wire-types.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// Request and response shapes of the Vortex analysis endpoint, exactly as they go over the
+// wire. A leaf module on purpose: it imports nothing, so every SQAA module can depend on it.
+
+export type SqaaAnalysisDepth = 'STANDARD' | 'DEEP';
```
damien-urruty-sonarsource, line 24: Not sure it was the case for other clients, do we expose wire types via a separate file? I don't know what makes the most sense, maybe the types could be part of the client files, because anyway we need to import them together

damien-urruty-sonarsource (approved):
LGTM, also suggestions for improvement

src/commands/import/_common/import-api.ts
charafsalmi: Agreed, and as you say it predates this PR. 

Leaving it out on purpose: the argument this stack makes to reviewers is that it only relocates the API layer, and folding in an unrelated folder move would blur that. 

Good candidate to fold into whatever next touches import.

I will also create a ticket for that.

src/commands/remediate/remediate-api.ts
charafsalmi: The reason is in the comment above that call: getOrganizationLegacyId maps any failed lookup to null, so an unreachable server would be reported as not_eligible instead of unknown. 

OrganizationsClient has no method today that returns the organization record and preserves the error/absence distinction

resolveOrganizationAccess preserves it but deliberately returns no record, and this code needs the legacy id. Adding one is the clean fix. I'd rather do that as a follow-up than widen this PR.

I will create a ticket for that.

src/commands/analyze/sqaa-wire-types.ts
charafsalmi: No other client does this, and the reason is the import graph rather than a principle: those types are imported by around 14 sqaa-* modules that never touch the client (display, chunking, validation, telemetry). 

Keeping them in sqaa-analysis-client.ts would force all of them to import the client just for types. 

Where types and client share the same consumers (Organization in organizations.ts, for instance) I did keep them together. So the working rule is "types live with their client unless they have a wider audience than it". 

I can write that into CLAUDE.md if you'd like it explicit.

PR #782, opened 2026-09-04
CLI-844 Wrap SonarHttpClient methods in a Result monad

charafsalmi:
## TL;DR

Converts `SonarHttpClient` and every domain client under `src/core/server/` from throw-on-failure to `neverthrow`'s `ResultAsync`, with real chaining through the whole client layer rather than a `Result` that gets unwrapped immediately at every call site. Adds `.orThrow()`, a method (not a free function) for the one place a chain still needs to collapse back to a throw, at a command's own boundary. Reviewed twice by Gitar and twice by dedicated verification passes; currently approved with 0 open findings. No change in CLI behavior visible to users.

## Summary

`SonarHttpClient` threw on failure. Every domain client above it either let that propagate or wrapped the call in an ad hoc `try/catch`. This replaces throwing with `ResultAsync` across the whole client layer, so failure composes with `.map()`/`.andThen()`/`.orElse()` instead of being caught at every call site.

This started as the narrower hand-rolled `Result<T>` described in the ticket's original scope. Mid-review the team decided to go further and adopt `neverthrow` for real. See the ticket comments for why.

### What changed

- `SonarHttpClient`'s HTTP methods (`get`, `getOrNotFound`, `getSafe`, `post`, `postForm`, `postFormJson`, `genericRequest`) all return `ResultAsync<T, HttpClientError>` now. `getSafe` used to be the one exception that still rejected on a transport failure; it doesn't anymore, so there is a single contract across the class instead of two.
- `HttpClientError` is a closed union (`src/core/server/errors.ts`), not a bare `Error`, so `.mapErr()` and matching on the error type are exhaustive at the call site.
- Every domain client under `src/core/server/` (branches, components, issues, measures, metrics, organizations, project-bindings, projects, pull-requests, quality-gates, sca, system, users, duplications) chains off the client's `ResultAsync` instead of unwrapping it immediately.
- `.orThrow()` is a new method on `ResultAsync`/`Ok`/`Err` (`src/core/result.ts`, added via TypeScript declaration merging plus a runtime prototype patch, since `neverthrow` doesn't ship one). It replaces an earlier free-function `unwrapOrThrow(x)` with `x.orThrow()`, so most call sites no longer need to import anything from `core/result.ts` at all. Constrained to `E extends Error`, the same guarantee the old free function had. A new `no-restricted-imports` rule blocks importing `neverthrow` anywhere except that one file, so nothing can build a `Result` without also loading the method.
- `resolveOrganizationAccess()`/`isOrganizationAccessible()` stay `Promise`-returning on purpose: `.match()` already folds every outcome into a value, so there is nothing left on the error rail to propagate past that point.

### Verification

Every conversion was checked against its pre-conversion throwing behavior: the success case, every distinct error status the old code special-cased, every swallow-to-fallback branch, every `isCriticalFailure` guard. A separate pass swept the whole diff for anything missed: dead `try/catch` blocks, leftover `unwrap`/`unwrapOrThrow` references, discarded `Result` chains, and call sites still assuming an old throwing contract that had changed.

### Regressions found and fixed during this work

- `toGetResult()`/`toStatusCheckedResult()` used `fromSafePromise()` around a promise (`buildStatusError()`) that can itself reject. Fixed by switching to `fromPromise(..., toError)`.
- `getSafe()` built the request URL synchronously before entering the wrapped async block, so a malformed server URL threw instead of producing an `Err`. Fixed by moving the URL construction inside.
- `getSafe()` also classified a malformed JSON body on an otherwise-2xx response as `TransportError` (critical), while `toStatusCheckedResult` (used by `post`/`postFormJson`) classified the identical situation as `UnexpectedApiError` (non-critical). Both now split the fetch and body-parse steps the same way, so a malformed body classifies consistently.
- Three call sites (`remediate/index.ts`, `import/index.ts`, `resolve-options.ts`) still called `.orThrow()` on a method whose "never throws" contract had changed earlier in this branch (`isCriticalFailure` now lets a critical failure propagate, per this ticket's own scope), without being updated to handle the newly possible failure. One now falls back to the raw project key regardless of failure, matching its own inline comment; the other two wrap the failure in a `CommandFailedError` with a remediation hint, matching every other failure path in the same files.
- A mechanical `unwrapOrThrow(x)` → `x.orThrow()` migration briefly introduced a `return;` / orphaned-expression ASI bug in three files, caught by `tsc` and confirmed via the integration suite.
- Two now-dead `try/catch` blocks (left over from before their callees stopped throwing) removed in `src/core/vortex/entitlement.ts` and `src/commands/integrate/_common/preflight-summary.ts`.

## Test plan

`bun run typecheck`, `bun run lint`, `bun run test:unit`, and the full
`bun run test:integration` all pass. No change in CLI behavior visible to users.


src/core/server/http-client.ts
```diff
@@ -188,44 +193,50 @@ export class SonarHttpClient {
...
     }
 
-    await this.raiseForStatus(response, method);
+    const error = await this.buildStatusError(response, method);
+    if (error) {
+      return Err(error);
+    }
 
-    return await response.text();
+    return Ok(await response.text());
```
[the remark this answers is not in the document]
charafsalmi: Confirmed and fixed in d4088306. Every `Result`-returning method on `SonarHttpClient` now wraps its body in `try/catch`, converting a thrown transport failure (network/DNS/timeout, malformed body) into `Err(...)` instead of letting it propagate as an uncaught rejection. This restores parity with the pre-`Result` `try/catch`-wrapped behavior for every swallow-to-fallback call site.

src/core/server/http-client.ts
```diff
@@ -284,7 +298,7 @@ export class SonarHttpClient {
     endpoint: string,
     params: Record<string, string>,
     timeoutMs: number = POST_REQUEST_TIMEOUT_MS,
-  ): Promise<void> {
+  ): Promise<Result<void>> {
```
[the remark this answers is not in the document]
charafsalmi: Fixed in d4088306 — reworded the doc comment to describe the `Result`/`Err` contract instead of the old throwing behavior.

src/core/result.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function Err<E extends Error>(error: E): Result<never, E> {
+  return { ok: false, error };
+}
+
+export function unwrap<T, E extends Error>(result: Result<T, E>): T {
+  if (!result.ok) {
+    throw result.error;
+  }
+  return result.value;
```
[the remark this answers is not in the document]
charafsalmi: Added `tests/unit/core/result.test.ts` in d4088306, covering `Ok`/`Err` construction and `unwrap` (including that it re-throws the exact `Error` instance, not a copy).

charafsalmi:
@gitar-ai Addressed in a99cf801:

- **Finding 1** (`NetworkConfigError` silently downgraded to a generic `TransportError`, dropping its remediation hint) — fixed. `toError` now passes `NetworkConfigError` through unwrapped.
- **Finding 5** (`isCriticalFailure` missed 500/502/504) — fixed. Added a `ServerError` class for any 5xx other than 503, wired into `buildStatusError` and `isCriticalFailure`.
- **Finding 6** (`toError`'s doc claimed it never sees a parsed-response failure) — fixed. Softened the doc to acknowledge a body read/parse failure on an otherwise-ok response also lands there.
- **Finding 3** (`getSafe` inconsistency vs. its siblings) — added the suggested NOTE doc comment. Behavior is unchanged: `getSafe` staying throwing-on-transport-failure is intentional and out of scope for this PR (its callers already have their own try/catch).
- **Finding 4** (`postForm` doc's inaccurate "malformed body" clause) — fixed, doc now matches behavior.
- **Finding 2** (`isCriticalFailure` not wired into existing swallow-to-fallback call sites like `fetchOrganizationInfo`) — deliberately left as-is. This ticket's scope is the categorization primitive itself, opt-in for callers to use ("so callers can decide" per the ticket). Retrofitting every existing swallow site to use it is a separate, larger change and stays out of this PR.

Added test coverage for the new `ServerError` critical-failure cases and the `NetworkConfigError` passthrough. Full verification (format/lint/typecheck/unit tests, 2634 passing) and Vortex analysis are clean.

gitar unblock

charafsalmi:
gitar unblock

src/core/server/errors.ts
```diff
@@ -86,3 +99,23 @@ export class RequestPayloadTooLargeError extends Error {
...
+ *
+ * `Result`-returning `SonarHttpClient` methods don't apply this themselves: they hand
+ * back whichever error they built, critical or not, and leave the decision to the
+ * caller. Swallow-to-fallback call sites should check this before discarding an error,
+ * so an outage or a misconfigured proxy fails loudly instead of looking like the normal
+ * "not found" case it is being folded into.
+ */
+export function isCriticalFailure(error: Error): boolean {
+  return (
+    error instanceof TransportError ||
```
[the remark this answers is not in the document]
charafsalmi: Fixed — isCriticalFailure is now wired into every swallow-to-fallback site with this shape, not just fetchOrganizationInfo: fetchOrganizationInfo, checkBillingEntitlement (both in organizations.ts), and checkComponent / getComponentId (components.ts) now rethrow when the discarded error is critical instead of folding it into their fallback value. The other swallow sites in this PR (resolveOrganizationAccess, sca.ts's getScaEnablement, remediate-api.ts's checkAiRemediationEntitlement, sqaa-analysis-client.ts's createAnalysis) already distinguish a failure from a genuine empty/absent result through their own return shape (a 3-state status, or an unconditional rethrow), so they didn't need the classifier.

src/core/result.ts
```diff
@@ -0,0 +1,87 @@
...
+// reaches the end of any chain built anywhere in the codebase. If a future neverthrow
+// version ships its own `orThrow` with an incompatible signature, the `declare module`
+// interface merge above will fail to compile rather than silently disagreeing with it.
+//
+// A prototype method can't itself be generic over the instance's T/E (those are erased
+// at runtime, and TypeScript has no way to attach a per-call type parameter to an
+// existing property slot) — the `any`s below are that, not a shortcut; the typed surface
+// callers actually see is the `declare module` block above.
+/* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-return */
+ResultAsync.prototype.orThrow = function (this: ResultAsync<any, any>): Promise<any> {
```
charafsalmi: @damien-urruty-sonarsource : 👀

src/commands/remediate/remediate-api.ts
```diff
@@ -49,45 +50,46 @@ export class RemediateApiClient {
...
-      const org = orgs.at(0);
-      if (!org) return { status: 'not_eligible' };
-
-      const configEndpoint = `/fix-suggestions/organization-configs/${org.id}`;
-      const config = await this.client.get<{
-        codeReviewAgent: { organizationEligible: boolean; delegateIssuesEnabled?: boolean };
-      }>(configEndpoint, undefined, this.client.apiHostFor(configEndpoint));
+      )
+      .andThen((organizations) => {
+        const org = organizations.at(0);
```
damien-urruty-sonarsource: Nitpick: I would try to keep those andThen closures a bit more compact, else it's harder to read the flow. Maybe we could extract long bodies into separate functions,

src/core/server/components.ts
```diff
@@ -33,49 +43,43 @@ export class ComponentsClient {
...
-      await this.client.get('/api/components/show', { component: projectKey });
-      return true;
-    } catch {
-      return false;
-    }
+  checkComponent(projectKey: string): ResultAsync<boolean, HttpClientError> {
+    return this.client
+      .get('/api/components/show', { component: projectKey })
+      .map(() => true)
+      .orElse((error) => (isCriticalFailure(error) ? errAsync(error) : okAsync(false)));
```
damien-urruty-sonarsource: This seems to be a change of behavior, in the middle of this refactoring PR. Is it really needed and safe at this point? 

How do we decide what is critical or not? For example, if the server returns a 400, shouldn't it be considered a critical failure?
My 2c: I think the previous behavior was wrong by always swallowing, but maybe only 404 should be considered a special case, and all others should rethrow

src/core/server/components.ts
```diff
@@ -33,49 +43,43 @@ export class ComponentsClient {
...
 
-  async hasProjectBeenAnalyzed(projectKey: string): Promise<boolean> {
-    const result = await this.client.getOrNotFound<{ analyses?: unknown[] }>(
-      '/api/project_analyses/search',
-      { project: projectKey, ps: 1 },
-    );
-    return (result?.analyses?.length ?? 0) > 0;
+  hasProjectBeenAnalyzed(projectKey: string): ResultAsync<boolean, HttpClientError> {
+    return this.client
+      .getOrNotFound<{ analyses?: unknown[] }>('/api/project_analyses/search', {
```
damien-urruty-sonarsource, line 68: Unrelated to your changes, but I'm not sure I understand the meaning of `getOrNotFound`. Could we find a better name?

src/core/server/project-bindings.ts
```diff
@@ -36,106 +38,138 @@ export class ProjectBindingsClient {
...
+        .get<{
+          projectBindings: Array<{ projectKey: string; repository: string }>;
+          page: { total: number; pageSize: number; pageIndex: number };
+        }>('/api/v2/dop-translation/project-bindings', {
+          pageSize,
+          pageIndex,
+          dopSettingId,
+        })
+        .andThen((result) => {
+          for (const binding of result.projectBindings) {
```
damien-urruty-sonarsource: Same here, the body is a bit long

tests/unit/core/server/project-bindings.test.ts
```diff
@@ -48,7 +48,7 @@ describe('ProjectBindingsClient', () => {
       fetchSpy = mockFetch({
         projectBindings: [{ projectId: 'proj:123', projectKey: 'my-project' }],
       });
-      const key = await client.getProjectKeyByGitRemote(remoteUrl);
+      const key = await client.getProjectKeyByGitRemote(remoteUrl).orThrow();
```
damien-urruty-sonarsource, line 51: As mentioned during our discussion, this orThrow() in all test methods annoys me. I think we should aim form something like:

```
const result = await client.getProjectKeyByGitRemote(remoteUrl);
expect(result).toBeOkWith('my-project');
```

But this can come later

package.json
```diff
@@ -105,6 +105,7 @@
     "commander": "15.0.0",
     "js-yaml": "5.4.1",
     "launchdarkly-node-client-sdk": "3.4.3",
+    "neverthrow": "8.2.0",
```
damien-urruty-sonarsource, line 108: Also later, we could introduce the `eslint-plugin-neverthrow` dependency to check we don't forget to consume a result

damien-urruty-sonarsource (approved):
LGTM, some suggestions for improvement

src/commands/remediate/remediate-api.ts
charafsalmi: Good call. Extracted into a private method (`resolveEntitlementForOrg`). Reads as a pipeline now instead of one long closure.

src/core/server/components.ts
charafsalmi: You're right. This is a real behavior change.

Checked the only caller. It already collapses both branches to the same "not accessible" status. So it has zero effect today.

Still, I agree `isCriticalFailure` was the wrong tool here.

Removed `checkComponent`. Its one caller now goes through `componentExists`. That method already does what you suggested: only 404 is special, everything else propagates. It's already covered by tests for 429/503/403.

Same fix applied to `getComponentId`.

src/core/server/components.ts
charafsalmi: Agreed, not the clearest name. It's pre-existing on master though, so I'd rather not fold an unrelated rename into this diff. Happy to do it as a quick follow-up if you want (`getIfFound`?).

src/core/server/project-bindings.ts
charafsalmi: Extracted the pagination merge/last-page logic into two module-level helpers (mergeBindingsPage, isLastPage), same pattern as requireSingleBinding already in this file.

tests/unit/core/server/project-bindings.test.ts
charafsalmi: Filed as CLI-1086 (https://sonarsource.atlassian.net/browse/CLI-1086), which also covers propagating Result through command handlers so orThrow() stops being scattered at every call site, not just in tests.

package.json
charafsalmi: Good idea for later. Added as a note in CLI-1086.

charafsalmi:
Confirmed and fixed in 574efdc2. My conversion to `.match()` only guarded the Err rail; a throw building entries from a malformed payload rejected the promise instead of degrading gracefully. Restored the try/catch + `orThrow()` shape already used by the sibling issues/duplications enrichment functions, and added a regression test that reproduces the exact TypeError this comment cites.

PR #786, opened 2026-09-07
CLI-1067 Build categories foundation

Krosovok:
1. Split the tests for breakdown into separate files so that it's easier to write tests in parallel.
2. Adjust data structure for the metrics mapping.
3. Act on feedback: move the values in front in the table view.


kirill-knize-sonarsource (approved):
LGTM

PR #787, opened 2026-09-07
CLI-1066 Colocate types and errors with their consumers

damien-urruty-sonarsource:
## Summary
- Move `sqaa-progress` next to analyze, `command-error` into `src/core/commands`, and `formatTable` into `list/issues`
- Fold `streams.ts` into private `TerminalConsole` methods, and drop the `src/core/ui` barrel and `types.ts` by colocating those types with their owners
- Split the git-hook example token literal in `shared.ts` so restaging that import change does not trip the pre-commit secrets scan


charafsalmi (approved):
LGTM

charafsalmi (approved):

PR #793, opened 2026-09-07
CLI-1077 Reject non-Server connections before prompting for a GitLab token

charafsalmi:
## Summary
- `sonar admin onboard-ci gitlab` on a SonarQube Cloud connection used to prompt for a GitLab personal access token before rejecting the connection with "requires a SonarQube Server connection" — the user pastes a secret for nothing.
- Extracted the connection-type check into an exported `assertOnPremiseConnection()` guard (`src/commands/admin/onboard-ci/gitlab/index.ts`) so the message is defined once, and call it in the command handler (`src/commands/command-tree.ts`) before `resolveGitlabToken()`. `preflight()` still calls the same guard, so behavior for the on-premise path is unchanged.

## Test plan
- [x] Added an integration test asserting a Cloud connection exits 1 without ever printing the GitLab token prompt (interactive session, mock TTY).
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run test:unit`
- [x] `bun test tests/integration/specs/admin/onboard-ci-gitlab.test.ts` (48 pass)
- [x] `sonar analyze agentic --depth DEEP` on changed files — clean (one pre-existing file-size finding on `command-tree.ts`, unrelated to this change)

CC @sonarsource/orchestration-workflow-squad for review — this path is CODEOWNERS-owned by your squad.

Jira: https://sonarsource.atlassian.net/browse/CLI-1077

### DIFF
--- a/tests/integration/specs/admin/onboard-ci-gitlab.test.ts
+++ b/tests/integration/specs/admin/onboard-ci-gitlab.test.ts
@@ -112,11 +112,12 @@ describe('sonar admin onboard-ci gitlab', () => {
       harness.withAuth(sqsServer.baseUrl(), 'test-token', 'my-org');
       harness.withExtraEnv({ SONARQUBE_CLI_MOCK_TTY: '1', GITLAB_TOKEN: '' });
 
-      const session = harness.runInteractive(`admin onboard-ci gitlab --group ${GROUP}`);
-      const result = await session.waitFinish();
+      const result = await harness.run(`admin onboard-ci gitlab --group ${GROUP}`);
 
       expect(result.exitCode).toBe(1);
-      expect(result.stderr).toContain('SonarQube Server');
+      expect(result.stderr).toContain(
+        'sonar admin onboard-ci gitlab requires a SonarQube Server connection.',
+      );
       expect(result.stdout).not.toContain('GitLab personal access token');
     });
 
--- a/tests/integration/specs/admin/onboard-ci-gitlab.test.ts
+++ b/tests/integration/specs/admin/onboard-ci-gitlab.test.ts
@@ -107,6 +107,19 @@ describe('sonar admin onboard-ci gitlab', () => {
       expect(result.stderr).toContain('SonarQube Server');
     });
 
+    it('exits 1 when connected to SonarQube Cloud without prompting for a GitLab token', async () => {
+      const sqsServer = await harness.newFakeServer().withAuthToken('test-token').start();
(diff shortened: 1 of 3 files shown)


tests/integration/specs/admin/onboard-ci-gitlab.test.ts
```diff
@@ -107,6 +107,19 @@ describe('sonar admin onboard-ci gitlab', () => {
       expect(result.stderr).toContain('SonarQube Server');
     });
 
+    it('exits 1 when connected to SonarQube Cloud without prompting for a GitLab token', async () => {
+      const sqsServer = await harness.newFakeServer().withAuthToken('test-token').start();
+      harness.withAuth(sqsServer.baseUrl(), 'test-token', 'my-org');
+      harness.withExtraEnv({ SONARQUBE_CLI_MOCK_TTY: '1', GITLAB_TOKEN: '' });
+
+      const session = harness.runInteractive(`admin onboard-ci gitlab --group ${GROUP}`);
+      const result = await session.waitFinish();
```
damien-urruty-sonarsource: This can be replaced with a simple:

> const result = harness.run(`admin onboard-ci gitlab --group ${GROUP}`);

If the agent was not able to figure this out, we might need to add some guidance in the agent file

tests/integration/specs/admin/onboard-ci-gitlab.test.ts
```diff
@@ -107,6 +107,19 @@ describe('sonar admin onboard-ci gitlab', () => {
...
+
+      expect(result.exitCode).toBe(1);
+      expect(result.stderr).toContain('SonarQube Server');
```
damien-urruty-sonarsource: Why not asserting the full message?

damien-urruty-sonarsource (approved):

tests/integration/specs/admin/onboard-ci-gitlab.test.ts
charafsalmi: Good catch, fixed, using harness.run() now. The manual runInteractive()/waitFinish() was leftover copy-paste from another spec, not a case where harness.run() doesn't fit.

tests/integration/specs/admin/onboard-ci-gitlab.test.ts
charafsalmi: Done, now asserting the full error message (sonar admin onboard-ci gitlab requires a SonarQube Server connection.) instead of the partial match.

PR #794, opened 2026-09-07
CLI-1076 Use OrganizationsClient for the AI remediation entitlement lookup

charafsalmi:
## Summary

`checkAiRemediationEntitlement` queried `/organizations/organizations` itself instead of going through `OrganizationsClient`, because that client used to hide network failures behind a `null`. CLI-844 (#782) fixed that: `OrganizationsClient.fetchOrganizationInfo` now throws on a critical/transport failure (via `isCriticalFailure`) and only returns `null` for an ordinary negative answer. The workaround comment (already removed from master in #782) can go too.

`RemediateApiClient` now exposes `organizations: OrganizationsClient` as a `readonly` field alongside `issues`/`components`, per the existing convention — no forwarding methods.

**Decision (per CLI-1076):** a non-critical failure (e.g. a 403 on the organization) is treated the same as "not found" — reported as `not_eligible`. This is the simpler of the two options laid out in the ticket, and matches "a 403 on the organization means the caller is not eligible".

Out of scope: the `/fix-suggestions/organization-configs/{id}` call stays as-is — remediation's own endpoint, not a shared client's domain.

Based on #782 (CLI-844) since this needs `isCriticalFailure`; will be retargeted to `master` once that merges.

## Test plan

- [x] `bun run lint` — 7 pre-existing warnings, unrelated
- [x] `bun run test:unit` (remediate) — 5 pass
- [x] `bun run test:integration` (remediate) — 41 pass

### DIFF
--- a/src/commands/remediate/remediate-api.ts
+++ b/src/commands/remediate/remediate-api.ts
@@ -21,11 +21,11 @@
 // Every SonarQube API call `sonar remediate` makes, in one place next to the command.
 
 import logger from '@/core/observability/logger.ts';
-import { okAsync, type ResultAsync } from '@/core/result.ts';
+import { okAsync } from '@/core/result.ts';
 import { ComponentsClient } from '@/core/server/components.ts';
-import type { HttpClientError } from '@/core/server/errors.ts';
 import { type SonarHttpClient } from '@/core/server/http-client.ts';
 import { IssuesClient } from '@/core/server/issues.ts';
+import { OrganizationsClient } from '@/core/server/organizations.ts';
 
 export interface AgentJobRequest {
   projectId: string;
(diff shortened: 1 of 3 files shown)


charafsalmi:
Applied Gitar's minimum-fix suggestion: `OrganizationsClient.fetchOrganizationInfo` now logs (debug) the swallowed non-critical failure before returning `null`, so an expired token / 403 / 404 is diagnosable instead of silently disappearing. Kept the boolean `not_eligible`/`unknown` split rather than the tri-state alternative — the ticket's explicit decision was to accept 403-as-not_eligible for simplicity, and a debug log line is enough to make the swallowed cause diagnosable without reopening that decision.

src/commands/remediate/remediate-api.ts
```diff
@@ -43,49 +43,44 @@ export class RemediateApiClient {
...
-        { organizationKey: orgKey, excludeEligibility: 'true' },
-        this.client.apiHostFor(orgsEndpoint),
-      )
-      .andThen((organizations) => this.resolveEntitlementForOrg(organizations.at(0)))
+    return this.organizations
+      .getOrganizationLegacyId(orgKey)
+      .andThen((orgId) => {
+        // A non-critical failure (e.g. a 403 on the organization) is treated the same as "not
+        // found": the caller is not eligible either way.
+        if (!orgId) return okAsync({ status: 'not_eligible' as const });
```
damien-urruty-sonarsource: Nitpick: I would argue the method was clearer when this was made in a separate function

damien-urruty-sonarsource (approved):

src/commands/remediate/remediate-api.ts
charafsalmi: Fair, extracted it back into a named `resolveEntitlementForOrgId` method in 243f1b2.

PR #795, opened 2026-09-07
CLI-1073 Move telemetry identity lookups onto domain API clients

charafsalmi:
## TL;DR

Telemetry resolved its four identity fields by calling `getSafe` directly, the last module outside the domain-client rule. Each lookup now sits behind the client that owns its area, with a new `EnterprisesClient` for the one that had none.

CLI-844 is why this lands now rather than with CLI-842: before it, `get<T>()` threw on any non-2xx, so a client method could not express the three states telemetry needs (value, confirmed-absent, retry later).

See : `UsersClient -> getCurrentUserId()`, `OrganizationsClient -> getOrganizationRecord()`, `EnterprisesClient -> getEnterpriseIdForOrganization()`, and `SystemClient -> getInstallationId()`.

## Worth challenging

The failure contract. The four methods return `Ok(null)` only for a field the server answered without, and `Err` for anything else, because telemetry caches `null` permanently as confirmed-absent.

`getOrganizationRecord` is shared with `getOrganizationId` and `getOrganizationLegacyId`, whose callers do want a non-critical failure collapsed to `null`. That swallow now lives in a private helper used only by those two. I checked their four callers, none changes behaviour, and tests pin both halves.

The alternative was a second public method for telemetry. I had it and dropped it: two identical signatures differing only in what `null` means.

## Already settled

- `FieldFetchResult` and `OrganizationLookupResult` are gone. Both were a `Result` re-encoded by hand: `resolved` is `isOk()`, `value` is `unwrapOr(null)`.
- `identity-api-mock.ts` built responses with no `status`, so the failure tests were passing through a `TypeError` instead of the status path. Fixed, with 403 cases added next to the 500s.

### DIFF
--- a/src/core/telemetry/identity-fetch.ts
+++ b/src/core/telemetry/identity-fetch.ts
@@ -25,13 +25,16 @@ import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 
 import type { ResolvedAuth } from '@/core/auth/auth-resolver.ts';
+import { errAsync, okAsync, type Result, type ResultAsync } from '@/core/result.ts';
+import { EnterprisesClient } from '@/core/server/enterprises.ts';
+import type { HttpClientError } from '@/core/server/errors.ts';
 import { SonarHttpClient } from '@/core/server/http-client.ts';
+import { type OrganizationRecord, OrganizationsClient } from '@/core/server/organizations.ts';
+import { SystemClient } from '@/core/server/system.ts';
+import { UsersClient } from '@/core/server/users.ts';
 
 import { getTelemetryDir } from '../config-constants.ts';
 import type { AuthConnection, ServerType } from '../state/state.ts';
-
-const ORGANIZATIONS_ENDPOINT = '/organizations/organizations';
-const ENTERPRISE_ORGANIZATIONS_ENDPOINT = '/enterprises/enterprise-organizations';
 
 export interface TelemetryIdentity {
   user_uuid: string | null;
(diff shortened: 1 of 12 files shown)


src/core/telemetry/identity-fetch.ts
```diff
@@ -210,25 +211,15 @@ interface FieldFetchResult {
...
 function fetchUserUuid(client: SonarHttpClient): Promise<FieldFetchResult> {
-  return client.getSafe<{ id: string }>('/api/users/current').match(
-    ({ response, value }): FieldFetchResult => {
-      if (!response.ok) {
-        return { value: null, resolved: false };
-      }
-      return { value: value?.id ?? null, resolved: true };
-    },
+  return new UsersClient(client).getCurrentUserId().match(
+    (value): FieldFetchResult => ({ value, resolved: true }),
```
damien-urruty-sonarsource: Nitpick: Do we need to explicitly type `FieldFetchResult` here? Can we do it once?

src/core/telemetry/identity-fetch.ts
```diff
@@ -210,25 +211,15 @@ interface FieldFetchResult {
...
     (): FieldFetchResult => ({ value: null, resolved: false }),
```
damien-urruty-sonarsource: I'm not sure this intermediate type `FieldFetchResult` is of much value

src/core/telemetry/identity-fetch.ts
```diff
@@ -237,43 +228,24 @@ function fetchOrganizationRecord(
...
-        if (!response.ok) {
-          return { uuidV4: null, id: null, resolved: false };
-        }
-        const org = value?.[0];
-        return { uuidV4: org?.uuidV4 ?? null, id: org?.id ?? null, resolved: true };
-      },
-      (): OrganizationLookupResult => ({ uuidV4: null, id: null, resolved: false }),
-    );
+  return new OrganizationsClient(client).getOrganizationRecord(orgKey).match(
+    (org): OrganizationLookupResult => ({
```
damien-urruty-sonarsource: Same here for `OrganizationLookupResult` and below in other methods.

damien-urruty-sonarsource (commented):
It looks good to me and I would approve, but there are Gitar findings. I will have another look when you address the comments

src/core/telemetry/identity-fetch.ts
charafsalmi: Correcting my earlier answer here: I first only dropped the redundant per-callback annotations and argued for keeping the interfaces. That missed your actual point. Both types are now gone.

`FieldFetchResult` was `{ value: string | null; resolved: boolean }`, which is `Result<string | null, E>` re-encoded by hand, one line after receiving an actual `Result` from the domain client. Same for `OrganizationLookupResult`. Since CLI-844 the clients return `ResultAsync`, so the four wrapper functions existed only to flatten a Result into a struct that the caller immediately unpacked again. `resolved` is `isOk()` and `value` is `unwrapOr(null)`.

`fetchMissingFromApi` now calls the clients directly and `resolveEnterpriseUuid` returns a `ResultAsync` like everything else. Net -50 lines in that file.

src/core/telemetry/identity-fetch.ts
charafsalmi: Removed. You were right that it carried no weight: it was a hand-rolled `Result`. The client returns `ResultAsync<string | null, HttpClientError>` and the wrapper immediately flattened it to `{ value, resolved }`, where `resolved` is just `isOk()` and `value` is `unwrapOr(null)`. The caller then unpacked the two fields one by one.

`fetchMissingFromApi` now awaits the client call directly:

    const user = await new UsersClient(client).getCurrentUserId();
    user_uuid = user.unwrapOr(null);
    resolved.user = user.isOk();

That also let the three `fetchXxx` wrappers go.

src/core/telemetry/identity-fetch.ts
charafsalmi: `OrganizationLookupResult` is gone too, same reasoning.

Chasing it down turned up something worse that I had introduced. To give telemetry the failure fidelity it needs, I had added a second public method `getOrganizationRecordOrFail` next to `getOrganizationRecord`: same signature, opposite meaning of `null`, and silently uncached. Exactly the naming problem you raised on #744, and a trap for the next caller.

Reversed it. There is now one public `getOrganizationRecord`, cached, where `Ok(null)` only ever means the server resolved no organization and every failure stays an `Err`. The swallow moved down into a private `organizationRecordOrNull` used only by `getOrganizationId` and `getOrganizationLegacyId`, which is where that contract was always documented. Behaviour for their callers is unchanged, and I pinned both halves with tests since the swallow changed location. The record type is now a named `OrganizationRecord` export instead of being spelled inline four times.

damien-urruty-sonarsource (approved):

PR #796, opened 2026-09-07
CLI-1078 Move command-tree helpers into core telemetry and SonarCommand

damien-urruty-sonarsource:
## Summary
- Keep `src/commands/` as a composition root with only `command-tree.ts` at the top level
- Inline `isTableFormatOption` into the command tree, move `commitTelemetryFacts` / `buildCommandExecutedFact` into `src/core/telemetry/index.ts`
- Derive command/subcommand on `SonarCommand.commandAndSubcommand()`, including passthrough remainder via `setPassthroughSubcommand()` instead of a WeakMap


charafsalmi (approved):
Good !

PR #799, opened 2026-09-07
CLI-1081 Harden generated CI output

nquinquenel:
----
## Summary by Gitar

- **CI generation hardening:**
    - Added proper shell and YAML quoting functions in GitLab CI template generation
    - Implemented validation to reject line breaks in project keys, server URLs, and scanner properties


src/commands/admin/onboard-ci/gitlab/index.ts
```diff
@@ -205,6 +205,13 @@ async function preflight(
   dopSettingKey: string;
   gitlabUrl: string;
 }> {
+  if (/[\r\n]/.test(auth.serverUrl)) {
+    throw new CommandFailedError('The configured SonarQube server URL must be a single line.', {
+      exitCode: 2,
+      remediationHint: "Re-authenticate with 'sonar auth login' or fix SONARQUBE_CLI_SERVER.",
+    });
+  }
```
damien-urruty-sonarsource, line 229: Shouldn't this check be made earlier? I find it strange to have it only for a specific command

src/commands/admin/onboard-ci/gitlab/templates.ts
```diff
@@ -32,6 +52,9 @@ export function generateCiYml(
   >,
   isNewFile = false,
 ): string {
+  assertSingleLine('project key', projectKey);
+  assertSingleLine('server URL', serverUrl);
```
damien-urruty-sonarsource, line 56: Same here. We could already reject the URL or the project key earlier when users do 'auth login' or when we try to discover the project key

src/commands/admin/onboard-ci/gitlab/index.ts
nquinquenel: Agree it would be nicer to reject a broken URL at login / env-var auth, but that's a separate change. Here this ticket is only about the SQ issue that got raised, and it does not apply to other commands. Happy to create a follow-up general hardening ticket, but I think it's worth keeping this PR contained to the issue it's trying to fix.

nquinquenel: https://sonarsource.atlassian.net/browse/CLI-1093

damien-urruty-sonarsource (approved):

PR #800, opened 2026-09-07
CLI-1080 Harden resource path handling

nquinquenel:
----
## Summary by Gitar

- **Security enhancements:**
    - Added symlink checks to prevent unauthorized file access across resource paths in `common.ts`
    - Filtered out files resolving outside the repository in change-set generation for `sqaa-changeset.ts`
- **Error handling:**
    - Skip unsafe symlink removals during feature cleanup in `installer.ts`


src/core/framework/features/installer.ts
```diff
@@ -264,8 +264,13 @@ export class IntegrationInstaller {
...
   ): Promise<void> {
     for (const resource of resolveAllResources(feature)) {
-      await resource.remove(context);
-      callbacks.onResourceRemoved?.(resource);
+      try {
+        await resource.remove(context);
+        callbacks.onResourceRemoved?.(resource);
+      } catch (error) {
+        context.console.warn(`Skipping removal of ${resource.id}: ${(error as Error).message}`);
+        callbacks.onResourceSkipped?.(resource);
```
vnaskos-sonar, line 272: Something worth checking: src/core/framework/features/installer.ts:267-273 and :365-371 — the new catch (error) around resource.remove(context) catches any error, not just the new symlink-safety rejection. 

Previously, any failure during removal (a corrupted JSON/YAML/TOML file a patch resource can't parse, an EACCES, a genuine bug) aborted the whole removeFeature loop loudly. Now it's downgraded to console.warn('Skipping removal of ...') and the loop continues — for any exception type. The PR's stated intent (Jira CLI-1080, and the PR's own Gitar summary) is "continue when a resource is skipped as unsafe," which reads as scoped to the symlink case specifically. As written, a real bug in a resource remover would now silently look identical to "this was an unsafe symlink, skipped it," which could mask problems during troubleshooting/support. Consider narrowing the catch to a dedicated marker (e.g. export a distinct error class/symbol from symlinkRejected in common.ts and check error instanceof UnsafeResourcePathError before swallowing, else rethrow). No test currently distinguishes "removal failed because of an unrelated bug" from "removal was skipped as unsafe" — only the symlink case is covered.

nquinquenel: Agree, that catch was too broad for the ticket. Uninstall now skips only `UnsafeResourcePathError` (the symlink-safety rejection) and rethrows everything else, with a test that a non-symlink remove failure still aborts the loop

PR #806, opened 2026-09-08
CLI-989 Migrate installed integrations to global scope

georgii-borovinskikh-sonarsource:
----
## Summary by Gitar

- **Reconciliation & Migration:**
    - Added automatic migration in `reconcile.ts` to fold project-scope integration installs into a single global record during reconciliation
- **Framework Changes:**
    - Removed dynamic function scope resolution from `resolveFeatureScope` to enforce static scopes or defaults
    - Added `scope` property support to `SubfeatureDeclaration` for subfeatures that must remain project-scoped


georgii-borovinskikh-sonarsource:
there's an issue with how this PR calls shouldInstall on global features/subfeatures when there's no existing global installed feature/subfeature. this needs to be resolved before the PR can be merged

PR #811, opened 2026-09-09
CLI-1089 Make sonar import command visible in the command list

bjorn-vonubisch-sonarsource:
Removes `{ hidden: true }` from the `sonar import` command registration and adds `rootHelp({ category: 'integrate' })` so it appears in the root help menu under the integrate group alongside other integration-related commands.


src/commands/command-tree.ts
```diff
@@ -330,9 +330,13 @@ function buildCommandTree(runtime: CliRuntime, console: Console): SonarCommand {
       qualityGateStatus({ ...options, file }, ctx),
     );
 
-  // Import repositories from DevOps platforms into SonarQube (hidden while in development)
-  COMMAND_TREE.command('import', { hidden: true })
-    .description('Import repositories from a connected DevOps platform into SonarQube')
+  COMMAND_TREE.command('import')
```
damien-urruty-sonarsource, line 333: You probably want to check the new `stage(Stage)` API we recently introduced on the command declaration builder. I think since you just introduced the command, it should be marked as alpha (or maybe beta)? More details [here](https://xtranet-sonarsource.atlassian.net/wiki/spaces/CodeRemedi/pages/5508530275/SonarQube+CLI+-+Feature+lifecycle)

damien-urruty-sonarsource (commented):
Waiting for your input regarding the feature stage before moving forward

PR #814, opened 2026-09-09
CLI-422 Split CLAUDE.md into nested files and path-scoped rules

charafsalmi:
`CLAUDE.md` reached 94 KB. Claude Code warns about it at every session start, it loads in full whatever you are working on, and most of it only matters in one part of the tree. The root keeps what applies everywhere and drops to 11 KB; everything else now loads on demand.

**This is a move, not a rewrite — so review the table below, not the diff.** The prose is carried over unchanged apart from heading levels, so reading the moved text line by line has no yield.

| Section of the old CLAUDE.md | Now lives in |
| --- | --- |
| About this project, Running checks, Writing code, Network access, Error handling, Documentation, Docs site, Vortex protocol | `CLAUDE.md` (root) |
| Commands: framework, per-command notes, system reset/status, deprecated aliases | `src/commands/CLAUDE.md` |
| Agent secrets hooks, Vortex, SQAA delivery, Context Augmentation, Claude hook dispatch, integrate scope and git hooks | `src/commands/integrate/CLAUDE.md` |
| Declarative integration engine, resource removers, container + subfeatures | `src/core/framework/CLAUDE.md` |
| Self-update lifecycle | `src/core/update/CLAUDE.md` |
| Telemetry | `src/core/telemetry/CLAUDE.md` |
| Tests | `tests/CLAUDE.md` |
| State, auth, project discovery | `.claude/rules/state-and-auth.md` |
| SonarQube API layer | `.claude/rules/sonarqube-api.md` |

**Two mechanisms, one criterion.** Both load on demand and neither costs anything at session start. A nested `CLAUDE.md` when the content's callers are a directory; a `.claude/rules/` file with a `paths:` list when they are not. The last two rows are the ones that are not: state/auth/discovery is called from `src/core/state/`, `src/core/auth/`, four top-level `src/core/` modules, four under `src/core/host/` and two commands, and API clients live both in `src/core/server/` and next to the commands that own them. A directory-shaped file would either miss callers or load 17 KB on anyone touching `src/core/telemetry/`.

**"Network access" is the one section that stayed in the root.** Its content is a repo-wide invariant — never call `fetch` directly, go through `fetch.ts` so proxy and TLS settings are applied — and it binds any code issuing an HTTP request, including the two call sites it names outside `src/core/server/`. Scoping it to where the clients live would have stopped it reaching exactly the places it governs.

**Nothing is lost, and it is checkable in ten seconds.** The files were generated from a line-range partition of the old file that refuses to run unless every line is claimed by exactly one destination. To verify independently:

```bash
git show origin/master:CLAUDE.md | sed -E 's/^#+ //; s/^- //; s/[[:space:]]+$//' | grep -v '^$' | sort -u > /tmp/a
cat CLAUDE.md tests/CLAUDE.md src/commands/CLAUDE.md src/commands/integrate/CLAUDE.md \
    src/core/*/CLAUDE.md .claude/rules/*.md \
  | sed -E 's/^#+ //; s/^- //; s/[[:space:]]+$//' | grep -v '^$' | sort -u > /tmp/b
comm -23 /tmp/a /tmp/b
```

It prints 12 paragraphs at the base this branch was last synced to. Eleven differ only by a cross-reference that used to say "see the section above" and now names the file that section moved to. The twelfth is the Documentation sentence, which gained a clause telling you to update the nested file rather than regrow the root. Anything beyond those is either a real loss or a paragraph master gained after the last sync — it will read as one or the other, since a paragraph added upstream has no near-match in the new files at all.

**Already settled, no need to re-open:**

- No skills. An earlier iteration put the cross-cutting parts in `.claude/skills/`; a skill is for a repeatable task, and this is pure context, so it goes in files that load by path instead.
- `AGENTS.md` stays a symlink to the root file. Codex, Cursor and Copilot therefore get neither on-demand mechanism, which is why the root carries an index.
- `sonar system status`, `self-update` and `verify` were filed under "State and auth" upstream. They are command notes and moved to `src/commands/CLAUDE.md`.


damien-urruty-sonarsource (commented):
As discussed, I'm not sure moving some of the content to skills is the right move. To me, skills are mostly about repeatable tasks/actions. Those skill files here seem to be purely about context. I think skills would be valuable, e.g. to explain how to contribute a new command, but pure context I would put it in a different place, maybe in subfolders' CLAUDE.md files as you did for some others

damien-urruty-sonarsource (approved):

PR #818, opened 2026-09-09
CLI-868 Warning on 'auth login' when env vars are set

tsiaraskon-sonar:
**Warn when logging in while environment-variable auth is active**

`sonar auth login` now detects when `SONARQUBE_CLI_TOKEN`-based env var authentication is active and warns the user before proceeding, since env vars always take precedence and would make any newly saved token unused. 

The user is then prompted to confirm (default "No"); declining cancels the login before the server is even contacted. If they proceed anyway, a reminder is printed after a successful login pointing out that the saved token won't be used until the env var is unset.

### DIFF
--- a/src/commands/auth/login.ts
+++ b/src/commands/auth/login.ts
@@ -19,7 +19,7 @@
  */
 
 import { recordConnectionFromAuth } from '@/core/auth/auth-connection-recorder.ts';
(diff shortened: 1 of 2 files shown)


vnaskos-sonar (approved):
👏

PR #821, opened 2026-09-09
CLI-1095 Rename getOrNotFound to getOrNullIf404

charafsalmi:
## Summary
- Renames `SonarHttpClient.getOrNotFound()` to `getOrNullIf404()` — purely mechanical, no behavior change. The old name reads as if the method decides between a value and "not found", or throws on any failure; the new name says exactly what the method does: resolve to `null` specifically on a 404, and propagate every other non-2xx status as a normal typed error.
- Clarifies the method's JSDoc to spell out that 403/429/503/network errors all still yield the normal typed error, exactly like `get`.
- Updates the definition, all five call sites (`components.ts` x3, `quality-gates.ts`, `pull-requests.ts`, `system.ts`), the method's own test `describe` block, and the mention in `CLAUDE.md`.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint:fix`
- [x] `bun run test:unit`
- [x] Vortex analysis (DEEP) on all touched files — clean (2 pre-existing findings on untouched lines in `http-client.ts` left as-is)

### DIFF
--- a/src/core/server/components.ts
+++ b/src/core/server/components.ts
@@ -79,7 +79,7 @@ export class ComponentsClient {
       queryParams.pullRequest = scope.pullRequest;
     }
     return this.client
-      .getIfFound('/api/components/show', queryParams)
+      .getOrNullIf404('/api/components/show', queryParams)
       .map((component) => component !== null);
   }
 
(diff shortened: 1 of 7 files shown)


src/core/server/components.ts
```diff
@@ -79,7 +79,7 @@ export class ComponentsClient {
       queryParams.pullRequest = scope.pullRequest;
     }
     return this.client
-      .getOrNotFound('/api/components/show', queryParams)
+      .getIfFound('/api/components/show', queryParams)
```
damien-urruty-sonarsource: After thinking a bit more about it, I'm not sure this name is better. `getIfFound` kind of state the obvious. A more precise name would be something like `getOrNullIf404` or `getOrNullIfNotFound`.

charafsalmi: I've gone through different ideas like getOrNull or getNullable404, but I kept getIfFound because it I found it simpler. Ok, I'll change to `getOrNullIf404`.

charafsalmi: Done

damien-urruty-sonarsource (approved):

PR #828, opened 2026-09-11
CLI-1098 Centralize auth resolution on AuthResolver and invocation context

damien-urruty-sonarsource:
## Summary

- Introduce `CliRuntime` with a memoized `AuthResolver` and `FlagsResolver`, wiring command handlers through `ctx.resolveAuth()` instead of a module-level `resolveAuth()` free function
- Make `ResolvedAuth` a class (`comesFromEnv()`), keep env/state resolution private on `AuthResolver`, and replace LaunchDarkly startup with flag resolution before command-tree visibility refresh
- Update unit tests to inject mocked `AuthResolver` instances via `tests/_common/mock-auth-resolver.ts`

PR #831, opened 2026-09-11
CLI-1072 Provide the SonarQube HTTP client through the command invocation context

charafsalmi:
## Summary

21 call sites each built their own `SonarHttpClient` out of `auth.serverUrl` / `auth.token`. `CommandAuthenticatedInvocationContext` now exposes a memoised `httpClient` getter and the authenticated handlers read it instead. Damien asked for this on #773.

**This buys ergonomics, not performance — the ticket oversold it and has been corrected.** `SonarHttpClient` holds three plain fields (URL, token, a boolean) and no per-request state, `fetch.ts` keeps no connection pool, and the cache that matters is `orgInfoCache`, an instance field of `OrganizationsClient`. Sharing the transport therefore shares no cache and saves no measurable work. What it buys is one answer to "where do I get a client", and a single place to change when construction moves behind a seam.

## Worth a second look

`SqaaAuth` (`analyze/sqaa-auth.ts`). The Vortex chain never receives a context and threads this value object through every function, so the client went in there. Once it did, nothing read `serverUrl` / `token` any more — across `sqaa-api.ts` and `sqaa-analysis.ts` the only reads are `orgKey` and `client` — so the last two commits drop the credentials. The type is now `{ orgKey?, client }`, which is a request target rather than an auth value, and its name is wrong. The rename lands in #837 instead of here: 65 occurrences across 10 files, in the area #828 was refactoring concurrently.

## Scope

- **No injection seam here.** The getter builds the client inline, so the 13 `SonarHttpClient.prototype` stubs in `tests/` stay as they are. #837 injects through `createCliRuntime()` rather than adding a fifth constructor parameter.
- **`preflight-summary.ts:113` / `:171` is where org lookups genuinely repeat** (two separate `OrganizationsClient`). Out of scope per the ticket, and it needs a shared *domain* client, not a shared transport: CLI-603 owns that.
- The two hook chains (`git pre-commit`, `agent session-start`) have no authenticated context, so they build the client once where auth is confirmed and pass it down.
- `src/core/vortex/entitlement.ts` is listed out of scope here; #837 reopens that one file so its two resolvers receive a transport instead of constructing one.


src/commands/import/index.ts
```diff
@@ -21,19 +21,15 @@
...
 import { runWithConcurrencyLimit } from '@/core/concurrency/concurrency-pool.ts';
-import { SonarHttpClient } from '@/core/server/http-client.ts';
 import type { Console } from '@/core/ui/console.ts';
 
-import {
-  type DopRepository,
-  ImportApiClient,
-  type ProvisionedProject,
-} from './_common/import-api.ts';
+import { type DopRepository, ImportApiClient, type ProvisionedProject } from './import-api.ts';
```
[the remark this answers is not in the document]
charafsalmi: Fixed in 49e359d8. Both this and the transport-convention finding were required by the repo rule to keep CLAUDE.md in step with structural changes, which I had missed.

src/core/commands/invocation-context.ts
```diff
@@ -142,4 +145,14 @@ export class CommandAuthenticatedInvocationContext extends CommandInvocationCont
   ) {
     super(console, lifecycle, runtime);
   }
+
+  /**
+   * Shared `SonarHttpClient` for this invocation, built lazily from {@link auth} and
+   * memoised for its lifetime. Every domain client constructed during the invocation
```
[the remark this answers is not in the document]
charafsalmi: Documented in 49e359d8, with one change to the suggested wording.

I did not carry over "which is what keeps OrganizationsClient's organization cache effective instead of one cache per caller". That sentence predates this PR, and checking it while writing the docs showed it is wrong: orgInfoCache is an instance field of OrganizationsClient, so sharing a SonarHttpClient shares no cache. The rule that actually does that is the next one in the same section (sharing the domain client as a readonly field). The paragraph now says sharing the transport buys uniformity rather than caching, and points at that next rule.

Worth flagging because it was also the ticket's stated justification, so I have corrected the ticket and the PR description too.

charafsalmi:
## Decision record: two items deliberately deferred until #828 lands

Writing this down because both decisions depend on context that is not visible in the diff, and neither is obvious from the code alone. If you are picking this up cold, read this before changing either thing.

### Background: what this PR actually buys

Worth stating plainly, because the ticket originally claimed otherwise and I have since corrected it.

Sharing a `SonarHttpClient` instance has **no runtime effect**. The transport holds three scalars (`serverURL`, `isCloud`, `token`) and no per-request state; `core/server/fetch.ts` keeps no connection pool or dispatcher. The cache people assume is involved is `orgInfoCache`, an **instance field of `OrganizationsClient`** (`core/server/organizations.ts`), so sharing the transport shares nothing. The claim that it does came from the `SonarQube API layer` section of `CLAUDE.md`, written during CLI-842, and this PR corrects it there.

So the value of this PR is ergonomic: one answer to "where do I get a client", and the 21 call sites are migrated once so they do not need migrating again when construction moves. Keep that in mind when weighing anything below, because none of it is worth doing for performance reasons.

Related: the one place where organization lookups genuinely repeat is `integrate/_common/preflight-summary.ts:113` and `:171`, which build two separate `OrganizationsClient` instances. That is out of scope here, and fixing it needs a shared **domain** client, not a shared transport.

### Deferred 1: `SqaaAuth` is now misnamed

The Vortex chain never receives an invocation context. It threads a `SqaaAuth` value object through every function in `sqaa-api.ts`, `sqaa-run.ts`, `sqaa-json-report.ts` and `sqaa-analysis.ts`, so that is where the client had to go.

Reviewing that produced the useful finding of this PR: once the chain takes its client from `SqaaAuth`, nothing reads `serverUrl` or `token` any more. Across `sqaa-api.ts` and `sqaa-analysis.ts` the only reads are `orgKey` (3) and `client` (1). The client did not sit beside the credentials, it replaced them, and callers were passing the same credentials twice, once as fields and once inside the client built from them. Two sources of truth for where a request goes. Those fields are now dropped (`89804b06`), leaving `{ orgKey?, client }`.

That makes the name wrong: it is a request target, not an auth value. **Not renamed here on purpose.** It is 65 occurrences across 10 files, in exactly the area #828 (CLI-1098) is refactoring right now, and #828 may well change where this value comes from. Renaming now would be churn resolved against a moving target.

Note that `resolveSqaaAuth` / `resolveSqaaAuthAndProject` keep their names legitimately either way: they do validate auth (Cloud needs an organization) and they return the target.

### Deferred 2: no test-injection seam

CLI-1072 listed "unit tests can inject a fake instead of stubbing the transport class prototype" as a value, following Damien's remark on #773. **This PR does not deliver that.** `ctx.httpClient` memoises into a private field with no constructor parameter and no setter, so there is no seam, and the 13 `SonarHttpClient.prototype` stubs under `tests/` are untouched.

The reason for not adding one here: #828 introduces `createCliRuntime({ authResolver, flagsResolver, isPrivateBetaEnabled })`, a resolver container with an explicit test-override seam, and `AuthResolver` already memoises per process. That is the natural home for a client factory. Adding a fifth optional constructor parameter to `CommandAuthenticatedInvocationContext` now would invent a competing seam that we would then have to remove.

### What to do once #828 is merged

1. Decide whether the client is built by the context (as now) or resolved through `CliRuntime` alongside `authResolver`. If the latter, only the inside of the `httpClient` getter changes; the 21 call sites stay as they are, which is the whole point of having migrated them in this PR.
2. Add the injection seam wherever construction ends up, then convert the 13 prototype stubs. Worth its own ticket.
3. Rename `SqaaAuth` to something that says "request target". Mechanical, and cheaper once #828 has stopped moving that area.

src/core/commands/invocation-context.ts
```diff
@@ -146,6 +147,8 @@ export class CommandInvocationContext {
  * resolved auth for this invocation.
  */
 export class CommandAuthenticatedInvocationContext extends CommandInvocationContext {
+  private client?: SonarHttpClient;
```
damien-urruty-sonarsource, line 150: I think this field and the `httpClient()` method should belong to the parent class, and this is probably the reason why you still need to instantiate the client by hand in a couple of places, like hook handlers. A `CommandInvocationContext` is created for anonymous command handlers, but it does not mean the user is not authenticated, it can be the case. The only difference is that in the parent class, the `httpClient()` method should return `SonarHttpClient?`

src/core/commands/invocation-context.ts
```diff
@@ -162,4 +165,10 @@ export class CommandAuthenticatedInvocationContext extends CommandInvocationCont
   override resolveAuthOrNull(): Promise<ResolvedAuth> {
     return Promise.resolve(this.auth);
   }
+
+  /** Memoised for the invocation: domain clients are built from this, never from a fresh one. */
+  get httpClient(): SonarHttpClient {
```
damien-urruty-sonarsource, line 170: I wonder about the naming. It's not any http client, it's one that is preconfigured with the right URL and token, it corresponds to a specific server. It should not be used e.g. to send telemetry. It is highly cohesive with the auth, so maybe we should move the 2 to the same object? E.g. `connection`? It would be accessed as `ctx.connection.httpClient` and `ctx.connection.auth`

damien-urruty-sonarsource (approved):
LGTM, but I think we could go further

src/core/commands/invocation-context.ts
charafsalmi: You were right, it belongs to the parent. `CommandInvocationContext` now has `resolveConnection()`, which returns `SonarConnection | null` and is memoised for the invocation. It has to be async there since the parent resolves the auth first, and it returns `null` only when there are no usable credentials, not when the handler happens to be anonymous. The authenticated subclass keeps a synchronous `connection` getter because its auth is already in hand.

The hook handlers that built a client by hand now call `ctx.resolveConnection()`. What is left building one directly is all outside an invocation context: login (no auth yet), telemetry, and token validation against an explicit server and token.

Done in #846 rather than here, so the two PRs you already approved keep their diffs.

src/core/commands/invocation-context.ts
charafsalmi: Agreed, and that is what it became: `SonarConnection` in `src/core/server/connection.ts`, holding the auth and the client bound to its server, read as `ctx.connection.auth` and `ctx.connection.httpClient`.

The useful side effect is that helpers below a handler take the whole connection instead of a `(client, auth)` pair, so the organization key can no longer drift from the transport it belongs to. Telemetry is unaffected, it never went through the context.

Done in #846 rather than here, so the two PRs you already approved keep their diffs.

PR #832, opened 2026-09-11
CLI-422 Drop file-by-file descriptions an ls already gives

charafsalmi:
Stacked on #815. Last of the three.

The [CLAUDE.md guidance](https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md) lists "file-by-file descriptions of the codebase" among the things to keep out: they cost context on every load and go stale on every rename. Four passages were exactly that.

| Removed | Kept from the same passage |
| --- | --- |
| module inventory of `src/core/update/` | that the migrations run in a load-bearing order |
| the six files `sonar system reset` is split across | nothing else was in it |
| per-agent hook handler filenames | that Cursor's two read handlers map to `beforeReadFile` and `preToolUse` |
| the method list of `SonarHttpClient` | that every method goes through `fetchAuthenticated` |

The transport sentence is also corrected while being rewritten: since CLI-844 a non-2xx response is an error on a `ResultAsync`, not a throw. The convention itself is not restated here — `src/core/result.ts` documents it and two ESLint rules enforce it, so a copy in the docs would be the very thing #815 removes.

**Deliberately not touched:** the feature ids, matcher strings and subfeature names under `src/commands/integrate/`. That file is the largest of the set and its content does change often, which the same guidance warns about. But none of it is derivable from a directory listing, so cutting it would lose facts rather than duplication. If it should shrink, that is a separate call about what an agent needs to know, not a mechanical one.


charafsalmi:
Superseded by #817, which carries the same change and is the one GitHub's stack view tracks. Closing to keep a single chain: #814 → #815 → #817.

PR #836, opened 2026-09-14
CLI-1107 Quiet Vortex analysis when no project is configured

tomshafir-sonarsource:
----
## Summary by Gitar

- **Documentation updates:**
    - Updated `sqaa-instructions-feature.ts` to make Vortex analysis quiet when no project is configured


vnaskos-sonar (approved):
I am not fully sure about the `never run it again this session`. What if on the same session they edit files from a sonar enabled repo. But at the same time it can be tricky to control it otherwise.

PR #846, opened 2026-09-15
CLI-1072 Route the Cloud-without-org outcome through the resolution policy

charafsalmi:
Third and last follow-up on the CLI-1072 stack, stacked on #837.

`resolveSqaaTargetAndProject` documents itself as a resolver that reports what it found and leaves the skip-vs-fail decision to its caller, but the Cloud-without-organization case did not follow that: it threw from inside the resolver when `--project` was set, and warned and returned `null` otherwise. `no-org` now carries `explicitProject`, so `resolveSqaaContext` decides that case where it already decides the `no-project` one.

No behaviour change. The policy branches of `resolveSqaaContext` had no test; they have one now.


damien-urruty-sonarsource (approved):

charafsalmi:
Gitar review

PR #850, opened 2026-09-16
CLI-1111 Add durable SQLite storage layer for sonar stats

kirill-knize-sonarsource:
----
## Summary by Gitar

- **Storage & Database:**
    - Added SQLite-based persistent storage layer for stats in `src/core/stats/`
    - Implemented database migrations, connection management, and corruption auto-quarantine
- **Data Operations:**
    - Added support for recording stats events, rule message upserts, and seen-fingerprint deduplication


src/core/stats/db.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import { applyStatsMigrations } from './stats-migrations.ts';
+
+function openAndMigrate(dbPath: string): Database {
+  const db = new Database(dbPath, { create: true });
+  db.run('PRAGMA journal_mode = WAL');
+  db.run('PRAGMA busy_timeout = 200');
+  applyStatsMigrations(db);
+  return db;
+}
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: Deferring — needs an explicit close point to actually help (cache without a shutdown hook just trades "checkpoint per call" for "never checkpoint"), and that lifecycle belongs to CLI-1112's postAction wiring, not this storage-only ticket. Cost is also negligible at real scale: ~1-20 events per invocation, not thousands. Will revisit if profiling in 1112/1113 says otherwise.

tests/unit/build-scripts/compile-target-runtime.test.ts
```diff
@@ -0,0 +1,150 @@
...
+});
+
+describe('downloadCompileTargetExecutable()', () => {
+  const cleanups: Array<() => void> = [];
+
+  afterEach(() => {
+    for (const cleanup of cleanups.splice(0)) {
+      cleanup();
+    }
+  });
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: It's not the part of this PR. Why is it reported here?

kirill-knize-sonarsource:
> Other objectives on this issue, possibly covered elsewhere:
> 
> ◻️ Query event-class-specific fields using SQLite's JSON1 functions

Regarding this - it will be implemented in one of the following tickets. Scope for this one is storage only.

src/core/stats/db.ts
```diff
@@ -0,0 +1,80 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
```
Krosovok, line 20: `src/core/stats/stats-db.ts` and others repeat `stats` twice in the path. Do we need this redundancy? I think just `db.ts` in folder `stats` is telling enough. WDYT?

src/core/stats/db.ts
```diff
@@ -0,0 +1,80 @@
...
+    return db;
+  } catch (error) {
+    // Windows can't rename/delete a file with an open handle — release ours before the
+    // caller tries to quarantine it, or the retry will just fail on the same locked file.
+    db.close();
+    throw error;
+  }
+}
+
+function quarantineLedgerFiles(dbPath: string): void {
```
Krosovok, line 53: `archiveCorruptedLedgerFiles` or something along these lines maybe? "quarantine" has a meaning of taking an explicit action to limit/lock it somehow, at least for me.

src/core/stats/migrations.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+import type { Database } from 'bun:sqlite';
+
+export interface StatsMigration {
+  version: number;
+  up: (db: Database) => void;
+}
+
+const MIGRATIONS: StatsMigration[] = [
```
Krosovok, line 28: Have you considered having the migrations as separate SQL files?
Claude tells me that `.sql` can be bundled with the Bun binary if we explicitly import them. So we would have to import each of them. But at least TS and SQL are not mixed up in the same file too much this way.
Something like
```
import sql from './001_init.sql' with { type: 'text' }; 
```

tests/unit/core/stats/stats-db.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    // Windows can hold the just-closed db's WAL/SHM handles past our retries; best-effort
+    // only — the OS reclaims the temp dir regardless, same as tests/integration/harness/index.ts.
+    await rm(testSonarUserHome, {
+      recursive: true,
+      force: true,
+      maxRetries: IS_WINDOWS ? 15 : 5,
+      retryDelay: IS_WINDOWS ? 200 : 100,
+    }).catch(() => {});
+  }
+}
```
Krosovok: In other places (including `stats-store.test.ts`) it is done in form of `beforeEach`/`afterEach`. Shouldn't it be the same? There are a few other places in code where test `ENV_SONAR_USER_HOME` is set, and it's repeated over and over, mostly with the same comment. Maybe it's a good idea to move it to some common unit-test level helper? It's a bigger scope though.

src/core/stats/db.ts
kirill-knize-sonarsource: Renamed to `db.ts`, `store.ts`, and `migrations.ts` — matches the convention in `core/telemetry` and `core/auth`, neither of which prefixes filenames with the package name.

src/core/stats/db.ts
kirill-knize-sonarsource: I'd keep `quarantine` here — the function renames files that are suspected of corruption, without evidence yet that they're safe to discard. `archive` reads as routine backup/retention to me, which isn't what's happening: these files are set aside because they're suspect, not preserved for reuse. Happy to rename if you still feel strongly, but I think `quarantine` communicates the intent more precisely.

src/core/stats/migrations.ts
kirill-knize-sonarsource: Makes sense once we have more than one migration, but right now there's only `version: 1` — splitting into separate files today would mean adding the `import ... with { type: 'text' }` machinery for a single file, which is more ceremony than the current code needs. Let's revisit this once we're adding the second or third migration.

tests/unit/core/stats/stats-db.test.ts
kirill-knize-sonarsource: Good catch — extracted a shared `useTempSonarUserHome` helper (`tests/unit/core/stats/_helpers.ts`), so both this file and `store.test.ts` now use the same `beforeEach`/`afterEach` setup, and the duplicated Windows WAL/SHM cleanup comment is gone. Kept the scope to the stats tests for now — rolling the same helper out to telemetry/auth/mcp tests is the bigger cleanup you flagged, worth its own PR.

Krosovok (approved):

PR #853, opened 2026-09-16
CLI-1123 Run post-update migrations from the root preAction hook

tomshafir-sonarsource:
----
## Summary by Gitar

- **Migrations:**
    - Moved post-update migration execution from `src/index.ts` to a root `preAction` hook in `src/commands/command-tree.ts`
    - Provided `runtime` in `PostUpdateDependencies` for migrations requiring authentication resolver access


src/core/update/post-update.ts
```diff
@@ -18,6 +18,7 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+import type { CliRuntime } from '@/core/commands/cli-runtime.ts';
```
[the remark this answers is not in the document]
tomshafir-sonarsource: To be used in the following "global integrate migration" PR

vnaskos-sonar (approved):
🚀

src/commands/command-tree.ts
```diff
@@ -911,6 +915,20 @@ function buildCommandTree(runtime: CliRuntime, console: Console): SonarCommand {
     if (state) initSentry(state);
   });
 
+  let postUpdateRan = false;
+  COMMAND_TREE.hook('preAction', async () => {
+    if (postUpdateRan) return;
+    postUpdateRan = true;
```
damien-urruty-sonarsource: why do we have this guard? Can the `preAction` hook be called several times? A small comment would help understand why. If it's just overly defensive code, let's remove it

src/commands/command-tree.ts
```diff
@@ -911,6 +915,20 @@ function buildCommandTree(runtime: CliRuntime, console: Console): SonarCommand {
...
+    // Safely: a throw from a Commander hook would abort the user's command.
+    await runPostUpdateActionsSafely({
+      supportedIntegrations,
+      claudeIntegrationId: CLAUDE_INTEGRATION_ID,
```
damien-urruty-sonarsource: Unrelated to your changes, but this parameter looks weird. Why do we have to pass this here? Do we ever call `runPostUpdateActionsSafely` with another `claudeIntegrationId`?

tests/integration/harness/index.ts
```diff
@@ -235,6 +235,23 @@ export class TestHarness {
...
   }
 
+  /** Ceiling for a post-update run that downloads the CAG binary. */
+  private static readonly CAG_POST_UPDATE_TIMEOUT_MS = 150_000;
+
+  /**
+   * Runs the cheapest CLI invocation that triggers post-update migrations.
+   * Pass `useCagTimeout` when the migration will re-download the CAG binary.
+   */
+  async runToTriggerPostUpdate(
```
damien-urruty-sonarsource: I'm not a big fan of multiplying the `run` functions in the harness. We already have 3 different primitives, and this one looks very specific and technical. To me, the harness should be about driving program execution, as a user would do. Mentioning "trigger post update" exposes low-level detail about how we manage execution. Could we avoid introducing it? I think you could use `run` as it was used before, I see something specific for the CAG timeout, couldn't we set it once for the whole harness and not bother about it?

src/core/update/post-update.ts
```diff
@@ -100,6 +103,14 @@ export async function runPostUpdateActions(deps: PostUpdateDependencies): Promis
   }
 }
 
+export async function runPostUpdateActionsSafely(deps: PostUpdateDependencies): Promise<void> {
+  try {
+    await runPostUpdateActions(deps);
+  } catch (error) {
+    logger.debug(`Post-update actions failed: ${(error as Error).message}`);
```
damien-urruty-sonarsource: Is it sufficient to log in debug mode? Shouldn't we be a bit more verbose and signal the user?

src/commands/command-tree.ts
tomshafir-sonarsource: I think it's the legacy Claude migrations that we are still dragging with us. Will see if it can be removed from the parameters list.

damien-urruty-sonarsource (approved):

PR #856, opened 2026-09-17
CLI-1127 Keep tar archive names relative so GNU tar accepts Windows paths

damien-urruty-sonarsource:
## Summary
- Windows unit tests on master flaked because GNU tar treats `C:` as a remote host when packing/extracting with an absolute path
- Invoke tar with a relative archive name and `cwd` set to the temp dir in the compile-target fixture and in `downloadCompileTargetExecutable`


vnaskos-sonar (approved):
🙏

PR #860, opened 2026-09-17
CLI-1129 Match unbracketed IPv6 addresses in SONAR_NO_PROXY

damien-urruty-sonarsource:
## Summary
- Parse IPv6 `SONAR_NO_PROXY` entries the way curl and Go do: unbracketed `::1`, and `[IPv6]:port` only when a port is set
- Compare against `URL.hostname` with brackets stripped so `::1` actually bypasses the proxy


src/core/host/connectivity/network-config.ts
```diff
@@ -349,14 +349,37 @@ function matchesNoProxy(hostname: string, port: number, noProxy: string): boolea
...
+    return {
+      host: bracketedWithPort[1].toLowerCase(),
+      port: Number.parseInt(bracketedWithPort[2]),
+    };
+  }
+
+  const host = stripIpv6Brackets(entry).toLowerCase();
+  if (isIpv6Literal(host)) {
+    return { host, port: 0 };
+  }
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: Done — IPv6 noProxy hosts now go through the URL parser (`canonicalizeIpv6`) so expanded, compressed, and IPv4-mapped spellings match. Added coverage for `0:0:0:0:0:0:0:1`, `2001:0db8::1`, and `::ffff:192.168.1.10`.

src/core/host/connectivity/network-config.ts
```diff
@@ -349,14 +349,44 @@ function matchesNoProxy(hostname: string, port: number, noProxy: string): boolea
     });
 }
 
+// Accepted forms: hostname, IPv4, unbracketed IPv6 (`::1`), and `[IPv6]:port` when
+// a port is set (curl/Go). IPv6 is canonicalized through the URL parser so expanded
+// and compressed spellings match. A naive last-`:digits` split would turn `::1`
+// into host `:` / port `1`.
```
Krosovok: I'd remove this comment. The tests and code already say that. I'd rather leave one-line comment near `/^\[([^\]]+)]:(\d+)$/` regex explaining what it is.

Krosovok (approved):

src/core/host/connectivity/network-config.ts
damien-urruty-sonarsource: Done — dropped the block comment and left a one-liner on the `[IPv6]:port` regex.

PR #862, opened 2026-09-17
CLI-1131 Fix Vortex keep/remove decision not propagating to sibling hook features

vnaskos-sonar:
## Summary

Found while manually testing CLI-990: declining to keep an already-installed Vortex on re-run (`sonar integrate` / `sonar integrate claude`) removed the umbrella "Vortex" feature, but Claude's `PostToolUse` dispatch container (and Codex's SQAA hook subfeature) independently re-derived their own install decision from raw org entitlement, ignoring what the user had just answered — so the hook silently got reinstalled anyway.

- Added `resolvedFeatureDecisions` to `IntegrationInvocation`, populated as each top-level feature's decision resolves during the selection pass.
- `vortexInstallDecision()` now reads Vortex's own just-resolved outcome from that map instead of re-deriving it from entitlement, so every Vortex-gated feature agrees with the single Keep/Remove answer.
- Reordered Claude's `features` array so the "Vortex" umbrella (which owns the ask) resolves before the `PostToolUse` container that depends on it — this preserves the exact same visible prompt order as before (no user-facing reordering), while fixing the underlying inconsistency.

## Test plan
- [x] New regression test: declining Vortex removes it from both the umbrella feature and the PostToolUse container
- [x] `bun run test:unit` — 2755 pass
- [x] Integration suites for claude/codex/copilot/cursor/antigravity — 150 pass
- [x] `bun run lint` / `bun run typecheck` — clean


src/commands/integrate/_common/vortex.ts
```diff
@@ -112,8 +118,45 @@ export interface ResolvedVortexSetup {
...
+    return install();
+  }
+  if (resolved === 'uninstall') {
+    return uninstall();
+  }
+  if (resolved === 'skip' || resolved === 'declined') {
+    return skip();
+  }
+  return vortexInstallDecisionFromDisposition(invocation.options.vortexDisposition);
+}
```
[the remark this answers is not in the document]
vnaskos-sonar: Fixed in b96aa424: `vortexInstallDecision` now maps a resolved `'declined'` outcome to `uninstall()` (same as `'uninstall'`), so an already-installed sibling gets torn down instead of left in place. Added a regression test for exactly this case (stale PostToolUse container + fresh Install-Vortex? decline) — confirmed it fails without the fix.

src/core/framework/features/types.ts
```diff
@@ -60,6 +63,16 @@ export interface IntegrationInvocation<TOptions = Record<string, unknown>> {
...
+  /**
+   * Populated as each top-level feature's decision resolves during this
+   * invocation's selection pass (keyed by feature id) — lets a feature
+   * declared *later* in the integration's `features` array read an earlier
+   * sibling's resolved outcome instead of independently re-deriving it and
+   * risking disagreement (see `vortexInstallDecision` in
+   * `integrate/_common/vortex.ts`). Not populated yet for the feature whose
+   * own `shouldInstall` is currently being evaluated.
+   */
+  resolvedFeatureDecisions?: Map<string, FeatureSelectionOutcome>;
```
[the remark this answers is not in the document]
vnaskos-sonar: Fixed in b96aa424: documented the resolvedFeatureDecisions contract and declaration-order requirement in both src/commands/integrate/CLAUDE.md (Vortex paragraph) and src/core/framework/CLAUDE.md.

src/commands/integrate/claude/declaration.ts
```diff
@@ -78,6 +85,71 @@ export interface ClaudeIntegrationOptions extends IntegrateAgentOptions {
...
+      sqaaPostToolUseSubfeature,
+      cagPostToolUseSubfeature,
+    ],
+    resolveClaudeSkillPath,
+  );
+
+  return {
+    ...defaultVortex,
+    id: CLAUDE_VORTEX_FEATURE_ID,
+    replacedIds: [...(defaultVortex.replacedIds ?? []), VORTEX_FEATURE_ID, SQAA_HOOK_FEATURE_ID],
```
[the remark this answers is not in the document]
vnaskos-sonar: Confirmed this against the actual setup.js source (not just the summary) — collectDeclarativeHookNames() only checks top-level featureIds and never looks at subfeatures[], so this is a real break.

Fixed in SonarSource/sonarqube-agent-plugins#54: recognizes a `vortex-claude` feature carrying the `sqaa-posttooluse` subfeature as equivalent to the old top-level `sonar-sqaa-hook`, while keeping recognition of the old id for users still on a pre-merge CLI version. Verified against three simulated state.json shapes (merged container with/without the hook, and the legacy shape).

This PR should release together with (or after) that one, per your note.

src/commands/integrate/claude/declaration.ts
```diff
@@ -78,6 +85,65 @@ export interface ClaudeIntegrationOptions extends IntegrateAgentOptions {
...
+    resources: createPostToolUseDispatchResources({
```
georgii-borovinskikh-sonarsource: would be cleaner if this resource was defined in a constant like the other 2 subfeatures

tests/integration/specs/update/post-update.test.ts
```diff
@@ -515,17 +517,92 @@ describe('post-update migration', () => {
...
+          },
+        }),
+      );
+      harness.state().withContextAugmentationBinaryInstalled();
+
+      const result = await harness.run(POST_UPDATE_TRIGGER_COMMAND);
+
+      expect(result.exitCode).toBe(0);
+      expectFullClaudeVortexMigration();
+    },
```
georgii-borovinskikh-sonarsource: This test always seeds both legacy containers with their **full** subfeature set for an allowlisted org, so it can't distinguish "the merge carries over each old container's actual recorded per-subfeature state" from "the merge just recomputes from `defaultInstallSubfeatureIds` + `migrationEligible` and ignores the old containers' `.subfeatures` entirely" — which is what actually happens: `migrateReplacedFeatures`/`createFeatureApplication` in `reconcile.ts` always passes `subfeatureIds: undefined` for a replacement migration and never reads `predecessors[].subfeatures`, even though both predecessors here are container-shaped and carry that data.

Concretely, this test would pass identically if `sonar-sqaa-hook` were seeded missing `cag-posttooluse` for this same (currently-allowlisted) org — the merge would silently turn it back on via `migrationEligible` rather than preserving the old exclusion. Worth adding a variant that seeds partial/asymmetric predecessor state and asserts the merge reflects that old state, to actually pin down carry-over vs. recompute-from-defaults.

tests/integration/specs/update/post-update.test.ts
```diff
@@ -103,22 +106,20 @@ describe('post-update migration', () => {
...
-    const vortex = claude?.features.find((feature) => feature.featureId === VORTEX_FEATURE_ID);
     expect(vortex?.subfeatures?.map((subfeature) => subfeature.featureId)).toEqual([
       SQAA_INSTRUCTIONS_SUBFEATURE_ID,
       CONTEXT_AUGMENTATION_FEATURE_ID,
       CONTEXT_AUGMENTATION_HOOK_FEATURE_ID,
+      'sqaa-posttooluse',
+      'cag-posttooluse',
     ]);
     expect(harness.cwd.file('.claude', 'settings.json').asJson().hooks?.PostToolUse).toBeDefined();
     expect(harness.cwd.file('CLAUDE.md').asText()).toContain('# Vortex analysis protocol');
```
georgii-borovinskikh-sonarsource: This only asserts `hooks?.PostToolUse` is *defined*, not its matcher value — unlike the fresh-install tests in `claude.test.ts` (`'installs PostToolUse and PostToolUseFailure Context Augmentation hook...'` and `'does not install the Context Augmentation hook...'`), which pin the exact union-matcher string produced by `createPostToolUseDispatchResources`'s `resolveUnionMatcher` (full 6-tool union vs. `'Edit|Write'` alone).

None of the migration tests verify that the union-matcher logic actually renders the right matcher after a migration that excludes `cag-posttooluse` — the non-allowlisted-org migration test below only checks the recorded subfeature id list, not the resulting `settings.json` content. Worth asserting the matcher string here (and in that test) too.

tests/integration/specs/update/post-update.test.ts
```diff
@@ -515,17 +517,92 @@ describe('post-update migration', () => {
...
+        installedByCliVersion: '0.5.0',
+        installedAt: now,
+        updatedByCliVersion: '0.5.0',
+        updatedAt: now,
+        dependencies: [],
+        resources: [],
+        operations: [],
+        attrs: {
+          orgKey: CAG_HOOK_ALLOWED_ORG_KEY,
+          projectKey: 'p',
```
georgii-borovinskikh-sonarsource: This is the only test for merging two already-containerized legacy Vortex records, and it only covers the allowlisted-org branch. There's a non-allowlisted-org migration test above (`'excludes the CAG PostToolUse and PostToolUseFailure hooks when migrating for a non-allowlisted org'`), but it uses `seedPreUnificationFeatures` — the flat, pre-container legacy shape — not two already-containerized records.

So the one code path this PR actually introduces (folding two container-shaped predecessors together) has never been exercised with a non-allowlisted org. Worth adding a sibling test here with a non-allowlisted org key, asserting `cag-posttooluse`/`context-augmentation-hook` are excluded from the merged result.

tests/integration/specs/update/post-update.test.ts
```diff
@@ -515,17 +517,92 @@ describe('post-update migration', () => {
...
+  it(
+    'merges two already-containerized legacy Vortex records into the single vortex-claude container',
+    async () => {
+      // Two fully-installed containers, each with its own recorded subfeatures —
+      // contrast with the older pre-container flat records seeded above.
+      const now = new Date().toISOString();
+      const legacyContainer = (featureId: string, subfeatureIds: string[]) => ({
+        featureId,
+        scope: 'project' as const,
+        targetRoot: harness.cwd.path,
```
georgii-borovinskikh-sonarsource: Both legacy containers in this test are hardcoded to `scope: 'project'`. Vortex is documented as scope-agnostic (the container "pins neither `scope` nor `targetRoot`, following the invocation instead"), and other Claude tests exercise global-scope Vortex installs directly (e.g. `findClaudeFeature(harness, CLAUDE_VORTEX_FEATURE_ID, 'global')` in `claude.test.ts`) — but nothing proves this specific two-old-container merge behaves the same way when both predecessors were recorded at global scope instead of project scope.

src/commands/integrate/claude/declaration.ts
vnaskos-sonar: Fixed in abc304d4 — extracted into `claudePostToolUseDispatchResources`, defined right after the two subfeature constants it wires together.

tests/integration/specs/update/post-update.test.ts
vnaskos-sonar: Confirmed and fixed in abc304d4: added a test seeding `sonar-sqaa-hook` with only `sqaa-posttooluse` recorded for a currently-allowlisted org, asserting the merge turns `cag-posttooluse` back on — pinning down that this is recompute-from-current-eligibility, not carry-over, matching what you found in `reconcile.ts`.

tests/integration/specs/update/post-update.test.ts
vnaskos-sonar: Fixed in abc304d4: both `expectFullClaudeVortexMigration()` and the non-allowlisted-org migration test now assert the exact matcher string on `hooks.PostToolUse[0].matcher` (full 6-tool union vs. `Edit|Write` alone), matching the precision used in claude.test.ts.

tests/integration/specs/update/post-update.test.ts
vnaskos-sonar: Fixed in abc304d4: added a sibling test using the same two-already-containerized-predecessors shape for a non-allowlisted org, asserting cag-posttooluse/context-augmentation-hook are excluded and the settings.json matcher is Edit|Write alone.

tests/integration/specs/update/post-update.test.ts
vnaskos-sonar: Fixed in abc304d4: added a global-scope variant seeding both legacy containers at harness.userHome.path with scope: 'global', asserting the merge succeeds identically there (extended expectFullClaudeVortexMigration to take a scope/root option so it can check the right paths for either scope).

georgii-borovinskikh-sonarsource (approved):

PR #864, opened 2026-09-17
CLI-1112 Wire analyzers into stats ledger

kirill-knize-sonarsource:
----
## Summary by Gitar

- **New storage layer:**
    - Added durable SQLite storage layer with `stats-db.ts`, `stats-store.ts`, and `stats-migrations.ts`
    - Implemented event recording, rule message upserts, and seen-fingerprint deduplication for stats
- **Analyzer wiring:**
    - Wired `sonar-secrets`, `sqaa`, and `sca-scanner-cli` analyzers into the stats ledger
    - Added `commitStatsFacts` to handle post-action flushing and hook abort paths
- **Testing:**
    - Added comprehensive unit tests covering stats store, database management, migrations, and telemetry integration


src/core/commands/invocation-context.ts
```diff
@@ -157,6 +168,19 @@ export class CommandInvocationContext {
...
     return this.facts.slice();
   }
+
+  /** Record stats facts for `postAction` drain. */
+  recordStats(...facts: StatsFact[]): void {
+    if (facts.length === 0) {
+      return;
+    }
+    this.statsFactsBuffer.push(...facts);
+  }
```
Krosovok: Fro both telemetry and stats: I think we can drop `if`

src/core/commands/invocation-context.ts
```diff
@@ -78,6 +88,7 @@ const DISABLED_RUNTIME: CliRuntime = createCliRuntime({ authResolver: new NullAu
  */
 export class CommandInvocationContext {
   private readonly facts: TelemetryFact[] = [];
+  private readonly statsFactsBuffer: StatsFact[] = [];
```
Krosovok: Since they are mirroring each other, I think they should be named similarly to focus on difference:
`telemetryFacts` / `statsFacts` for example. We should not leave telemetry as just `facts` when there are 2 types of them.

src/core/stats/facts.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/** Records a stats event from an analyzer's own telemetry fact — the envelope is read off
+ *  it, so callers only ever supply `details` (their deduped/allowlisted finding counts). */
+export function recordAnalyzerStats(
+  ctx: CommandInvocationContext,
+  fact: TelemetryFact<AnalysisFactEnvelope>,
+  details: AnalyzerStatsDetails,
+): void {
+  ctx.recordStats(new StatsFact<AnalyzerStatsFactPayload>({ envelope: fact.payload, details }));
+}
```
Krosovok: This method feels a bit weird as it mutates `ctx`. It mostly is called alongside `ctx.recordTelemetry(...)`. I think it would be easier to follow if this method only did the building and callers would call `ctx.recordStats` themselves:
```
ctx.recordTelemetry(fact);
ctx.recordStats(buildStatsFromTelemetry(fact, {...}));
```

src/commands/analyze/secrets.ts
```diff
@@ -163,6 +164,61 @@ function buildSecretsAnalysisTelemetryFact(
...
+  file: string | undefined,
+  startLine: number | undefined,
+  startColumn: number | undefined,
+): string {
+  return `${ruleKey}|${file ?? ''}|${startLine ?? ''}|${startColumn ?? ''}`;
+}
+
+// rule_descriptions is a lookup, so upsert runs for every issue, not just newly-deduped ones.
+// `source` disambiguates issues with no `file`, so different --input scans don't collide.
+export function summarizeNewSecretsFindings(
```
Krosovok: Secrets also have a separate file for telemetry stuff, like the other 2: `secrets-analysis-telemetry.ts`. For some reason `buildSecretsAnalysisTelemetryFact` and other telemetry logic was not moved there. How about moving stats logic there together with some of the telemetry logic?

src/commands/hook/cursor-secrets-block.ts
```diff
@@ -45,10 +48,9 @@ export function secretsFoundInScan(result: { exitCode: number | null }): boolean
   return (result.exitCode ?? 1) === EXIT_CODE_SECRETS_FOUND;
 }
 
-/**
- * Deny a Cursor file-read/tool-use event with `message`, then exit.
- */
-export async function denyCursor(message: string): Promise<never> {
+export async function denyCursor(ctx: CommandInvocationContext, message: string): Promise<never> {
+  // process.exit() bypasses postAction, so drain stats here or this run's StatsFact is lost.
+  commitStatsFacts(ctx.statsFacts());
```
Krosovok: I can't find any reason why we are using `process.exit(CURSOR_BLOCK_EXIT_CODE);` - the line that makes the commit here necessary.
Ideally it should be an error throw, with setting code (that our errors already do). Also, we already have here the same problem with telemetry: it is not flushed. I think it's out of scope of this PR, but maybe we can create a ticket to follow up on it separately, and hopefully remove the special logic from here eventually?

tests/_common/stats-helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+  sonarUserHome: string,
+): Array<{ scope: string; fingerprint: string }> {
+  const db = openReadonly(sonarUserHome);
+  if (!db) return [];
+  try {
+    return db.prepare('SELECT scope, fingerprint FROM seen_fingerprints').all() as never;
+  } finally {
+    db.close();
+  }
+}
```
Krosovok: This one isn't used. Was there an integration test intended to use it in assers?

src/core/commands/invocation-context.ts
kirill-knize-sonarsource: Removed

src/core/stats/facts.ts
kirill-knize-sonarsource: Done

src/commands/analyze/secrets.ts
kirill-knize-sonarsource: Good idea

tests/_common/stats-helpers.ts
kirill-knize-sonarsource: Removed

src/core/commands/invocation-context.ts
kirill-knize-sonarsource: done

Krosovok (approved):

src/commands/hook/cursor-secrets-block.ts
kirill-knize-sonarsource: Created CLI-1141.

PR #865, opened 2026-09-17
CLI-1130 Move CAG e2e coverage to integration tests except the production CDN smoke

damien-urruty-sonarsource:
## Summary
Most of `tests/e2e/context/` was hitting production `binaries.sonarsource.com` on every post-update, with a 150s timeout, even though the assertions were CLI behavior. This keeps one production-CDN smoke and the real Claude Code e2e, and moves the rest to integration tests against the CAG stub or the harness fake binaries server.

### Kept as e2e
- `cag-offline.test.ts` — production CDN smoke only: download, PGP verify, extract, pinned `--version`, help. Long timeout stays here as `PRODUCTION_CDN_DOWNLOAD_TIMEOUT_MS`.
- `claude-integration.test.ts` — real Claude Code. Already used the fake binaries server; no special CAG install timeout.

### Moved to integration
- Skill → session-start hook refresh for Claude, Copilot, Codex, and Cursor (was `cag-offline` plus `cag-copilot` / `cag-codex` / `cag-cursor`) → one `it.each` in `post-update.test.ts` with the CAG stub.
- Missing project root, multi-root refresh, no-op when `cliVersion` is current, stale-binary cleanup, second self-update hook rewrite (was `cag-edge-cases` / nested `cag-offline`) → `post-update.test.ts`. Stale-binary cleanup uses the fake binaries server (local archive), not the CDN.
- Cursor SonarQube Server skip (was `cag-integrate`) → `context-augmentation.test.ts`. Claude SQS skip and org-not-allowed were already covered there.
- Non-zero CAG exit-code propagation (was `cag-passthrough`) → `passthrough.test.ts`. Auth error, bare `context`, and help forwarding were already covered.

### Deleted as duplicates
- `cag-passthrough.test.ts` — Gradle compression stays in `claude-integration.test.ts`; Compass / session-start content is CAG product behavior.
- `cag-integrate.test.ts` — remaining skip paths already lived in the integrate spec.

Shared seed helpers moved from `tests/e2e/context/_helpers.ts` to `tests/integration/harness/cag-state.ts`.


damien-urruty-sonarsource:
It would take a bigger effort to properly move E2E tests to integration tests because CAG tests use a stub binary. We won't do that now

